PR #1, opened 2016-04-06
fix storage quoting with long long values

vapier:
On 32-bit systems w/64-bit time_t's (like the x32 ABI) we fail to build:
../src/storage/sql_storage.cc: In member function 'virtual void SQLStorage::autoscanUpdateLM(zmm::Ref<AutoscanDirectory>)':
../src/storage/sql_storage.cc:2139:81: error: call of overloaded 'quote(time_t)' is ambiguous
         << " SET " << TQ("last_modified") << '=' << quote(adir->getPreviousLMT())

Handle long long types explicitly to fix that.


v00d00:
Sorry for the slow merge, i get spammed with so much github stuff this slipped through!

Thanks!

PR #5, opened 2016-10-29
Inform the user what port we're trying to bind to

gburca:


v00d00:
This looks great, thanks.

Issue #6, opened 2016-10-29
The web interface is not working


gburca:
Commit 5a7c4f3 broke the web interface. See comment on the commit for more details.

v00d00:
Yeah vanilla libupnp has no header methods in version 1.6, so I it's a WIP to port it over to the 1.8 API.

Corin-EU:
First of all, a super big thank you for reviving Mediatomb and bring it back from near death after it started suffering code rot with ffmpeg/libav and libffmpegthumbnailer library changes.

So the web login problem is not some misconfiguration problem then, because when I enter password, it says on login page and log file just keeps repeating

DEBUG: [/srv/repository/git/mediatomb/src/web_callbacks.cc:62] create_request_handler(): Filename: /content/interface?req_type=auth&return_type=xml&sid=null&action=get_token/index.html, Path: (null)

It would be nice to get mediatomb up and running again.

v00d00:
Hi Guys,

This should be sorted in 9939adef7e98ca4d3dfac2ef9507687363eaac0a.

You will need to build against libupnp 1.8.

Thanks!

gburca:
I tried the web interface with 2a1bfdd and it's still not working on my system. Tried it both with accounts enabled="yes" and enabled="no" and it's still prompting for username/password and refusing to accept the configured username/password. @Corin-EU, is it working for you?

Corin-EU:
I have built the latest v00d00 mediatomb  **r.11dea73** with the latest mrjiminez pupnp **r.7dbf417** on Debian 8.6 

      ldd /usr/local/bin/mediatomb   | grep upnp
         libupnp-1.8.so.10 => /usr/local/lib/libupnp-1.8.so.10 (0x76dd9000)

      ldd /usr/local/bin/mediatomb   | grep xml
         libixml-1.8.so.10 => /usr/local/lib/libixml-1.8.so.10 (0x76d9d000)

      ldd /usr/local/bin/mediatomb   | grep mozjs
         libmozjs185.so.1.0 => /usr/lib/libmozjs185.so.1.0 (0x7679d000)

but am still unable to get any login response to the web server page with the same debug messages

      2016-12-29 21:13:54   DEBUG: [/srv/repository/git/mediatomb/src/web_callbacks.cc:57] create_request_handler(): Filename: /content/interface?req_type=auth&return_type=xml&sid=null&action=get_token, Path: (null)
[...]

PR #8, opened 2016-12-06
Fix build and compilation errors

gburca:


v00d00:
Gabriel you are like santa, bringing lovely pull requests as I sleep...

🎄 🎅

gburca:
I was trying to test the fix for issue #6. I'm not a cmake expert. Perhaps there's a better way to handle the search for alternate lib names.  On my system `pkg-config --libs libupnp` comes back with `-pthread -lupnp4 -lthreadutil4 -lixml4`. Would be nice if those can be fed directly to `find_library()`, but in the mean time this works.

v00d00:
So based on your comment I assume you are using the debian package? I am running against the latest from https://github.com/mrjimenez/pupnp, maybe that is making the difference?

It also explains the compilation errors

gburca:
I'm running on Ubuntu 14.04 LTS, with the default upnp.

My `/usr/include/upnp/FileInfo.h` has:
```
EXPORT_SPEC void UpnpFileInfo_set_LastModified(UpnpFileInfo *p, const time_t *t);
```

It looks like the macro in the latest `pupnp` might not expand to the same function signature. See: https://github.com/mrjimenez/pupnp/commit/2616d4c76ac09ba8217fbf9817dcc45ccf1ba975

Looks like the 2nd arg will need to be modified at compile time to match the upnp version mediatomb is being built against. Or force a minimum upnp version? When building on my system, cmake reports:

```
-- Checking for module 'libupnp'
--   Found libupnp, version 1.8.0~svn20100401
-- Found the upnp libraries at 
-- Found the upnp headers at /usr/include
```

I assumed that matched your v1.8 of upnp. I guess it doesn't.

v00d00:
I think we can to that conclusion at the same time, I've added a check for version and the debian snapshot into CMake and will go through and ifdef where needed.

PR #9, opened 2016-12-07
Fix upnp library call compiler error

gburca:
- The 2nd arg should be `time_t *`, not `time_t`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.6)
 
-set(CMAKE_VERBOSE_MAKEFILE on)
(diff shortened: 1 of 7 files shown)


gburca:
Even though it builds successfully now with both versions of libupnp, only the https://github.com/mrjimenez/pupnp.git version works properly. The default libupnp package in Debian/Ubuntu (14.04 at least) causes the main web page to prompt for userid/password and gets stuck there.

v00d00:
I wonder what the difference is between the two versions, why is the content-type still not set even though we were putting it into the fileinfo. I guess it must be a bug.

Thanks for the fixes!

PR #12, opened 2016-12-12
Changes for Cover Art [folder, track] support, Solaris support too

whorfin:
beyond the cover art work we've been discussing in the forum, I added some tweaks to the Cmake configuration
I definitely appreciate the cmake work - a huge improvement!
however it was stomping on cflags [the -Wall] and broke solaris support, both of which I've fixed
also wee tweak to .gitignore FTW

to build under OmniOS with Joyent packages and a locally built libupnp, FWIW, this worked:
env PKG_CONFIG_PATH=/usr/local/lib/pkgconfig cmake CMakeLists.txt  -DCMAKE_C_COMPILER=gcc -DCMAKE_C_FLAGS=-m64 -DCMAKE_CXX_COMPILER=g++ -DCMAKE_CXX_FLAGS=-m64 -DWITH_MAGIC=1 -DWITH_MYSQL=0 -DWITH_CURL=0 -DWITH_JS=0 -DWITH_TAGLIB=1 -DWITH_AVCODEC=0 -DWITH_DVDNAV=0 -DWITH_EXIF=1 -DWITH_LASTFM=0

More tweaks may be needed for JS; once i verify this is running stably, i'll try that...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,7 @@ config.status
 build/mediatomb
 .idea
 cmake-build-debug
+CMakeCache.txt
+CMakeFiles
+cmake_install.cmake
+mediatomb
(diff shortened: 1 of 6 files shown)


v00d00:
This is awesome!

Big thanks for the Solaris stuff - it's really unlikely that I would have added that myself.

PR #13, opened 2016-12-13
Fix handling of the --debug command line option

gburca:
- The -D/--debug command line option flipped its sense when
  TOMBDEBUG was defined. Now it behaves consistently.
- Removed the redundant DEBUG_LOG_ENABLED (replaced with TOMBDEBUG)


v00d00:
I have similar changes locally but spent my evening trying to debug a memory crash.

Thanks for the patch!

PR #17, opened 2016-12-18
Re-support SIGHUP

whorfin:
This fell on the floor during the cmake migration
autodetecting [an attempt] to enable on X86 platforms. 
The real intent is not to detect X86, but to detect
and not enable on ARM, according to old README
...also a .gitignore tweak you can ignore, if you insist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,9 @@ mediatomb.spec
 mediatomb-service-optware
 .idea
 cmake-build-debug
+*.orig
+CMakeCache.txt
+CMakeFiles
+cmake_install.cmake
+Makefile
+mediatomb
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,6 +336,13 @@ if(NOT HOST_OS)
                 "mediatomb was unable to deterimine the host OS. Please report this. Value of CMAKE_SYSTEM_NAME: ${CMAKE_SYSTEM_NAME}")
 endif()
 
+# X86 Check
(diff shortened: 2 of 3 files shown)


v00d00:
Hi Whorfin!

I'm wondering if there any reason that the SIGHUP wouldn't work on ARM? I've googled it and nothing obvious appears, maybe we should just _always_ build it?

v00d00:
So based on README note I think we should just _always_ build support and strip out the idfefs. We can fix the libupnp shutdown issue if does arise? Sound sensible?

whorfin:
That sounds very reasonable.
As you suggest, it may be very likely the issues are due to the Old 'N Busted version of libupnp that were part of the old mediatomb.
I didn't want to hose anyone, so I went with the more conservative action of mimicking the automatic logic of "only x86", but feel free to go the simpler route.
Mind the "todo" comment at line 41 of src/main.cc, which darkly hints at lurking horrors.

v00d00:
> darkly hints at lurking horrors.

:ghost:

I can definitely understand being conservative, but as the saying goes you have to break a few eggs to make an omelette!

Do you want to close this PR or push a commit to go the simple and way remove the ifdefs?

whorfin:
I'll push a commit to do as we discussed
Won't get to it until later

whorfin:
OKdoke, here ya go

v00d00:
Thanks! We can remove the X86 check now too surely?

whorfin:
True
I left it in case it were needed again so someone else doesn't have to figure out how to do that again.
Stackoverflow missed a case...

Issue #14, opened 2016-12-13
C++ assumptions


gburca:
I wanted to have a bit of an open discussion about what the expectations/directions are for the code base going forward, specifically for this fork. The existing code base contains a number of constructs that are now provided by most modern C++ compilers.

Is it safe to modernize the code base? If so, should it target C++11? C++14?

Many of the utilities in the `zmm` and `hash` directories have more robust cousins in C++11 or later. For example: string, regex, atomic, array, stack, queue, hash, shared pointer. Is it worth transitioning to the standard C++ equivalents over time?

One of the reasons I'm bringing this up is that I've noticed various issues. For example, some song titles are garbled when viewed on the web. Is that due to MediaTomb's String class? Or is it somewhere else in the mp3/flac tags, taglib, internal MT code, iconv, JS scripts, html output chain? Who knows. There's lots of custom string handling code around, and I suspect some of it may be responsible. I'd feel better about it if std::u16string or one of its friends was used to handle user strings.

I've also seen error logging from the hash table code about "maximal number of iterations exceeded". I haven't debugged it, but I'd feel better about it if a standard C++ container was used instead.

That by itself is not enough justification to make massive sweeping changes, but perhaps over time new code can start using std::string, std::stack, etc... instead of the zmm::Foo versions.

v00d00:
> Is it safe to modernize the code base? If so, should it target C++11? C++14?

I think realistically for the project to be sustainable there is no choice but to modernize the codebase. The more standardised the code the more developers will be able to understand it and (hopefully) contribute which is important to keeping the project/fork healthy over time.

Locally I already have `set_property(TARGET mediatomb PROPERTY CXX_STANDARD 14)` in CMakeList.txt - I should probably commit that and add something to the README.

> Many of the utilities in the zmm and hash directories have more robust cousins in C++11 or later. For example: string, regex, atomic, array, stack, queue, hash, shared pointer. Is it worth transitioning to the standard C++ equivalents over time?

Absolutely and I would hope sooner rather than later. Perhaps a good transitionary step would be reimplement those classes as a wrapper of the modern standard equivalents.

> I've also seen error logging from the hash table code about "maximal number of iterations exceeded". I haven't debugged it, but I'd feel better about it if a standard C++ container was used instead.

Me too, I would like to get rid of `zmm` entirely. But that is not a small task!

gburca:
[...]

PR #18, opened 2016-12-18
Replace Custom Hash Implementions with STL

v00d00:
Start to modernise parts of the codebase as per #14.

This replaces the custom DB*Hash objects with their STL versions.


v00d00:
@gburca thoughts?

src/tools.cc
```diff
@@ -986,13 +986,14 @@ unsigned int stringHash(String str)
     return hash;
 }
 
-String intArrayToCSV(int *array, int size)
+String intArrayToCSV(std::shared_ptr<std::unordered_set<int> > array)
```
v00d00: Agreed.

PR #21, opened 2016-12-20
solaris linker flag tidying, as requested

whorfin:
will need help testing, this works on solaris
shows there is a [semantic] bug in the package finder for libuuid, works around it

BTW, github is insisting the pull you accepted didn't happen, even though i can see it did, and including previous commits[Dec 18] in this pull?


whorfin:
...it's only 09b7385 you want here

v00d00:
It's probably because I squash merged your commits. I guess you will need to repull from my repository?

v00d00:
@whorfin does 301194955c00fd8b410312138a492cf7e811a9e7 fix the UUID part at least?

whorfin:
I will check tonight; that's what I would have tried.
I've been chided in the past on other projects for touching included .cmake files, with the philosophy being that included files should only be changed upstream by their maintainers.
Personally I have no opinion in this matter, so if you're ok with that, i'm good, but that's why i didn't go that route.

Cheers

whorfin:
...had to both pull and merge to make things happy
Also, I went ahead and made the change to the .cmake file, and pulled that from CMakeLists, to
get UUID working all happy and clean [falls back on library searching if it can't find as package, which
seems like it ought to be default behavior in general but hey]

whorfin:
...looks like you added your solaris-tidy back to master, meaning the .cmake changes conflict
feel free to go with either one, looks like the only change is name-decorating on the conditionals

PR #22, opened 2016-12-26
Re-Fix Solaris, Further Fix for 1.8 LibUPNP

whorfin:
Re: 7310ef10a0989711da7bf35fefebea51c9bab1df, we can't totally remove OS detection, because there are #ifdef SOLARIS references in the code, required for proper compilation, and for safety with the LD_PRELOAD trick for Posix semantics on NULL printfs.
Hence, detection and insertion of -DSOLARIS is still required.

If you'd prefer trimming the non-solaroid os portions of detection, we can do that, but since there
is one dependency, it made sense to me to have a nice clear "here is where we detect the various OSs" section, distinct from "here is where we do something based upon this"

Also for bonus points, when I did a full merge, that required re-pulling pupnp for the 1.8 named directories.
Re-building didn't clear the old ones, which I did to ensure cleanliness.
...and promptly found a few places in mediatomb referencing non-1.8 paths; surely this is a bad idea.
Fixed.

Cheers


v00d00:
Ahhhhh balls. Yep you are right.

Good catch on the includes. Clearly my feel skills are weak!

v00d00:
Im happy to merge this but there are a load of extra commits showing up? Can you rebase or shall I just cherry pick be624a0e54fb3ea496e011fa0e788e8b5f050fb7 and e186429d9fa9d75cfddc63f8ad7d7f69c2ad6b45?

whorfin:
yeah, they're all still there but amount to nothing.
gimme a moment and i'll rebase

whorfin:
...hang on, this is embarassing...
please stand by, my branch is slightly buggered

whorfin:
OK, here ya go, sorry for previous bad hygiene...

v00d00:
Thanks!

PR #29, opened 2017-01-07
Eliminate sync.h - Use STL mutex and friends throughout

gburca:
The old pthreads mutex and condition variable (sync.{h,cc}) was completely eradicated in favor of the STL equivalents.

The Timer interface had to be modified to be compatible with the new Singleton usage, which in turn had to be modified to work with the STL mutex/recursive_mutex/conditional_variable.


v00d00:
Now that is a proper pull request!

I should hopefully be able to take a look at this tonight or tomorrow evening but a first glance looks good.

gburca:
Some tips for the review.

- The Singleton template has a mutex member (for creating the singleton), so most classes that inherit from Singleton will re-use that mutex for other purposes in the derived class.
- The Timer's add/remove subscriber used to take a `Ref` (in some cases `Ref(this)`!), but now it takes a plain pointer to an interface. I'm not sure smart pointers are appropriate in this case. The onus is now on the caller to remove the subscriber before deleting the subscribed object. I toyed with the idea of subscribing using std::function, or lambdas, but that tends to obscure what variables are captured, and how scopes and lifetimes are impacted. It's also trickier to un-subscribe a lambda.
- For Object, in order to be able to have proper const derived classes, the reference counting should be mutable, otherwise no derived object can be const.
- There was not much testing done because ... well, there are no test cases, and I have a limited environment to test in, but it builds (with the default settings), runs, and plays music. ;)

Thanks for looking at it. One step closer to eliminating STL duplicates.

v00d00:
Thanks, seems to work with my testing even when I threw all my devices at it! Great work, another one bites the dust.

PR #35, opened 2017-01-18
re-silence "event is ok" info message

whorfin:
Super simple - e4c0f5c769ed313fbf5fe95c983ee7251fb7b6eb left the un-commented "event is ok" message, which will crap up one's log rather quickly.  
I assume that was an unintended side-effect.
Just re-commenting it back out.


v00d00:
Yeah, whoops!

PR #39, opened 2017-01-25
Add includes for solaris for USHRT_MAX and signal handling

whorfin:
Recent changes broke my build; some things not defined by default.
Particularly because the ISO defines aren't necessarily standard, i put them in an ifdef block.
If you feel they're safe for everyone, by all means pull them out.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -36,6 +36,12 @@
 
 #include <getopt.h>
 
+// first do no harm; may be safe for others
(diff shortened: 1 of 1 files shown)


v00d00:
Can you move signal header out of the ifdef and change it to the c++ style `<csignal>` then happy to merge.

whorfin:
done

PR #45, opened 2017-02-10
Avoid warning by explict casting

whorfin:
...,but note that this all reveals that perhaps zmm::String should be replaced with std::string, and
anyway, why is zmm::String.length() not a size_t?


v00d00:
> perhaps zmm::String should be replaced with std::string
Or something that supports proper Unicode, absolutely, its on the ToDo.

> why is zmm::String.length() not a size_t?
Who knows!

Thanks for the patch

PR #50, opened 2017-03-17
CMake: don't add threadutil to UPNP_LIBRARIES

jcowgill:
In mrjimenez/pupnp#22 and mrjimenez/pupnp#25 the threadutil library was removed from libupnp 1.8 so mediatomb should not try to use it.


v00d00:
LGTM thanks!

Issue #54, opened 2017-03-24
Rebrand


v00d00:
So having spoken to the original authors of MediaTomb they would rather keep the name mediatomb associated with them, which is fair enough.

So we need a new name for the project going forward....

tommyp-vr4:
Some ideas to throw out there. 
UDMA (ultimate digital media archive; nice geeky play on letters)
dlnArchive (play on words with dlna but doesn't roll off the tongue)
hypermedia
mediarevelation (because mediatomb was brought back from the dead)
mediacache or mediadepot

v00d00:
I was thinking a bit more abstract probably an animal or a plant or something like that.

gburca:
I'd vote for something unique. Searching for UDMA is bound to return everything except this project. Bonus points if the domain name is not yet taken.

Unfortunately obvious names like MediaVault are already in use. FaSoLa seems more unique than DoReMi, but the domain names are taken.

Are the original authors OK with copying the static web site content to the new project?

v00d00:
Ive gone with "Gerbera" after a[ family of pretty flowers](https://en.wikipedia.org/wiki/Gerbera).

Here is my badly inkscaped attempt at logo:

<Img src="https://github.com/v00d00/gerbera/blob/master/artwork/logo-horiz.png?raw=true" width="50%"/>

And I have registered a fasionable `io` domain: `gerbera.io`, we are modern now! 😆 

@gburca I've mailed the original authors about the content on mediatomb.cc, its not obvious what licence the sites content is under it just says "copyright", so we shall see what they come back with.

v00d00:
If anyone knows a friendly web developer who wants to make us a website, please send them my way.

elmodaddyb:
Hello, Great to see Gerbera on GitHub.  I am playing around with the front-end of MediaTomb (frameset 👎 )... moving to a modern client application using Angular with ES6, Bootstrap 4.

Wondering if there is any appetite for a modern client side for Gerbera?  Or is branding, solely to remove the references to MediaTomb?

Any insights....thoughts...direction?

Thanks,
E

v00d00:
@elmodaddyb: There is absolutely an appitite for a modern UI, its high on the wish list.

In terms to technologies ideally for simplicity I would like a SPA that has no compilation requirement, BS4 is ideal, and uses jquery so perhaps that would be enough in terms of libraries for this case.

elmodaddyb:
[...]

PR #56, opened 2017-03-25
Script for generating server icons, perhaps useful in the Great Rebranding

whorfin:
Plopped this in scripts, useful for users looking to customize, more importantly perhaps helpful
to you as you deal with #54 

Cheers


v00d00:
Thanks!

PR #58, opened 2017-03-26
more great rename

tommyp-vr4:
Was compiling for Debian and noted these two items that needed renaming.  The xpm file is a file-save-as of one of the gerbera images.  The docs/gerbera.1 is sort of a hack job because I only edited with this webpage (I'm new here).  I'm not too sure how to remove the user and group settings writesups (I believe those options are removed).


v00d00:
Thanks!

This looks good.

Great to see someone new getting into open source, I would take a look at [this nice git/github intro](http://product.hubspot.com/blog/git-and-github-tutorial-for-beginners) to save you editing through the website.

PR #59, opened 2017-03-26
Install man page

whorfin:
While this ad9eefc77a7ac9ce2f9211841b1f59ad755dfd69 updated the man page, nothing was installing it.
Fixed


v00d00:
Nice spot, thanks!

Issue #47, opened 2017-03-13
Update JS Engine/Nuke Spidermonkey


v00d00:
Our spidermonkey code is very old, as is the version of spidermonkey it needs.

Its also way over the top for our scripting needs.

Lets switch to something smaller and designed to be embedded such as [Duktape](http://duktape.org/) or [MuJS](http://mujs.com/docs/reference).

sdrik:
Did you already start working on this ? If not, I'd like to volunteer.

v00d00:
Hey, I havent started working on it yet. I look forward to seeing what you come up with.

gburca:
Are the two proposed replacements strict supersets of Spidermonkey as far as gerbera is concerned? If not, some of the scripts that used to work might break when swapping the JS engine. Debugging JS is tedious for this type of embedded environment. At least with spidermonkey or node.js there's no question of what JS language features are supported.

v00d00:
In the case of `duktape`, which I am leaning towards, it implements ECMAScript 5.1, with SpiderMonkey (1.8.5) we are on ECMAScript 5.0. It also has nice things like [Encoder API](https://encoding.spec.whatwg.org/) and some more modern ES6/7 features. We _should_ only gain more support from using it.

v00d00:
It also has a [provided debugger](https://github.com/svaarala/duktape/blob/master/doc/debugger.rst).

gburca:
That sounds good.

v00d00:
This was done with #64. Thanks to @sdrik

PR #64, opened 2017-04-12
Replace SpiderMonkey with Duktape

sdrik:
This PR will track my work on the replacement of the SpiderMonkey JS engine by Duktape (see #47).
It is still work in progress and received almost no testing, I only checked that the standard import_script.js works for my audio collection.


v00d00:
Nice! Thanks for the patch. You will have to give me a couple of days to review this one. Any chance you can give me an overview to help with review?

sdrik:
As said before, this is work in progress, so I'm not sure it should go in yet. There are still areas which need improvement (like locking and destructors).
I kept as much as possible of the existing implementation, so the flow is the same as before: each specialized script is encapsulated in a subclass of the Script base class, with pretty much the same public interface.
There is one shared Duktape heap allocated in the Runtime singleton. Each Script instance get one dedicated Duktape context (basically a stack and a global environment). As heaps are not thread-safe I used mutex-locking for proper isolation.
There are several places where value are not passed between (internal) functions by argument but through the stack. The advantage is that it was easier to follow the existing code that way, the drawback is one need to keep in mind the status of the stack throughout the flow. For example, when a script is compiled in the _load() method it is left on the stack, then _execute() method can pick it up. Another example is the various setProperty/getProperty which act on the object present on the top of the stack instead of receiving a pointer. This will certainly need to be refined at some point.

v00d00:
Hows progress?

sdrik:
It's hopefully in good shape now for a review, but probably not bug free...

v00d00:
Hi, sorry about introducing the conflicts by removing the DVD stuff.

Can you rebase please? Thanks!

sdrik:
No problem at all.
PR rebased.

v00d00:
So this is really great, thanks! I've had look now and I like it so far. I will finish up on Monday most likely but 👍!

PR #69, opened 2017-04-13
Make SQL album art fetching work on MySQL

cweiske:
Resolves: https://github.com/v00d00/gerbera/issues/68


v00d00:
Thanks for the patch!

After a brief look i wonder if it just the missing whitespace that was causing your issue?

cweiske:
No, it's not the whitespace. MySQL is stricter than SQLite. If you use a backtick, it means the following is a SQL field name. That's why it says "Unknown column 'cover.jp%' in 'where clause'". Using correct quotes is essential here.

v00d00:
Ahh, I see. Thanks for the clarification.

PR #81, opened 2017-05-05
Fix broken build when WITH_JS=0, duktape patch for OmniOS/Solaris

whorfin:
duktape.o was always built even when WITH_JS unset, and this would fail since duktape.c is actually C++ but the compile switch was only done in the WITH_JS branch.
Cleanup duktape README.  
Quick patch for OmniOS/Solaris to duk_config.h, pending upstream patch svaarala/duktape#1514
  for extra hilarity, committing a change to the duktape git repo triggers a git fsck bug on OmniOS; had to do that from Ubuntu


v00d00:
This is great, thanks again Whorfin.

PR #90, opened 2017-05-27
Fix #86 - SessionManager checkTimer deadlock

### LINKED ISSUE #86: SessionManager checkTimer deadlock?
Hello All,

I am running a recent local build of Gerbera on my local and hitting a deadlock consistently when the `SessionManager.checkTimer()` clears the stale sessions.

> Here is the scenario.

1. Open Gerbera web UI @ `http://localhost:49152`
2. Establish a session (with SID and logged in)
3. Wait for `SessionManager::timerNotify` to fire, and cleanup the stale sessions

It looks like the `Timer` mutex is `AutoLock` 2 times. First on the `Timer::notify`, to call the SessionManager cleanup, then again on the `Timer::removeTimerSubscriber`

```
void  Timer::notify()
{
    AutoLock lock(mutex);
```
Sessions are removed via `SessionManager::timerNotify` and the `SessionManager::checkTimer` is called to clear the Timer.

```
void Timer::removeTimerSubscriber(Subscriber* timerSubscriber, zmm::Ref<Parameter> parameter, bool dontFail)
{
    log_debug("Removing subscriber...\n");

    AutoLock lock(mutex);

   ^^^^^ DEADLOCK ^^^^^
```

The deadlock consistently happens on the second call to the `AutoLock lock(mutex)` for the `Timer` instance.

My guess is that I have missed some simple configuration or runtime difference, but cannot figure it out.  Likely due to inexperience with C++.

Any helpful direction would be appreciated.  If anyone else sees similar issue and has ideas it would be great!.

Thanks,

Eamonn

elmodaddyb:
@v00d00 - Here is the `SessionManager::checkTimer()` deadlock fix.

1. Change `Timer` class to `recursive_mutex` - allowing calls for lock multiple times from same thread
2. Add intermediary `std::list` to avoid corruption of the `subscribers` list when iterating, then deleting a subscriber

I did the following tests.

1. Build and Run **gerbera**
2. Login/Logout
3. Wait for Session Timeout
4. Add Timed scan on a directory

I noticed that the project does not have any unit tests.  To that end, I also created the start of unit tests using GoogleTest.

Let me know if you think any additional tests are warranted (i.e. Travis CI build?).

>Please consider my experience level with C++ is **novice** so any best practices or bad behaviors feel free to point out directly.

Thanks,

Eamonn

### DIFF
--- a/src/timer.cc
+++ b/src/timer.cc
@@ -126,20 +126,28 @@ void Timer::triggerWait()
 void  Timer::notify()
 {
     AutoLock lock(mutex);
+
+    std::list<TimerSubscriberElement> toNotify;
+
     for (auto& element : subscribers) {
         struct timespec now;
         getTimespecNow(&now);
         long wait = getDeltaMillis(&now, element.getNextNotify());
 
         if (wait <= 0) {
-            element.notify();
-            if (element.isOnce()) {
-                element.disabled = true;
-            } else {
-                element.updateNextNotify();
-            }
+            toNotify.push_back(element);
         }
     }
+
+    for(auto& element : toNotify){
+        element.notify();
+        if (element.isOnce()) {
+            element.disabled = true;
+        } else {
+            element.updateNextNotify();
+        }
+    }
+
     subscribers.remove_if([](const auto& e) { return e.disabled; });
 }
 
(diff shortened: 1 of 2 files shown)


elmodaddyb:
Sadly, I let this soak for an hour...and it seems the intermediary list causes a memory leak :(

My c++ skills shining brightly here!! 👎 

Back to drawing board.

PR #91, opened 2017-05-28
Fix #86 - SessionManager checkTimer deadlock

### LINKED ISSUE #86: SessionManager checkTimer deadlock?

elmodaddyb:
Hello, here is round 2 for #86.  C++ won round 1.  Thanks to @gburca for suggestion on solution.

I was able to run **gerbera** locally, watch a few movies & listen to music.  The subscribers were cleaned up as expected when new web sessions were created, and timed out.  No obvious memory leaks after running for a few hours.

The only item **NOT** tested is the `element.isOnce()` condition.  Not sure how to test that directly?

Let me know what you think.  Any guidance on C++ convention and standard is appreciated. 👍

### DIFF
--- a/src/timer.cc
+++ b/src/timer.cc
@@ -126,21 +126,34 @@ void Timer::triggerWait()
 void  Timer::notify()
 {
     AutoLock lock(mutex);
-    for (auto& element : subscribers) {
+
+    std::list<TimerSubscriberElement> toNotify;
+
+    for (auto it = subscribers.begin(); it != subscribers.end(); ) {
+        TimerSubscriberElement& element = *it;
+
         struct timespec now;
         getTimespecNow(&now);
         long wait = getDeltaMillis(&now, element.getNextNotify());
 
         if (wait <= 0) {
-            element.notify();
-            if (element.isOnce()) {
-                element.disabled = true;
+            toNotify.push_back(element);
+            if(element.isOnce()){
+                it = subscribers.erase(it);
             } else {
                 element.updateNextNotify();
+                ++it;
             }
+        } else {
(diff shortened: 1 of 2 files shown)


v00d00:
LGTM, Thanks!

Issue #75, opened 2017-04-17
Listing artists and albums is slow


cweiske:
Since the introduction of folder images (which I absolutely love!), opening folders with many subfolders is too slow. It takes around 10 seconds to get the list of artists here (about 130). This is too long.

v00d00:
I've just tested this with my collection and speed seems fine, though I am using sqlite.

Could you try with sqlite and see if you have the same issue? That way we know to focus on mysql or not.

v00d00:
Do you mean in the Web UI?

cweiske:
No, not the web UI. I mean listing albums and artists is slow with a UPnP client.

I'll have a look at it later this week. The next few days I'll be away.

cweiske:
I enabled MySQL's general query log and got a log of all quries.

Now I see that fetching the cover image for a single album takes 200ms:

```
RESET QUERY CACHE;
SELECT `id`
FROM `mt_cds_object`
WHERE
 (
  lower(`dc_title`) LIKE 'cover.jp%'
  OR lower(`dc_title`) LIKE 'albumart%.jp%'
  OR lower(`dc_title`) LIKE 'album.jp%'
  OR lower(`dc_title`) LIKE 'front.jp%'
  OR lower(`dc_title`) LIKE 'folder.jp%'
 )
 AND `upnp_class`='object.item.imageItem'
 AND `parent_id` IN (
  SELECT `parent_id`
  FROM `mt_cds_object`
  WHERE
   `parent_id`='14343'
   AND `object_type`=2
 )
ORDER BY `dc_title` DESC
```

Doing that for the 261 albums I have takes time.

cweiske:
I wonder why that subquery is needed. It returns the `parent_id` for items with the `parent_id = 14343`.

When removing that subquery and replacing it with `AND parent_id='14343'` I'm down to 2ms for the query.

v00d00:
Hmm yeah, that doesnt make too much sense at first glance!

Are you happy having a stab at putting together a patch for this?

cweiske:
I digged a bit deeper.

The original patch https://sourceforge.net/p/mediatomb/discussion/440751/thread/6578cd15/#f4a5 did two subselects.
It was made for virtual folders and items, fetched the folder IDs of the *real/original files* - and looked there for images.

The current code only looks for files that are directly in the (virtual) folder.

cweiske:
[...]

PR #92, opened 2017-05-29
Re-enable Virtual Folder album art

whorfin:
I've now found the use-case, using eg foobar2000, where Virtual Folder cover art isn't returned.
the cover.jpg/etc artwork was implemented to work when using "PC Filesystem" mode in
DLNA clients.  Browsing via for example "Audio" then "Artists" was not working,
as the virtual root didn't match the physical.  Now both cases properly return
file-based cover art.

In case #75 reappears for MySQL folk, there is currently a define `ONLY_REAL_FOLDER_ART ` to disable this addition.
I note that in my scenario [six thousand albums] returning "Audio/Albums" hits the network timeout for
foobar2000 mobile, whether or not `ONLY_REAL_FOLDER_ART` is enabled.  With Sqllite, peformance
is otherwise fine in either case.

### DIFF
--- a/src/storage/sql_storage.cc
+++ b/src/storage/sql_storage.cc
@@ -1236,7 +1236,28 @@ String SQLStorage::findFolderImage(int id, String trackArtBase)
        << "LIKE " << quote(_("folder.jp%"));
     *q << " ) AND ";
     *q << TQ("upnp_class") << '=' << quote(_(UPNP_DEFAULT_CLASS_IMAGE_ITEM)) << " AND ";
+#ifndef ONLY_REAL_FOLDER_ART
+    *q << "( ";
+    *q <<       "( ";
+        // virtual listing via Album, Artist etc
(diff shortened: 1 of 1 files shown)


whorfin:
8player on iOS handles all Albums in Audio/Albums mode, so timeout behavior seems to have more to do with how DLNA client makes requests
foobar2000 seems to try all at once
8player does smaller batches, and therefore also seems wickedly fast

v00d00:
Hey Whorfin,

Can you post an EXPLAIN for the super slow query?

whorfin:
I don't understand the request, what would you like?

Background:
In #75 it was observed that MySQL had troubles with a more complicated album art query.  That query was in fact fundamentally flawed.
This query does the right thing, but I'm cautious only because the behavior seen in #75 was still a bit surprising.
I also observed that different DLNA clients batch requests differently.

PR #95, opened 2017-05-31
Fix minor spelling mistake

jcowgill:
This was in one of Debian's mediatomb patches so it seemed sensible to submit it upstream so that patch can be dropped.


v00d00:
In a comment as well!

All patches are welcome though, thanks!

Issue #87, opened 2017-05-23
macOS build fails


tristan-k:
I tried to build gerbera on macOS (10.12.4). It fails because not all dependencies can be satisfied. Please have a look. 

```
$ brew install libexif libmagic lzlib mp4v2 spidermonkey sqlite taglib ffmpeg ffmpegthumbnailer id3lib expat libupnp ossp-uuid mysql
```

```
$ git clone https://github.com/v00d00/gerbera.git
$ mkdir build
$ cd build
$ cmake ../ -DWITH_MAGIC=1 -DWITH_MYSQL=1 -DWITH_CURL=1 -DWITH_JS=1 \
-DWITH_TAGLIB=1 -DWITH_AVCODEC=1 -DWITH_EXIF=1 -DWITH_LASTFM=0
-- The C compiler identification is AppleClang 8.1.0.8020042
-- The CXX compiler identification is AppleClang 8.1.0.8020042
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- Found Threads: TRUE
-- Looking for nl_langinfo
-- Looking for nl_langinfo - found
-- Looking for setlocale
-- Looking for setlocale - found
-- Could NOT find INOTIFY (missing:  INOTIFY_INCLUDE_DIR)
-- Performing Test ICONV_SECOND_ARGUMENT_IS_CONST
-- Performing Test ICONV_SECOND_ARGUMENT_IS_CONST - Failed
[...]

PR #97, opened 2017-05-31
Add macOS launchd script

elmodaddyb:
As a follow up to #87 - Add the launchd script to the `scripts` directory.


v00d00:
Thanks!

Issue #107, opened 2017-07-04
lastfmlib


marceloneil:
It took me a bit to figure out that I needed the 2009 [lastfmlib](https://code.google.com/archive/p/lastfmlib/) rather than the more recent [liblastfm](https://github.com/lastfm/liblastfm). Would it make sense to create an `install-lastfm.sh` script similar to the other dependency install scripts, or is there something that I'm missing here. If an installer script would be useful, I've already made one and would be happy to make a PR

v00d00:
If you would like to contribute your script, please do!

I think the newer lib needs Qt, so cant use that.

PR #108, opened 2017-07-05
lastfmlib install script

marceloneil:
Completes #107 

Similar to any other install script in the `scripts/` directory, run it using bash. It requires a c++ compiler, libcurl, make, pkg-config and wget to run.


v00d00:
Thanks!

Issue #121, opened 2017-07-25
Enhanced Metadata for image/jpeg


r-clips:
Would really appreciate ability to access Xmp tags for image files.

I have been using mediatomb for several years to serve up a photo collection with a virtual directory structure organised by keyword tags applied by photo oganisers like ‘Shotwell’, ‘Digikam’ etc 
These Tags are not available through libexif and are typically stored as Xmp or IPTC tags ( specifically Xmp.dc.Subject ).

With Mediatomb, I hacked together this functionality through libextractor, but I see that libextractor has gone from Gerbera.  The alternative is to use exiv2 directly to access Xmp metadata - Exiv2 still seems to be present in Gerbera code, but is currently disabled.

I have put together a set of patches that :
      -  re-enable exiv2 library interface in gerbera
      -  add code to exiv2_handler to support ‘auxdata’  library-option to access either Exif or Xmp tags

My coding skills ( particularly in C++ ) are fairly rudimentary, but the attached patches work for me. Make use of anything that might be useful.

[gerbera-exiv2-patches.zip](https://github.com/v00d00/gerbera/files/1172476/gerbera-exiv2-patches.zip)

Also, I note that it looks like exiv2 supports matroska mkv containers, so this could go some way towards addressing Issue #30 

Incidentally, really appreciative that Mediatomb has been resurrected and polished up for 2017

v00d00:
Hi,

Thanks for this, it looks great. Can you open a pull request for these?

v00d00:
Which email/name do you want me to use when committing these?

r-clips:
OK, just reading some git-hub howto's ....
As soon as I get to the bit that explains pull-requests, I'll give it a shot .....

>> Which email/name do you want me to use when committing these?
 
  I assumed this would be the git-hub user name.

v00d00:
To clarify, I asked for you email/username as I could have taken the patches from the zip and commited them on your behalf if you would have preffered. It looks like you got the PR sorted though, so no need.

Thanks for the contribution!

We need to put some documentation together really.

PR #122, opened 2017-07-26
Re-enable EXIV2 library for image/jpeg metadata

r-clips:
This change provides enhanced metadata support for image/jpeg files as per Issue #121 
In particular, access to Xmp tags is provided to enable a uPnP virtual folder structure to be organised according to Photo Tags stored in Xmp.dc.Subject by applications such as Shotwell or Digicam.

The original EXIV2 library interface that was present in the original Mediatomb code base is reenabled by setting the 'WITH_EXIV2' build flag.

exiv2_handler code is modified to allow both Exif and Xmp metadata tags to be retrieved by means of the 'auxdata' interface if configured  in config.xml as per the following example : 

`    <library-options> 
      <exiv2>
        <auxdata>
         <add-data tag="Exif.Image.Model"/>
         <add-data tag="Exif.Photo.DateTimeOriginal"/>
         <add-data tag="Exif.Image.Orientation"/>
         <add-data tag="Exif.Image.Rating"/>
	 <add-data tag="Xmp.dc.subject"/>
        </auxdata>
      </exiv2>
    </library-options>`


v00d00:
Thanks!

PR #132, opened 2017-08-13
Fix native-uuid for Solaris/OmniOS

whorfin:
The changes to support "native uuid" once again broke Solaris/OmniOS
This should fix that


v00d00:
Urgh, those includes need tidying, thanks for the patch as always @whorfin :+1:

PR #137, opened 2017-08-20
upnp includes no longer installs in "upnp-1.8" dir, but rather just upnp

whorfin:
Tracking latest upnp, i noticed it no longer installs headers in "upnp-1.8".
You won't notice this if you've updated, as the old -1.8 include subdir is not
removed, leading to a dangerous case of versionitis.

### DIFF
--- a/src/action_request.h
+++ b/src/action_request.h
@@ -33,7 +33,7 @@
 #ifndef __ACTION_REQUEST_H__
 #define __ACTION_REQUEST_H__
 
-#include <upnp-1.8/upnp.h>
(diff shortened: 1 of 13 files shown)


v00d00:
I think we might need to un-qualify those headers, I made some steps supporting 1.8.2/master in https://github.com/v00d00/gerbera/commit/e5e5e4f4e5a87390d5a4b4f7dd97da4703621184. Those changes will mean we pass the correct full header path (upnp or upnp-1.8) instead of the parent include directory.

PR #140, opened 2017-08-31
Updated Scripts and Build Instructions for FreeBSD

ghormann:
This adds instructions for compiling under FreeBSD and fixes two of the install scripts to be compatible with FreeBSD.


v00d00:
This is awesome thanks!

I would prefer to keep a single README.md, so would you be okay to in-line those changes?

ghormann:
No problem.   Consolidated via 0954108905a563dcbf24cc5cde49e1ba922b0db1

v00d00:
Great!

PR #141, opened 2017-08-31
Solaris SMF ingredients

whorfin:
Greetings

Somewhere along the way, I lost my commits/pulls/whatnot for the actual goodies to have gerbera run as a proper managed service under Solaris.  scripts seemed like the right place for them.  There's an
assumption here about where config files etc live, but it should be easy to understand for a Solarian

Cheers


v00d00:
Nice, thanks!

PR #164, opened 2017-09-25
Fixes #157 - Open /tmp using O_EXCL to avoid race condition

### LINKED ISSUE #157: File in /tmp opened without O_EXCL
Gerbera opens a file in `/tmp` (the default value of `CFG_SERVER_TMPDIR`) without `O_EXCL` here (and later on in that function): https://github.com/gerbera/gerbera/blob/master/src/process.cc#L62

This is vulnerable to the classical `/tmp` directory race conditions (such as symlink racing) which may happen if the chosen filename is created by another user just before it is opened. This can be fixed by either opening the file using `O_EXCL`, or preferably using a function like `mkstemp`.

There are other uses of `tempName` in `src/transcoding/transcode_ext_handler.cc`, but I think they're safer because `mkfifo` will fail if the file exists.

elmodaddyb:
Tested with simple Active Item that hit a shell script.  Files generated OK in `/tmp` and cleaned up afterward.

### DIFF
--- a/src/process.cc
+++ b/src/process.cc
@@ -59,7 +59,7 @@ String run_simple_process(String prog, String param, String input)
         
     Ref<ConfigManager> cfg = ConfigManager::getInstance();
     String input_file = tempName(cfg->getOption(CFG_SERVER_TMPDIR), temp_in);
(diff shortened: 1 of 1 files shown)


v00d00:
Great thanks!

PR #173, opened 2017-11-02
Fix static linking with ffmpeg

ffontaine:
avformat depends on avcodec so it should be added first otherwise static
linking will fail


v00d00:
Thanks! Let us know how you get on with your static build.

ffontaine:
static build was successful after one update to pupnp (to rename uuid_unpack as it is also defined in libuuid) and one more update to gerbera (missing swresample dependency). I will make a new PR for this one.

PR #174, opened 2017-11-02
fix static linking: add swresample

ffontaine:
Add swresample to FFMPEG_LIBRARIES in FindFFMPEG.cmake otherwise static
linking will fail as swresample is a dependency of avformat and avcodec


v00d00:
Thanks!

PR #177, opened 2017-11-07
updated documentation for FreeBSD installation

imp1sh:
changed documentation because when installing on FreeBSD -DWITH_SYSTEMD=0 has to be set


v00d00:
Awesome, thanks!

Issue #138, opened 2017-08-21
Browsing Media Library Slow with MySQL


LinuxUser2017:
Hello,

I'm noticing a fairly consistent delay when browsing through folders. It can take up to 30 seconds (if I have the patience to wait) for the contents of folders to show.

I am using Mysql on another computer to handle the database. I have found it's doing a 'SELECT 'id' FROM 'mt_cds_object' WHERE ( lower ' ..... etc query while I'm waiting.

Would I be right in saying it's looking for cover images /album art ?

I will do a bit more testing to investigate. Is there any documentation on how to setup a fresh Mysql database for Gerbera ? (I'm using my old mediatomb one). It's ~83 Meg, (Is that too big, or a feasible size to query quickly?)

Thanks,

v00d00:
I will have to set up a MySQL instance to test with, there have been reports of it being slow, but fixes where merged to improve it.

v00d00:
Sqlite has always reported to be fast, though.

v00d00:
@LinuxUser2017 would it possible for you to mail me a mysqldump of your Gerbera DB so I can run some tests? thev00d00 AT gentoo DOT org

LinuxUser2017:
Hello,

To avoid wasting your time, I have setup a fresh database (not a recycled mediatomb database). I willl report my findings once it approaches the same size.

LinuxUser2017:
Hello,

I fired up a fresh Gerbera (populated) database ~40mb.

Took 75 seconds to timeout on showing the media library. (bubble upnp).

Sql query in process:
|  734 | gerbera | securitypi:44172 | gerbera | Query   |   75 | Sending data | SELECT `id` FROM `mt_cds_object` WHERE ( lower(`dc_title`) LIKE 'cover.jp%' OR lower(`dc_title`) LIK |

Also noteworthy mysql is running at 100% CPU usage while this is happening.
Once this query clears the library becomes responsive again.

v00d00:
That is not ideal. Can you send me your DB so I can experiment?

Really we need to re-engineer the artwork stuff entirely, but that is no small task.

LinuxUser2017:
Hello,

I sent you a file 24 Aug.

cweiske:
Can this be closed? That bug is 7 years old, any gerbera internals changed quite a bit over time.
If the problem can be reproduced on a current 2.2.0 gerbera version it can re-opened. Until then it should be closed.

KarlStraussberger:
Good point

PR #183, opened 2017-11-11
Optimize album art

whorfin:
I've tried a number of strategies, which often end up being undone by the SQL scheduler.
Without changing the database schema [additional complicated functional indices] this is the best I've got at the moment.
...I made a mistake at some point and had my database in a non-default state; this works as long as you've not messed around with PRAGMA

Perhaps it will help with #138 

Cheers

### DIFF
--- a/src/storage/sql_storage.cc
+++ b/src/storage/sql_storage.cc
@@ -1212,7 +1212,7 @@ String SQLStorage::incrementUpdateIDs(shared_ptr<unordered_set<int>> ids)
 }
 
 // id is the parent_id for cover media to find, and if set, trackArtBase is the case-folded
-// name of the track to try as artwork
+// name of the track to try as artwork; we rely on LIKE being case-insensitive
 String SQLStorage::findFolderImage(int id, String trackArtBase)
 {
     Ref<StringBuffer> q(new StringBuffer());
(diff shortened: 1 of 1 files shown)


v00d00:
Which DBs did you test with?

whorfin:
Only SQLite.
_should_ work fine with MySQL but I didn’t test

whorfin:
A note on these two additional commits:
I noticed that not all DLNA clients behave the same, even same client on different platform [ios vs android]
I found that 8player was working fine but foobar2000 was timing out.
These two changes further improved performance significantly.
They again _should_ work on MySQL

An additional table change
CREATE INDEX mt_cds_object_parent_type ON mt_cds_object(parent_id,object_type);
would speed some queries by as much as 3x; yes these fields are present in an existing index, but the scheduler is affected in a way that can matter.

But we can avoid a database change by leaving this alone [though CREATE INDEX ... IF NOT EXIST would also work easily]; this requested PULL works well and is defensible - there's no particular reason to prefer alphabetically first cover filename match when there are several.

whorfin:
this last bit helps some DLNA clients avoid a timeout
it makes eg 8player really really responsive at metadata-driven [ie "Genres" or "Albums"] lists

v00d00:
Lets merge this, and hopefully we get some feedback on MySQL.

PR #186, opened 2017-11-14
Improve consistent toast size for error messages

elmodaddyb:
When the system suffers a catastrophic error, the toast will show the entire message.  This causes elements to go off-screen and be unreadable.  This fix also applies to **tasks** which show at the top of the screen.

> There is presently no standard CSS way to easily do _multi-line_ wrapped ellipsis. (requires -webkit, etc)

### BEFORE
gerbera-toast-big-error

### AFTER
gerbera-fixed-width-toast

### DIFF
--- a/gerbera-web/mock-api/items/oversize-error-msg.json
+++ b/gerbera-web/mock-api/items/oversize-error-msg.json
@@ -1,0 +1,6 @@
+{
+  "success": false,
+  "error": {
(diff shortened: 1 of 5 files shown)


v00d00:
How do you see the full error?

elmodaddyb:
Yea, good point. I will work on something suitable to show the whole thing. I am thinking of a **expand** icon that resizes the toast to a max size with wrapping. 

On a side note, the Gerbera.io theme is really cool, so might try to mock up a zeplin view of what that could look like applied to the new UI.

PR #203, opened 2017-12-10
Fixes #201 - Remove row on click, instead of waiting for refresh

### LINKED ISSUE #201: Gerbera UI - Double clicking delete icon from items list causes 2 calls
If the user **double-clicks** the delete icon from the items list, 2 requests to delete are sent. This results in the second request producing a message `object id not found`. Object is successfully deleted.

elmodaddyb:
A simple change to manually remove the item table row `<tr>` on click...which removes possibility of double clicking, while waiting for items to refresh from server.


v00d00:
When are we getting the UI tests added into Travis?

elmodaddyb:
@v00d00 I will work on adding the UI tests to Travis, this week. The unit tests should be easy enough. The e2e tests might take some effort, with headless browser (xvfb, etc)

PR #233, opened 2018-01-06
Add support for Google Test

astylos:
Want to do some dev on gerbera but would like support for unit testing first. Included a couple of trivial tests just to show it working.

New to Git (still using SVN at work) and this is my first pull request. Content of dcf1c5d is not showing in the diff but is listed so if it does look like it would get merged (search_handler.cc/h and some other changes) then abort and let me know.  That's a clue as to what I'm wanting to implement...


elmodaddyb:
Way to go @astylos , this is awesome!!!

I too am on a google test kick!  I was able to get it to dynamically download googletest **when enabled**.  So users, don't need to install google test.

[https://github.com/elmodaddyb/gerbera/blob/shoutcast-radio/cmake/RetrieveGoogleTest.cmake](https://github.com/elmodaddyb/gerbera/blob/shoutcast-radio/cmake/RetrieveGoogleTest.cmake)

https://github.com/elmodaddyb/gerbera/tree/shoutcast-radio

I'll try to adapt this to your PR when its merged!!!

astylos:
I took the dynamic download approach at first but then I saw cmake had proper support for GoogleTest so went with this approach. I'm not fussed how it is done, just as long as I can run unit tests.

elmodaddyb:
nice, TDD for the win.

cweiske:
For anyone wondering what google test is: An open source unit testing library, https://en.wikipedia.org/wiki/Google_Test

Issue #238, opened 2018-01-13
MariaDB 10.1: Browsing broken


cweiske:
When using MariaDB 10.1 ([Default on Debian 9](https://mariadb.com/resources/blog/mariadb-server-default-debian-9)) as database, I cannot browse my library anymore with clients:

They get a HTTP error 501, "action failed".

The log says why:
```
INFO: Exception: Mysql: mysql_real_query() failed:
mysql_error (1235): "This version of MariaDB doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'";
query: SELECT `id` FROM `mt_cds_object` WHERE ( `dc_title`LIKE 'cover.jp%' OR `dc_title`LIKE 'albumart%.jp%' OR `dc_title`LIKE 'album.jp%' OR `dc_title`LIKE 'front.jp%' OR `dc_title`LIKE 'folder.jp%' ) AND `upnp_class`='object.item.imageItem' AND ( ( `parent_id` IN (SELECT `parent_id` FROM `mt_cds_object` WHERE `upnp_class`='object.item.audioItem.musicTrack' AND `id` IN (SELECT `ref_id` FROM `mt_cds_object` WHERE `parent_id`='564' AND `object_type`=2 LIMIT 100))     ) OR `parent_id`='564') LIMIT 1
```

This `LIMIT 100` was introduced in 6dfe3ee7b513116cf6a01f0e2e3620a121b82ac4 by @whorfin.
Reverting that commit makes it work again on MariaDB.

Also see https://mariadb.com/kb/en/library/subquery-limitations/

> MariaDB doesn't support LIMIT in subqueries with certain subquery operators.

whorfin:
I'd be tempted to say that means MariaDB isn't compatible with gerbera, then.
That LIMIT is important for performance with large collections, at least with current optimizers for sqlite

If it is helpful, I'd be happy to add a #define switch allowing that to be disabled, though rely on someone else to hook it up to detect MariaDB.
Is it only the subquery LIMIT which is unsupported?

whorfin:
[maybe a simpler way would be to simply allow setting MAX_ART_CONTAINERS to 0, and disable the LIMIT in that case]

astylos:
[...]

PR #239, opened 2018-01-14
Allow the subquery LIMIT to be disabled by -DMAX_ART_CONTAINERS=0

whorfin:
This should allow #238 to be addressed by appropriate cmake shenanigans


v00d00:
Can we use `#ifndef` instead?

As for the CMake part, how are we supposed to check this?

whorfin:
The idea was to disable in cmake configs with known incompatibility
Since it appears MySQL still doesn't support this, i'm going to retract the pull, and redo this to default to off.
[the mechanism was done so that it would be default enabled; i'll default disable it now]

PR #243, opened 2018-01-14
Only enable MAX_ART_CONTAINERS when sqlite3 driver is in use

whorfin:
Fix for  #238
"this time for sure"

Since it appears MySQL also still doesn't support subquery LIMITS, this is only used when the sqlite3 driver is active.  I realized this can't be safely determined strictly from cmake as ultimately this is a runtime choice from config files.

Perhaps someone running MySQL can investigate further - creating a subtable or otherwise purpose-optimizing this query would likely still be useful on mysql for performance.  Ultimately the cover art mechanism could use a refactor.

Cheers


v00d00:
I guess slower is better than not working at all 👍

PR #246, opened 2018-01-24
add ffmpeg/auxdata tags

oreojack:
I've been using custom import scripts with additional tags, and found the existing "pretend id3tag section is for ffmpeg" didn't work very well. With these changes I can use this in my config xml file:
```
     <library-options>
      <ffmpeg>
        <auxdata>
          <add-data tag="show"/>
          <add-data tag="season_number"/>
          <add-data tag="episode_sort"/>
```
(etc.)


v00d00:
This is great! Thanks :+1:

PR #248, opened 2018-01-28
Repeatable Sphinx build for documentation

elmodaddyb:
* Support python virtualenv for documentation build
* Check for virtualenv to avoid failures

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ mediatomb
 **/node_modules/
 **/.idea/
 **/coverage/
+**/_build
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,160 @@
+# Overview
+
+Gerbera uses [Sphinx Python Document Generator](http://www.sphinx-doc.org) to generate documentation.
+
+Below you will find instructions on how to:
+
+* Generate documentation
+* View documentation
+* Add to documentation
+* Publish documentation
+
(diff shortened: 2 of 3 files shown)


elmodaddyb:
This PR improves the sphinx build script to use a insulated `virtualenv` with python.  This is in preparation for the future features like:

1. Publish **docs** when gerbera builds in Travis CI.
2. Automate build of **docs** for Gerbera UI

I was going to cleanup the `/doc` directory but was not sure what is outdated? 

@v00d00 can you help with document status...is only `index.rst` valid?

I tested the script on the following systems **successfully**

* macOS High Sierra
* Fedora 27

elmodaddyb:
Improved iteration speed when developing locally by introducing a `--clobber` option which avoids removing/recreating a new `virtualenv` environment on every run.

Document building is very much a back/forth activity, so this helps tremendously.

v00d00:
Most of the readme.lyx is still correct, same for scripting.lyx and transcoding.lyx but they all need converting to rst and sphinx-ifying, then they can get removed.

elmodaddyb:
@v00d00 - thanks for the reply.  I will iterate over those documents(`readme`,`scripting`,`transcoding`), using the script..and see how it goes.

You can merge this PR if you prefer, or wait for the documents to be converted and included with this PR.  Likely I will work on the document conversion this weekend.

> I also have a full UI document nearly complete.

PR #259, opened 2018-03-04
Improve C++ Unit Test Framework

### LINKED ISSUE #258: Improve Test Framework for repeatable tests
As a developer, I would like to easily add C++ tests to Gerbera.

1. Do not require **Google Test** or **Google Mock** to be installed globally.
2. Template the **Google Test** implementation for easy additions.
3. Replicate test file arrangement across all tests.
4. Enable CMake testing via `make test`.
5. Enable individual testing using `ctest`.

elmodaddyb:
Fixes #258 

* Retrieve googletest automatically, implement repeatable test layout
* Enable CMake unit tests in TravisCI build
* Add boost-dev to TravisCI for unit testing
* Add CXX_FLAGS in TravisCI to compile gtest with clang++-4.0 properly


.travis.yml
```diff
@@ -37,14 +38,16 @@ matrix:
             - ubuntu-toolchain-r-test
           packages:
             - g++-4.9
+            - libboost-all-dev
```
elmodaddyb: The shoutcast feature will leverage **Boost** during unit test to write files to the `/tmp` directory.  This helps test the output of `config.xml` file for settings.

.travis.yml
```diff
@@ -37,14 +38,16 @@ matrix:
...
     - env: CXX=g++-5 CC=gcc-5
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
           packages:
             - g++-5
-    - env: CXX=clang++-4.0 CC=clang-4.0
+            - libboost-all-dev
+    - env: CXX=clang++-4.0 CC=clang-4.0 CXXFLAGS=-stdlib=libc++
```
elmodaddyb: `CXXFLAGS` is used in the `RetrieveGoogleTest.cmake` file to build googletest with the proper linker library.  This was only an issue in **clang++-4.0**

CMakeLists.txt
```diff
@@ -555,29 +555,10 @@ find_package (ZLIB REQUIRED)
...
-        add_executable(gerberatest ${gerberatestFILES} $<TARGET_OBJECTS:libgerbera>)
-        target_include_directories(gerberatest PRIVATE "${CMAKE_SOURCE_DIR}/src")
-        target_include_directories(gerberatest PRIVATE "${CMAKE_SOURCE_DIR}/test")
-        define_file_path_for_sources(gerberatest)
-
-        get_target_property(configuredLibs gerbera INTERFACE_LINK_LIBRARIES)
-        target_link_libraries(gerberatest gtest gmock_main ${configuredLibs})
-        add_test(NAME example_test COMMAND gerberatest)
-    endif()
+if(WITH_TESTS)
```
elmodaddyb: This **if** surrounds the sub-directory `/test`.  All `CMakeLists.txt` are children.  This allows the `/test` directory to be a self-contained system.

cmake/RetrieveGoogleTest.cmake
```diff
@@ -0,0 +1,50 @@
...
+# Enable ExternalProject CMake module
+include(ExternalProject)
+
+# Download and install GoogleTest
+ExternalProject_Add(
+        gtest
+        URL https://github.com/google/googletest/archive/master.zip
+        PREFIX ${CMAKE_CURRENT_BINARY_DIR}/gtest
+        CMAKE_ARGS
+          -DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}
```
elmodaddyb: `CMAKE_ARGS` needed during build of **gtest** to build under `clang++-4.0` properly.

test/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+include("RetrieveGoogleTest")
+include_directories (../src)
+find_package( Boost REQUIRED COMPONENTS filesystem )
+
+## Add tests below
+add_subdirectory(test_dictionary)
+add_subdirectory(test_runtime)
```
elmodaddyb: New test suites can be added easily by **add_subdirectory**

* Each sub-directory is a self contained test suite.

test/test_dictionary/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+add_executable(testdictionary
+        $<TARGET_OBJECTS:libgerbera>
```
elmodaddyb: Adding reference to the `libgerbera` defined in the parent **CMakeLists.txt**.  This allows any tests within here to utilize the library.

test/test_dictionary/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+        main.cc
+        test_dictionary.h
+        test_dictionary.cc
+        )
+
+include(DefFileName)
+define_file_path_for_sources(testdictionary)
+
+include_directories(
```
elmodaddyb: Add `include` and `linked` libraries for any testing needs.

> It seems simpler to add all requirements here, as it makes creating tests _rinse & repeat_

astylos:
Looks good to me, a better effort than mine which was really only done for my own benefit. This has a much better structure and I guess more easily integrated with an IDE.

elmodaddyb:
@astylos - let me know if the test structure supports the tests you have written so far (_sorry to require merge_).  I would like to make it most easy to use and migrate to.

> I did not test this on Visual Studio compiler, as I do not have Windows. :sob:

astylos:
@elmodaddyb I'll merge once this PR gets merged to master, but don;t see any issues.

I hope you wasn't thinking I use VS! :smile (emacs on Ubuntu)

.travis.yml
v00d00: I dont think we want a boost dependency just for the unit tests!

elmodaddyb: I agree, was kind of on the fence with the inclusion of boost, as it’s a pretty heavy weight library. 

I will remove, and rework the shoutcast tests.

elmodaddyb:
@v00d00 - Boost is no longer a unit test dependency as of 7ff5c98effeda1cca62023c270da8914d09d30fa

Thanks for reviewing!

.travis.yml
```diff
@@ -44,7 +44,7 @@ matrix:
             - ubuntu-toolchain-r-test
           packages:
             - g++-5
-    - env: CXX=clang++-4.0 CC=clang-4.0
+    - env: CXX=clang++-4.0 CC=clang-4.0 CXXFLAGS=-stdlib=libc++
```
v00d00: This shouldn't be needed?

https://github.com/gerbera/gerbera/blob/58b7f800f3b82e883257a4464e735564f9611aab/CMakeLists.txt#L286

cmake/RetrieveGoogleTest.cmake
```diff
@@ -0,0 +1,50 @@
...
+## See http://kaizou.org/2014/11/gtest-cmake/
+
+# We need thread support
+find_package(Threads REQUIRED)
+
+# Enable ExternalProject CMake module
+include(ExternalProject)
+
+# Download and install GoogleTest
+ExternalProject_Add(
```
v00d00: Does this work offline too?

Linux packages usually make the download set separate from the build phase.

elmodaddyb: Yea, valid point, unless you downloaded it prior...this would fail.

.travis.yml
elmodaddyb: Golly, I flew by that statement....For simplicity, I will roll the retrieve back to a standard `find_package(...)`  That will eliminate some complexities and solve the offline scenario.

elmodaddyb:
Sadly, I cannot seem to find a ubuntu trusty package which provides GTest.  It seems like others solved this by using the retrieval pattern previously implemented.

Any suggestions on GTest install approach for Travis CI?

v00d00:
What about https://packages.ubuntu.com/source/trusty/devel/gtest and gtest-dev

They provide the source which is what is needed, if I understand correctly?

Gentoo and Arch provide sources and the libs:
On gentoo there is a library https://packages.gentoo.org/packages/dev-cpp/gtest
Or arch https://www.archlinux.org/packages/community/x86_64/gtest/files/

elmodaddyb:
OK, I think I got the issues worked out.  Here is some background:

1. Google Test team does not recommend pre-compiled versions of the **gtest** source as it will enforce cpp flags from the original compilation..somewhat eliminating its purpose.
2. To this end....Ubuntu provides `libgtest-dev` which only installs the **sources**
3. To get GTest running on Ubuntu it needs to be compiled and installed manually

I am working through these steps for **travis-ci** as part of `before_install` and should have an update soon.

Other projects on github elected to use the [ExternalProject](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html) capability.  However as discussed this puts a big reliance on us maintaining it :-1:

elmodaddyb:
OK, I tried `libgtest-dev` from Ubuntu Trusty repository, but it was version 1.6.0 which had a few bugs with relation to **clang++-4.0** whereby it would still register includes for GCC.  These were fixed and available in 1.7.0.

The **GoogleTest** framework is downloaded/installed using a shell script.  This is same pattern as other build requirements.  This removes the dynamic compile within CMake scripts.

Let me know what you think?

Thanks,
Eamonn

elmodaddyb:
This PR is ready to rock and roll.  Let me know if you have suggestions for improvement!

Issue #227, opened 2018-01-05
Improved script debugging


tcrass:
Hi there,

I am currently facing the problem of a Gerbera installation perfectly scanning my video collection when starting up, but failing to properly reflect changes during run-time. For instance, when renaming or adding a video file while Gerbera is already running, I find a message "ERROR: Exception caught: Script: failed to execute script" in the system journal, but no further information _why_ the script's execution failed. Maybe I just missed the place where to increase Gerbera's debug level -- if not, I'd appreciate if generally some more detailed information could be written to the journal about the underlying cause of a script execution failure.

Apart from that, I'm *so* happy that MediaTomb lives on in the form of Gerbera -- I just *love* the possibility to create my own library layout through custom import scripts! :) (And have I mentioned already that I very much prefer the friendly Gerbera logo over the morbid scull of MediaTomb...?)

Cheers --

-- Torsten

elmodaddyb:
@tcrass - you can try to append `--debug` to the gerbera command and it runs in debug mode with verbose logs.

My guess is that duktape is running the import script, but having troubles.  The debug logs may help reveal the flaw.

tcrass:
Eamon,

seems I made a mistake when replying to the e-mail I received from you upon my initial posting I'll just re-post my replies here...

--------8<--------

Eamonn,

> @tcrass <https://github.com/tcrass> - you can try to append |--debug| to the gerbera command and it runs in debug mode with verbose logs.

thanks, that did the trick! I was expecting a debug level setting somewhere in the config.xml file, didn't think of the command line...

Cheers --

    Torsten 

--------8<-------

OK,

seems I'm running into trouble because the variable "object_script_path" is null when a change in a video file is detected and the import script gets called for the changed file. I must admit I just re-used a line of code involving that variable without understanding its meaning since I couldn't find any documentation about its purpose. So could you please clarify for me what data is supposed to be available in "object_script_path"? (It doesn't get set in import.js nor in common.js.)

Thanks --

    Torsten 

--------8<--------

OK, it seems object_script_path initially contains the "root" path to a media directory as specified in the <autoscan><directory>...</directory></autoscan> section of config.xml. Here's the result of an additional print(...)-statement I inserted into the addVideo(obj) function:

  JS: object_script_path = /mnt/data/movies

When a video file changes, I get:

  JS: object_script_path = null

So... is this a bug or a feature? ;)

Cheers --

    Torsten

elmodaddyb:
[...]

PR #264, opened 2018-03-17
#227 Pass rootpath to import.js on Inotify event

tcrass:
Dear gerbera maintainers,

I added an overload to the ContentManager::addFile method, allowing for passing a container's root path as additional parameter. Using this method from within the event handling code in AutoscanInotify::threadProc should fix the problem of undefined object_script_path values in import.js as described in https://github.com/gerbera/gerbera/issues/227.

Cheers --

Torsten


v00d00:
Looks good thanks!

PR #268, opened 2018-03-20
Replace zmm::StringBuffer with std::ostringstream

gburca:
The only remaining occurrences of zmm::StringBuffer are essentially in sql_storage.cc. Once that's also converted to use std::ostringstream,zmm::StringBuffer can be eliminated.

### DIFF
--- a/src/zmm/stringbuffer.cc
+++ b/src/zmm/stringbuffer.cc
@@ -35,6 +35,7 @@
 #include "memory.h"
 #include "strings.h"
 #include "exceptions.h"
+#include "atrailers_content_handler.h"
 
 using namespace zmm;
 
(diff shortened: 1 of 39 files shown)


src/zmm/stringbuffer.cc
```diff
@@ -35,6 +35,7 @@
 #include "memory.h"
 #include "strings.h"
 #include "exceptions.h"
+#include "atrailers_content_handler.h"
```
v00d00: Probably a bad import?

gburca: Yes.

v00d00:
Thanks @gburca!

I look forward to the removal of `zmm::stringbuffer` entirely soon!

PR #277, opened 2018-04-02
227 rootpath still missing in js

tcrass:
Dear gerbera maintainers,

I still encountered some problems regarding null values for object_script_path when renaming whole directories in inotify-watched sub-trees. I introduced some further code changes which seemed to have solved the problem.

Hope that helps!

Cheers --

    Torsten

### DIFF
--- a/src/content_manager.cc
+++ b/src/content_manager.cc
@@ -1318,8 +1318,6 @@ int ContentManager::addFile(zmm::String path, bool recursive, bool async,
 int ContentManager::addFile(zmm::String path, zmm::String rootpath, bool recursive, bool async,
     bool hidden, bool lowPriority, bool cancellable)
 {
-    //if (check_path(path, true))
-    //    rootpath = path;
     return addFileInternal(path, rootpath, recursive, async, hidden, lowPriority, 0, cancellable);
 }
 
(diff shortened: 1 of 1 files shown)


v00d00:
The second chunk makes sense, what is the removal of the check_path needed for?

tcrass:
> The second chunk makes sense, what is the removal of the check_path 
> needed for?

If a path to a directory is passed, we don't want the root path to be 
replaced with that directory path, do we?

Cheers --

	Torsten



-- 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
le alte Computer sowie Zubehör +++ Amiga +++ Apple II +++ At
ari +++ Commodore PET/VC20/C64/C128/C16/Plus4 etc. +++ Sincl
air +++ TI99/4 +++ alles, was nicht PC oder Mac ist +++ Samm
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

------------------------------------------------------------
Dr. rer. nat. Torsten Crass
Biochemiker, Bioinformatiker,
Software-Entwickler

Web-Design / IT-Service
http://www.tcits.de

Breite Lade 20
31275 Lehrte
Tel. 05132/8219177
http://www.tcrass.de

Issue #209, opened 2017-12-19
Config file example ?


Adirelle:
After compiling and installing gerbera, I tried to "touch /etc/gerbera/config.xml", as told in the doc, and it did not work. I had to google a bit to find an example configuratrion file. I am not fond of build an XML file from its XSD schema. Would it be possible to provide a default or example configuration file to help newcomers ?

v00d00:
Gerbera will generate you a default config on first run?

Adirelle:
That would be cool but maybe too much. However, a default configuration could be generated on build (like the systemd service file), or at least, an example could sit somewhere in the source, with most options provided with their default values.

I had to take http://p.cweiske.de/433 as a starting point.

v00d00:
I mean, the code already does generate you an example config on first run (or if it cant fine a config file).

I guess we could generate one at install time, but I dont see any advantage.

Adirelle:
Oh... This was what I supposed at first. But as I followed the instructions, I created an empty configuration file and gerbera complained it was invalid.

elmodaddyb:
Hello, 

As you noticed the systemd readme is a bit misleading with the `touch config.xml` example. 

That’s my doing 👎 my bad translation of **make sure it exists**

I will be writing up some ui docs soon and will clean it up too. 

If you make sure your `gerbera` runtime user has write access to the path, Gerbera will auto-generate the config for you, if it doesn’t exist.

Adirelle:
Ehh. I'd prefer the service not to have write access to its configuration file unless it is absolutely necessary. Follwoing that idea, maybe Gerbera could have an explicit flag (say -C) to send a default configuration to stdout. This would allow to create it without giving the daemon too much privileges. E.g.

```bash
# Generate the default configuration
sudo mkdir -m 0755 -p /etc/gerbera
gebera -C | sudo tee /etc/gerbera/config.xml
```

elmodaddyb:
Hello, I took a look at the Gerbera code and it is reasonable to make a utility method which generates a default configuration file.  This actually happens today, but the file is written out to file-system instead of standard-out.

Is the feature still desirable?  I will pull something together by leveraging:

```
String ConfigManager::createDefaultConfig(String userhome)
```

1. No more default generation of `config.xml` (_this might be an issue for quick start_)
2. Runtime option to generate the `config.xml`

What are your thoughts?

elmodaddyb:
[...]

Issue #196, opened 2017-12-06
Should not store UDN in config file.


v00d00:
We currently generate and then write back a UDN if its not in the config which means we have to have write access to our config file, which is frowned upon.
See https://bugs.gentoo.org/632122

elmodaddyb:
Yep, UDN generation should be easy enough to migrate to the `ConfigGenerator`

Happenstance, I used the UUID generation to generate a test folder for testing read/write of the config.xml when running unit tests!! 😖 

> I will include UDN in the PR, likely be ready this weekend.

elmodaddyb:
@v00d00 - I plan to completely remove modification capability, which will include remove of...

`ConfigManager.h`
```c++
void save();
void save_text(zmm::String filename, zmm::String content);
void migrate();
void prepare_udn();
```
The `validate` method remains active.

> Do you agree?

elmodaddyb:
All set...

https://github.com/elmodaddyb/gerbera/blob/config-generator/src/config/config_generator.cc#L416

elmodaddyb:
@v00d00 - do you think this bug is solved, or is it pending a new gentoo build?

v00d00:
In all honesty I thought this already closed!

PR #282, opened 2018-04-08
Implement Configuration Generator

elmodaddyb:
This feature implements a separate generation mechanism for the Gerbera configuration system.  The list below outlines the changes.

### Fixes Issues
#209 - Config file example
#196 - Should not store UDN in config file

### New Features
* Add new command line option `--create-config` to generate configuration
* Move configuration generation to separate class implementation
* Generate UDN when generating `config.xml`

### Removed Features
* Remove automatic generation of `config.xml` during runtime
* Remove configuration migration capability

> This change affects how easily users can startup gerbera.  It puts ownership onto the user by removing automated generation of configuration.

All code was written using TDD so the functional coverage is adequate.  I ran the system through Valgrind and eliminated all possible memory leaks when running `--create-config`.  This is my first sizable implementation of C++.

Honest reviews welcome.  If there are incorrect c++ patterns please do let me know.

Thanks,

Eamonn


.travis.yml
```diff
@@ -26,7 +26,7 @@ before_install:
 script:
   - mkdir build
   - cd build
-  - cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DWITH_SYSTEMD=0 -DWITH_TESTS=1 .. && make && make test
+  - cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DWITH_SYSTEMD=0 -DWITH_TESTS=1 .. && make && make test ARGS=-V
```
elmodaddyb: This add verbose output of unit tests.  It is helpful when the test fails to know the exact failure.

config/config2.xsd
```diff
@@ -149,11 +149,46 @@
     <xs:element name="extended-runtime-options">
         <xs:complexType>
             <xs:all>
+                <xs:element ref="ffmpegthumbnailer" minOccurs="0"/>
                 <xs:element ref="mark-played-items" minOccurs="0"/>
             </xs:all>
         </xs:complexType>
     </xs:element>
 
+    <xs:element name="ffmpegthumbnailer">
```
elmodaddyb: When producing the fixtures to test, there was several missing XSD elements.  Cleanup and implement accurate XSD.

src/config/IConfigGenerator.h
```diff
@@ -0,0 +1,36 @@
...
+/// \file IConfigGenerator.h
+
+#ifndef GERBERA_ICONFIGGENERATOR_H
+#define GERBERA_ICONFIGGENERATOR_H
+
+class IConfigGenerator {
+
+ public:
+  virtual ~IConfigGenerator() {}
+  virtual std::string generate(std::string userHome, std::string configDir, std::string prefixDir, std::string magicFile) = 0;
```
elmodaddyb: In the future we should eliminate `mxml` implementation and use a standard XML library.  This class begins to insulate the XML generation and expose standard signatures.

src/config/config_generator.cc
```diff
@@ -0,0 +1,449 @@
...
+Ref<Element> ConfigGenerator::generateServer(std::string userHome, std::string configDir, std::string prefixDir) {
+  Ref<Element> server(new Element(_("server")));
+
+  server->appendElementChild(generateUi());
+  server->appendTextChild(_("name"), _(PACKAGE_NAME));
+
+  server->appendElementChild(generateUdn());
+
+  std::string homepath = userHome + DIR_SEPARATOR + configDir;
+  server->appendTextChild(_("home"), String::copy(homepath.c_str()));
```
elmodaddyb: This line had a memory leak.  By passing `std::string` into the function, but then returning the `Ref<Element>`.  The  **std::string** value is not used until later in the **generate** method.

I chose the `Ref<String>` implementation of `String::copy` which duplicates the string.  I previously tried using `strdup(...)` but this creates a `char*` which needs to be **free'd** later.

Any insight or suggestions here would be appreciated.

src/config/config_generator.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "IConfigGenerator.h"
+
+class ConfigGenerator : public IConfigGenerator {
+ public:
+  ConfigGenerator();
+  ~ConfigGenerator();
+
+  std::string generate(std::string userHome, std::string configDir, std::string prefixDir, std::string magicFile) override;
+  zmm::Ref<mxml::Element> generateServer(std::string userHome, std::string configDir, std::string prefixDir);
```
elmodaddyb: I looked for a suitable XML DOM generation in C++ but could not find anything easier to implement than the current `mxml::Element`.  Ideally this method will have standard library for XML and expose only the **generate** method.

src/config_manager.cc
```diff
@@ -106,30 +107,18 @@ void ConfigManager::init()
...
 
     String home = userhome + DIR_SEPARATOR + config_dir;
-    bool home_ok = true;
 
     if (filename == nullptr) {
         // No config file path provided, so lets find one.
-        if (home_ok && (!check_path(userhome + DIR_SEPARATOR + config_dir + DIR_SEPARATOR + DEFAULT_CONFIG_NAME))) {
-            home_ok = false;
-        } else {
+        if (check_path(home + DIR_SEPARATOR + DEFAULT_CONFIG_NAME)) {
```
elmodaddyb: All we do now is:

1. Check if the default path is OK
2. Load default path OR throw exception

src/config_manager.cc
```diff
@@ -106,30 +107,18 @@ void ConfigManager::init()
...
     }
 
-    if (filename == nullptr) {
-        throw _Exception(_("\nThe server configuration file could not be found in ~/.gerbera\n") + "Gerbera could not determine your home directory - automatic setup failed.\n" + "Try specifying an alternative configuration file on the command line.\n" + "For a list of options run: gerbera -h\n");
-    }
-
     log_info("Loading configuration from: %s\n", filename.c_str());
     load(filename);
-
-    prepare_udn();
```
elmodaddyb: Moved to `config_generator.cc`

src/main.cc
```diff
@@ -236,20 +240,16 @@ For more information visit " DESC_MANUFACTURER_URL "\n\n");
...
+                create_config = true;
+                break;
             }
             break;
         default:
             break;
         }
     }
 
-    if (print_version) {
```
elmodaddyb: Moved this **below** the `--create-config` execution.  This allows `--create-config` to generate cleanly.

test/test_config/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+add_custom_command(TARGET testconfig POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/fixtures $<TARGET_FILE_DIR:testconfig>/fixtures
+)
+
+add_test(NAME testconfig
+        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/test/test_config
+        COMMAND ./testconfig)
+
+add_definitions(-DCMAKE_BINARY_DIR="${CMAKE_BINARY_DIR}")
```
elmodaddyb: I added the `CMAKE_BINARY_DIR` to allow the tests to write files to the build/test directory.  This allows for more robust testing.

test/test_config/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+        ${MYSQL_CLIENT_LIBS}
+        ${SQLITE3_LIBRARIES}
+        ${ICONV_LIBRARIES}
+        ${GTEST_LIBRARIES}
+        ${GERBERA_INTERFACE_LIBRARIES}
+        ${CMAKE_THREAD_LIBS_INIT}
+        )
+
+add_custom_command(TARGET testconfig POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/fixtures $<TARGET_FILE_DIR:testconfig>/fixtures
```
elmodaddyb: Copy all the **fixture** files to the build directory, thus allowing easy compare within GoogleTest

test/test_config/test_configgenerator.cc
```diff
@@ -0,0 +1,226 @@
...
+};
+
+#if defined(HAVE_FFMPEG) && defined(HAVE_FFMPEGTHUMBNAILER) && defined(HAVE_MYSQL) && defined(HAVE_MAGIC) && defined(HAVE_JS) && defined(ONLINE_SERVICES)
+TEST_F(ConfigGeneratorTest, GeneratesFullConfigXmlWithAllDefinitions) {
+  std::string mockXml = mockConfigXml("fixtures/mock-config-all.xml");
+
+  std::string result = subject->generate(homePath, configDir, prefixDir, magicFile);
+
+  // remove UUID, for simple compare...TODO: mock UUID?
+  std::regex reg("<udn>uuid:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}</udn>");
```
elmodaddyb: Now that `udn` is generated, it is difficult to test.  I simply removed the value and assert `<udn/>`

The **udn** is tested independently.

test/test_config/test_configmanager.cc
```diff
@@ -0,0 +1,134 @@
...
+class ConfigManagerTest : public ::testing::Test {
+
+ public:
+
+  ConfigManagerTest() {};
+
+  virtual ~ConfigManagerTest() {};
+
+  virtual void SetUp() {
+    gerberaDir = createTempPath();
```
elmodaddyb: `ConfigManager` requires a filesystem file to be initialized.  It also checks the validity of the script files referenced in config.

Initially was using **boost** library to generate temporary dir and files, but its a bit heavy weight to include just for testing.

test/test_config/test_configmanager.cc
```diff
@@ -0,0 +1,134 @@
...
+  String config_file;
+  std::string home;
+  std::string confdir;
+  std::string prefix;
+  std::string magic;
+  ConfigManager *subject;
+};
+
+TEST_F(ConfigManagerTest, LoadsWebUIDefaultValues) {
+  subject->setStaticArgs(config_file, _(home.c_str()), _(confdir.c_str()), _(prefix.c_str()), _(magic.c_str()));
```
elmodaddyb: This test can be extended to test any of the configuration elements.  I plan to use the same to test the **SHOUTcast** capability.

test/test_server/test_main.cc
```diff
@@ -0,0 +1,85 @@
...
+  ServerTest() {};
+  virtual ~ServerTest() {};
+
+  virtual void SetUp() {
+  }
+
+  virtual void TearDown() {
+  };
+
+  std::string exec(const char* cmd) {
```
elmodaddyb: Here we are actually launching the `gerbera` binary with the additional options to test its output.

elmodaddyb:
When ready, we can **squash & merge** these revisions for simplicity.

v00d00:
Looks good!

I do wonder if there is anything to gain anything by creating `IConfigGenerator` for a single method with a single implementation?

elmodaddyb:
Yeah, I agree... I guess it was more an exercise in c++ learning for me. 

I will remove the abstract class `IConfigGenerator`

elmodaddyb:
@v00d00 - I removed the `IConfigGenerator` abstract class.

Let me know if you have any other suggested improvements.

Thanks,

Eamonn

v00d00:
Thanks! Merged, I squashed locally so that it kept your name on the commit.

PR #300, opened 2018-05-12
Fix atrailers build

ffontaine:
i2i function is used in atrailers_service.cc but this function is
defined only if defined(HAVE_JS) || defined(HAVE_TAGLIB) ||
defined(YOUTUBE) || defined(HAVE_LIBMP4V2) as a result compilation
fails if HAVE_CURL is set but HAVE_JS and HAVE_TAGLIG are not.
As youtube and libmp4v2 support have been dropped, replace those by
ATRAILERS in this list.

### DIFF
--- a/src/string_converter.cc
+++ b/src/string_converter.cc
@@ -226,7 +226,7 @@ Ref<StringConverter> StringConverter::p2i()
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


v00d00:
Nice, thanks!

PR #304, opened 2018-05-20
fix a couple errors in forSqlUpdate and forSqlDelete

compguy284:
Found this problem when I was using my PS3 to play my media content.

### DIFF
--- a/src/storage/sql_storage.cc
+++ b/src/storage/sql_storage.cc
@@ -2581,8 +2581,8 @@ std::shared_ptr<std::ostringstream> SQLStorage::sqlForInsert(Ref<CdsObject> obj,
 std::shared_ptr<std::ostringstream> SQLStorage::sqlForUpdate(Ref<CdsObject> obj, Ref<AddUpdateTable> addUpdateTable)
 {
     if (addUpdateTable == nullptr || addUpdateTable->getDict() == nullptr
-        || (addUpdateTable->getTable() == _(METADATA_TABLE) && addUpdateTable->getDict()->size() != 1))
-        throw _Exception(_("sqlForDelete called with invalid arguments"));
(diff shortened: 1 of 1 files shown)


elmodaddyb:
@compguy284 - Great Job with this issue... I will try to run 1.3.0alpha against a PS3 UPnP media render to see if similar bad issues occur.

> I don't yet understand 2 vs 1 in this case

compguy284:
If tablename is "mt_metadata" and dictionary size != 1 then it would it throw an exception and cause problems on the PS3 with mark-played-items enabled on first play of media.

elmodaddyb:
Hello All,  

I figured out the relationship of this issue.  Prior to d04736dc73dae3c78968029845363fa65f0c3d51 the system established an array of `1` for the updates to the `CDS_OBJECT_TABLE` when an item is marked as **played**.  So the storage update verification in the PR was previously checking that the update statement was valid with a count of `1`.

After the Meta data table inclusion, the system pushes `2` updates when an item is played as per:

[sql_storage.generateMetadataDBOperations](https://github.com/gerbera/gerbera/blob/master/src/storage/sql_storage.cc#L321)

As a result the **verification** step needs to say `!=2` as per @compguy284  :1st_place_medal: 

I think the verification may also fail when the metadata size > 1 `(e.g. OBJ table(1) + META table(1))`

[sql_storage -- returnValueSize](https://github.com/gerbera/gerbera/blob/master/src/storage/sql_storage.cc#L296)

I am not yet sure if this is a valid scenario....so might just be conjecture....
```
int updateSize = 1 +  (obj->getMetadata() == nullptr) ? 0 : obj->getMetadata()->size();
addUpdateTable->getDict()->size() != updateSize
```

If the `mark-list` includes only `video` then perhaps the meta-data count will always be == 1.

> Hope my ramblings shed some light on the solution....

PR #306, opened 2018-05-24
Migrate mock-api to ES6, DRY up mocks

elmodaddyb:
Follow up with additional javascript updates:

* Convert **mock-api** to ES6
* Remove duplicate **mock-api** functions
* Organize code

> This code is not part of the end-user deliverable, so ES6 is preferred.


elmodaddyb:
Wow, `MockReponder`  gosh that's a bad spelling...let me fix that.

Issue #111, opened 2017-07-20
UE40KU6000UX Samsung TV not detecting a working Gerbera server 


Stepperer:
Server itself is working fine and I can see it on my PC and PS4. But my TV cannot. What can be done?

asavah:
AFAIK in mediatomb there was a workaround for samsung tvs via a custom header.
```<custom-http-headers>
        <!-- Samsung needs it -->
        <add header="transferMode.dlna.org: Streaming"/>
        <add header="contentFeatures.dlna.org: DLNA.ORG_OP=01;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01500000000000000000000000000000"/>
    </custom-http-headers>
```
If memory serves currenly in gerbera or libupnp there is no support for custom headers.
https://github.com/v00d00/gerbera/blob/master/src/server.cc#L171

v00d00:
libupnp 1.8 should have headers support we need to implement this.

v00d00:
mrjimenez/pupnp#46 might be useful here.

hollaho:
Just stumbled into the same roadblock. Makes gerbera currently unusable with Samsung TVs.

Yes, the crappy implementation of the Samsung TVs is also to blame. But doesn't help. The fat elephant Samsung wont move, so we need to fix it on this side... Hopefully it is fixed soon.

v00d00:
@Smartsmurf that patch looks really nice! Thanks!

Perhaps you can open a pull request?

I do have a older Samsung, but its only wired network, which is a problem to test with at the moment.

The link to libupnp patch seems to be the same? You will possibly want to open a PR there too https://github.com/mrjimenez/pupnp

Smartsmurf:
@v00d00 thanks for the hint. I corrected the libupnp link.

b4r0z:
Sorry I don’t understand how to patch libupnp. What file needs editing? I have a Samsung tv and files are listed but nothing can be played (unsupported file) while it used to work with mediatomb. Everything is playing on the phone so gerbera (1.1 because 1.3 compile but the web interface wont work) does its job.

elmodaddyb:
Hello @b4r0z - I remember we had a bug #218 - where UI would not work because the ARM chip architecture had problems with the UI's session ID.  Are you running Gerbera server on an ARM system?

Can you elaborate on what error you get on the web UI?

b4r0z:
[...]

PR #307, opened 2018-05-24
Improve compatibility with Samsung DTVs via DLNA

Smartsmurf:
Added contentFeatures.dlna.org field and fixed lengths of D_TR_FLAGS*

Discussion of this patch was conducted here:
https://github.com/gerbera/gerbera/issues/111


v00d00:
Thanks!

PR #311, opened 2018-06-03
Fix #310 - Config Generator to use native BSD UUID when available

### LINKED ISSUE #310: FreeBSD 11.1 "make" installation step can't find uuid/uuid.h
Installing ossp-uuid doesn't help.

However, it compiles and works if `/usr/include/uuid.h` is copied to `/usr/include/uuid/uuid.h`.

elmodaddyb:
The method code **does** use native UUID, but the import did not.


v00d00:
Ahh, excellent!

Issue #296, opened 2018-04-30
Issue with importing metadata


chiraag-nataraj:
I am currently using Debian sid/experimental (so gerbera version 1.1.0 at present) and I'm having trouble getting gerbera to import my music's metadata correctly.

1. My metadata is in various scripts (Latin, Bangla, Devanagari, Kannada, Telugu, Tamil, etc). I thought this might be the problem, but setting the metadata charset to UTF-8 didn't change anything.
2. I thought this might be the same issue as #26, but I have a new enough version of TagLib (1.11.1), so I don't think that's the problem.
3. _However_, the issue manifests itself in the same way as the above issue - that is, certain metadata is missing when I look at the debug output.

I'm probably doing something wrong, but I have no idea what.

Edit: attaching output from running `gerbera` with debugging output and then adding a folder to the database.
[gerbera.log](https://github.com/gerbera/gerbera/files/1961688/gerbera.log)

v00d00:
Hi @chiraag-nataraj, thanks for reporting!

I took a look at your log, and it looks like the taglib handler is never called?

```
2018-04-30 12:13:55    INFO: Adding /home/chiraag/Music/various/Classical_Songs_From_Films/
2018-04-30 12:13:55   DEBUG: [src/content_manager.cc:2004] run(): running add file task with path /home/chiraag/Music/various/Classical_Songs_From_Films/ recursive: 1
2018-04-30 12:13:55   DEBUG: [src/singleton.cc:69] registerSingleton(): registering new singleton... - 6 -> 7
2018-04-30 12:13:55   DEBUG: [src/content_manager.cc:816] addRecursive(): Checking path /home/chiraag/Music/various/Classical_Songs_From_Films/
2018-04-30 12:13:55   DEBUG: [src/content_manager.cc:834] addRecursive(): IS TASK VALID? [1], taskoath: [/home/chiraag/Music/various/Classical_Songs_From_Films/]
2018-04-30 12:13:55   DEBUG: [src/metadata/ffmpeg_handler.cc:222] fillMetadata(): Running ffmpeg handler on /home/chiraag/Music/various/Classical_Songs_From_Films//13_मीठे_बोल_बोले.flac
2018-04-30 12:13:55   DEBUG: [src/metadata/ffmpeg_handler.cc:117] addFfmpegMetadataFields(): Identified metadata track: 1141364544
[...]

PR #313, opened 2018-06-06
Add audio/flac to relevant places

chiraag-nataraj:
I think this should work? `gerbera --create-config` seems to show the right value. Fix for #296

### DIFF
--- a/config/mappings.xml
+++ b/config/mappings.xml
@@ -24,6 +24,7 @@
         <treat mimetype="audio/mpeg" as="mp3"/>
         <treat mimetype="application/ogg" as="ogg"/>
         <treat mimetype="audio/x-flac" as="flac"/>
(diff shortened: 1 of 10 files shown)


elmodaddyb:
Can you update the unit test fixtures to match XML?

elmodaddyb:
Looks good.  Any thoughts on the testing approach.  Its been in place for a bit and wondering if you found it easy or troubling to know where to update?

> Any thoughts?

chiraag-nataraj:
> Looks good. Any thoughts on the testing approach. Its been in place for a bit and wondering if you found it easy or troubling to know where to update?

The tests weren't too bad to update. The only reason I needed a second commit was that for some reason a tab was inserted in one of the tests instead of a space (which I caught only when the test failed).

v00d00:
Looks good, thanks!

Issue #308, opened 2018-05-27
transcoding of ogg not working under ubuntu 18.04


cot-al:
it was working fine with mediatomb under 17.10, but since I changed it is not working anymore. yet I am using teh same config.xml, vlc and ogg123 are working

v00d00:
Can you post a lot with debug log level?

cot-al:
Yes of course just let me know how to get this log

elmodaddyb:
Hello,

I made some progress on this issue, although encountered the same issue as @cot-al .

1. Installed latest `bionic` release of ubuntu onto docker
2. Compiled latest `gerbera`, everything works OK

3. Enable **transcoding**
   ```xml
    <profile name="oggflac2raw" enabled="yes" type="external">
        <mimetype>audio/L16</mimetype>
        <accept-url>no</accept-url>
        <first-resource>yes</first-resource>
        <accept-ogg-theora>no</accept-ogg-theora>
        <agent command="ogg123" arguments="-d raw -o byteorder:big -f %out %in"/>
        <buffer size="1048576" chunk-size="131072" fill-size="262144"/>
      </profile>
   ```

### Problems
Transcoding is working OK, but the `ogg123` command line tool is the problem.  When outputting `RAW` format, the `audio/ogg` file is somehow transcoded into an **image** type.

I tested locally, just to eliminate **gerbera**

```
ogg123 -v -d raw -o byteorder:big -f test.pcm /development/gerbera/ogg/test.ogg
```
> this produces complete garbage file, when played is video feedback

I tried this on both Fedora & Ubuntu, and results are the same.

### Summary
It looks like `ogg123` is not actually producing a **RAW** audio file.  Checking the `vorbis-tools` repository looks like it hasnt seen active development for years.

> Perhaps we need to find a suitable replacement for `ogg123`

It is strange that `ogg123` is not working...but there really isnt much conversation about this utility since early 2000.

elmodaddyb:
Hello,

I was able to stream an OGG file from Gerbera using the transcoding profile with **vlc**
```xml
<transcoding enabled="yes">
  <mimetype-profile-mappings>
    <transcode mimetype="audio/ogg" using="oggflac2raw"/>
  </mimetype-profile-mappings>
  <profiles>
    <profile name="oggflac2raw" enabled="yes" type="external">
      <mimetype>audio/mpeg</mimetype>
      <accept-url>yes</accept-url>
      <first-resource>yes</first-resource>
[...]

PR #316, opened 2018-06-10
Update HTTP Response to solve protocol issue with transcoding audio

elmodaddyb:
When running a transcode profile the HTTP protocol misses a `/r/n` on the `Content-Disposition` header when transcoding audio types.  This causes transcoded audio to fail loading in the media renderer.

## HTTP Protocol Helper

Externalize a method that adds the `\r\n` characters if needed to the set of custom headers added by Gerbera.  This handles both scenarios where the termination character **does not** exist, and where the termination character **already exists**

> Transcoded audio would not load into PS3/PS4 Media player in any condition.  This still needs more work.

## Update default OGG Transcoding Profile

Since `ogg123` was not working properly, updated the default configuration to include `ffmpeg` version with `ogg2mp3` profile.


elmodaddyb:
This fixes #308

PR #320, opened 2018-06-26
fix headers gluing like transferMode.dlna.org: StreamingCONNECTION: c…

crodar33:
…lose


elmodaddyb:
We faced similar header issues in #316. Many thanks for additional fixes!!

crodar33:
in #316 there is a solution to the problem, so my request is not relevant.

elmodaddyb:
@crodar33 - #316 is merged to master, hopefully solves your problem.  There were a few header logical flows through the code where it **would not** add the `/r/n` and some flows where it did.

I can work to fix up the HTTP protocol if the issue persists.

PR #326, opened 2018-07-07
Update Unit Tests to ES6/7

elmodaddyb:
### Update unit tests to ES6/7

* `let` and `const` instead of `var`
* `async + await` instead of `.then(...)`
* `() =>` instead of `function ()`
* Terminate all lines with `;`

### DIFF
--- a/gerbera-web/.eslintrc
+++ b/gerbera-web/.eslintrc
@@ -5,6 +5,7 @@
     "jquery": true,
     "es6": true,
     "mocha": true,
+    "jasmine": true,
     "node": true
   },
   "extends": "eslint:recommended",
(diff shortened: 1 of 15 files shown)


elmodaddyb:
@v00d00 - What do you think about using ES6/7 for client-side code for the Web UI?

https://kangax.github.io/compat-table/es6/

Duktape does not have much support yet for ES6+ so that needs to stay ES5.

v00d00:
Yay for semicolons!

As long as there is good compatibility I see no reason not use use es6 on the client-side also.

Martii:
Sure would like to know the advantage, other than others being extremely lazy in typing, to using arrows for function declarations. Unless there's a hidden usage it's definitely not an easy switch to make and in the synapses I'm always converting it back. :smile_cat:

elmodaddyb:
Yea, I agree that arrows doesn’t add much value. If only to be up to latest spec capability. 

Here is a summary of the capability difference of arrows function

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

PR #330, opened 2018-07-18
Improve accuracy of error message for missing config

### LINKED ISSUE #329: Config file not found : misleading exception message
When a config file is not found, an exception is raised with a static message :
> The server configuration file could not be found in ~/.config/gerbera

This message is misleading when using options to configure the location of the configuration file, like `--cfgdir` for instance.

The message should be dynamic.
https://github.com/gerbera/gerbera/blob/master/src/config_manager.cc#L116

elmodaddyb:
Utilize `home` variable and improve inline message format using `std::ostringstream`

Fixes #329 

**Example**
```
The server configuration file could not be found in: ~/.config/gerbera
Gerbera could not find a default configuration file.
Try specifying an alternative configuration file on the command line.
For a list of options run: gerbera -h
```


v00d00:
Nice!

Issue #193, opened 2017-11-23
FFmpegthumbnailer not works


MrDini123:
Hi again,

Now I wanted to enable the FFmpegthumbnailer funcionality in the config file, like this:

```
      <ffmpegthumbnailer enabled="yes">
        <thumbnail-size>128</thumbnail-size>
        <seek-percentage>10</seek-percentage>
        <filmstrip-overlay>yes</filmstrip-overlay>
        <workaround-bugs>no</workaround-bugs>
        <image-quality>8</image-quality>
      </ffmpegthumbnailer>
```

But it doesn't seems to do anything.

In turn, I compiled an FFmpeg 3.4 and FFmpegthumbnailer 2.2.0 just for this purpose (to use with gerbera). :)

Is there a way to see all the available config options with "what these actually does" comments/documentations?

Here is my gerbera binary's ldd output:

```
	libpthread.so.0 => /ffp/lib/libpthread.so.0 (0x4001a000)
	libiconv.so.2 => /ffp/lib/libiconv.so.2 (0x40035000)
	libuuid.so.1 => /ffp/lib/libuuid.so.1 (0x4011c000)
	libupnp.so.10 => /ffp/lib/libupnp.so.10 (0x40127000)
	libixml.so.10 => /ffp/lib/libixml.so.10 (0x40156000)
	libexpat.so.1 => /ffp/lib/libexpat.so.1 (0x40166000)
	libsqlite3.so.0 => /ffp/lib/libsqlite3.so.0 (0x4018e000)
	libduktape.so.202 => /ffp/lib/libduktape.so.202 (0x4023e000)
	libcurl.so.4 => /ffp/lib/libcurl.so.4 (0x40281000)
	libtag.so.1 => /ffp/lib/libtag.so.1 (0x402e9000)
	libmagic.so.1 => /ffp/lib/libmagic.so.1 (0x403c3000)
	libavcodec.so.57 => /ffp/lib/libavcodec.so.57 (0x403e7000)
	libavformat.so.57 => /ffp/lib/libavformat.so.57 (0x416de000)
	libavutil.so.55 => /ffp/lib/libavutil.so.55 (0x418d5000)
	libswscale.so.4 => /ffp/lib/libswscale.so.4 (0x41965000)
	libvorbis.so.0 => /ffp/lib/libvorbis.so.0 (0x419d9000)
	libvorbisenc.so.2 => /ffp/lib/libvorbisenc.so.2 (0x41a0c000)
[...]

PR #336, opened 2018-08-15
Adds Resource Attributes to scripting.

mayanez:
Looks like this was a pending todo in the code. Primarily useful for dealing with Stereo & Multichannel audio in your collection.

Only part I was not too sure about was `res = obj->getResource(0);`. I only found one resource to ever be used and didn't know how key collisions should be handled so left it as is. 

Let me know if I should change it.

### DIFF
--- a/scripts/js/import.js
+++ b/scripts/js/import.js
@@ -91,6 +91,19 @@ function addAudio(obj) {
     // comment the following line out if you uncomment the stuff above  :)
     var track = '';
 
+// uncomment this if you want to have channel numbers in front of the title
+/*
+    var channels = obj.res[R_NRAUDIOCHANNELS];
+    if (channels) {
+        if (channels === "1") {
+            track = track + '[MONO]';
+        } else if (channels === "2") {
+            track = track + '[STEREO]';
+        } else {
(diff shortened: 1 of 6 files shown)


elmodaddyb:
Hello @mayanez - Great contribution.

My understanding of the multiplicity of resources is related to the content that is being served by the UPNP server.  For example, the system can serve up a music resource as the `0th` and a cover art image as the `1th`.  Then a media render can play the music and show the cover art accordingly.

You can see how this is done in the code for thumbnail resources:

https://github.com/gerbera/gerbera/blob/master/src/cds_resource_manager.cc#L116

Take a look at the output XML from the server here:

https://github.com/gerbera/gerbera/issues/193#issuecomment-346905325

The multiplicity is helpful when sending transcoded content which might be of different format from the original. See below:

https://github.com/gerbera/gerbera/blob/master/src/cds_resource_manager.cc#L265

I think that many resource elements in the XML allows the the UPNP server to provide various resources and leaves it up to the media renderer to choose how & what to render.

> Long explanation for, I think using the `0th` is an OK assumption in majority of cases.

PR #346, opened 2018-08-31
Update install.rst

lukesoft76:
I have added the installation for CentOS.
The location of the RPM files for installation is given.
All files are hosted on github since i don't run a repository which is accessible from the WWW.

Regards,
Luke


v00d00:
Thanks!

PR #355, opened 2018-09-29
Use latest Google Test Release v1.8.1 - supporting pre C++11

### LINKED ISSUE #354: Fix GoogleTest - to continue pre C++11 compilation
Hello,

Gerbera uses the GoogleTest framework, and downloads the master branch from GitHub for TravisCI builds.  The recent GoogleTest master removed a few support items for variadic templates in pre C++11 compilers.

Since gerbera attempts to build with g++4.9 it fails.

### Option 1 
Force all build types to support C++11

### Option 2
Utilize GoogleTest v1.8.1 instead of master to stabilize build.

elmodaddyb:
GoogleTest has began to drop support for pre C++11 compilers.  This caused recent failure in Gerbera build.

GoogleTest v1.8.1 is the last release to support pre C++11 compilation.

This pull request updates install scripts to use `googletest-release-1.8.1` to stabilize build.

Fixes #354

### DIFF
--- a/scripts/install-googletest.sh
+++ b/scripts/install-googletest.sh
@@ -8,13 +8,13 @@ fi
 ##
 pwd=$(pwd)
 unamestr=$(uname)
-if [ -f "$pwd/master.zip" ]; then rm -f "$pwd/master.zip"; fi
-if [ -d "$pwd/googletest-master" ]; then rm -Rf "$pwd/googletest-master"; fi
+if [ -f "$pwd/release-1.8.1.zip" ]; then rm -f "$pwd/release-1.8.1.zip"; fi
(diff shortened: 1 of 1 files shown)


elmodaddyb:
We had a good run of stable builds.  This brings the build back to green.  The proper solution is to update Gerbera to follow along with latest C++ tech! 👔 

> I will do some research to see future-ready improvements.

v00d00:
We say we need a c++14 compiler anyway right? Do we just need to make it more obvious?

elmodaddyb:
Found a better way to set the `std=c++14` when compiling GoogleTest master.  Will submit PR for this instead.

PR #371, opened 2018-11-08
Add Server Name to Web UI Menu & Title

### LINKED ISSUE #363: UI should indicate server <name>
I have multiple instances of gerbera running on my server, hosting different kinds of media (adult musik, childrens audio books, adult movies, childrens movies) to be able to select a specific library (listening to random music, not suddenly ending up in my daughter's audio books).
these are running on different [ports] just fine, and I have assigned a different [name] for each. so I can easily select the wanted library in the player.

I would love the UI indicate what [name] I have assigned to the current instance, so I don't get confused managing these different instances.

elmodaddyb:
Support ability to show configured server name in the Web UI. Includes:

* Show in the Home menu option
* Show as the web page title

gerbera-friendly-name

Defaults to CMake Project Name(_Gerbera_) when not defined in `config.xml`

Fixes #363

### DIFF
--- a/gerbera-web/mock-api/auth/config.json
+++ b/gerbera-web/mock-api/auth/config.json
@@ -19,6 +19,7 @@
     },
     "actions": {
       "action": []
-    }
+    },
+    "friendlyName" : "Gerbera Media Server"
   }
 }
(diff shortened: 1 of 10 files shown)


elmodaddyb:
The one drawback to this implementation is that it is not-optional.  

> Do you think we should add a configuration that enables/disables the feature, or is that overkill?

PR #388, opened 2018-12-09
cmake - Set ffmpeg includes to avoid deprecated struct member

jkiusAlarus:
ffmpeg deprecated a while ago the codec member of the struct AVStream, replacing it with codecpar

When building gerbera with -DWITH_AVCODEC=1, the cmake detection process fails to set HAVE_AVSTREAM_CODECPAR because ffmpeg headers are in a specific location.

### DIFF
--- a/cmake/FindFFMPEG.cmake
+++ b/cmake/FindFFMPEG.cmake
@@ -206,6 +206,7 @@ IF (FFMPEG_INCLUDE_DIR)
 
                 SET(FFMPEG_LIBRARIES ${FFMPEG_LIBRARIES} CACHE INTERNAL "All presently found FFMPEG libraries.")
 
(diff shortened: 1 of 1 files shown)


elmodaddyb:
I tested this out and the `as_codecpar` error is now gone! Great Job

On a side note, looks like we still have some deprecation to fix.

```c++
development/gerbera/src/metadata/ffmpeg_handler.cc: In member function ‘virtual void FfmpegHandler::fillMetadata(zmm::Ref<CdsItem>)’:
/development/gerbera/src/metadata/ffmpeg_handler.cc:263:21: warning: ‘void av_register_all()’ is deprecated [-Wdeprecated-declarations]
     av_register_all();
```

v00d00:
Thanks @jkiusAlarus !

PR #392, opened 2018-12-14
Set required CMake version to 3.8

mjbayer:
It is not compiling with older versions (it failed with 3.7)

> The "Compile Features" functionality is now aware of C++ 17. 
https://cmake.org/cmake/help/v3.8/release/3.8.html


v00d00:
Thanks!

PR #395, opened 2018-12-15
Support command line options with string values

### LINKED ISSUE #394: CXXOpts throws std::bad_cast when setting configuration file option
### Issue
Following 6cf290d725a98b7d9b4f3d09db1e94ea66c35258, string options passed on command line are not properly parsed, causing `std::bad_cast` exception
```
$ gerbera -c /home/gerbera/.config/gerbera/config.xml
```
Causes gerbera to throw a bad cast exception and fail startup at 
https://github.com/gerbera/gerbera/blob/master/src/main.cc#L141

```
terminate called after throwing an instance of 'std::bad_cast'
  what():  std::bad_cast
Signal: SIGABRT (Aborted)

Process finished with exit code 1
```
> using no options allows Gerbera to start.

### Expected Outcome

Gerbera should support command line specification of the configuration file location.

### Possible Root Cause

It looks like the new options parser is defaulting all non-typed options to the first typed option.

```c++
...
("D,debug", "Enable debugging", cxxopts::value<bool>()->default_value("false"))
// et. al
("c,config", "Path to config file")  <---- becomes cxxopts::value<bool>
```
I am currently running C++ `c++ (GCC) 8.2.1 20181105`  so maybe I am missing something simple.

Explicitly setting the types for all `cxxopts` solves the problem.

elmodaddyb:
Fixes #394 

Debugging revealed that the command line options were not parsing as expected when no type is specified (_e.g. `-c ./config.xml`_).

> If a runtime configuration is missing perhaps the code change is not necessary.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -98,7 +98,20 @@ int main(int argc, char** argv, char** envp)
 
     cxxopts::Options options("gerbera", "Gerbera UPnP Media Server - https://gerbera.io");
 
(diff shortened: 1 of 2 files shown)


elmodaddyb:
I was running the new `gerbera-integration` docker suite, and found this issue :1st_place_medal:

Issue #402, opened 2018-12-29
AJAX Calls are cached when server no longer enforces Cache-Control


elmodaddyb:
Since the removal of `Cache-Control` headers on the server, 

https://github.com/gerbera/gerbera/commit/094923d03c2b6aca20ad415410a0452a3fb6e4f1#diff-e3ba151b835166ef3bd739d2f2a36f12

all AJAX calls from the client are **cached** by default.  This causes many things to break including long-polling updates, tree view refresh, etc.

## Introduce Cache-Control on Requests
This issue is to introduce `Cache-Control` on the client-side AJAX calls similar to below
```javascript
return $.ajax({
  url: GERBERA.App.clientConfig.api,
  type: 'get',
  data: requestData,
  headers: {
    'Cache-Control': 'no-cache, must-revalidate'
  }
})
```

elmodaddyb:
@v00d00 - I tested this locally(_still more testing required_) and it appears to resolve the caching issue.

> Since we control the client-side, it seems reasonable to include these headers.

Any thoughts?

v00d00:
Sounds good, nicer than adding a timestamp to the request at least.

elmodaddyb:
I found a way to set this header globally for all `$.ajax` calls...so makes it really straightforward

v00d00:
Closed by #403

PR #403, opened 2018-12-29
Set Cache-Control on all AJAX calls

elmodaddyb:
Adjust AJAX calls to have no client-side cache.  Cache-Control header allows repeated requests to always be directly to the server, rather than returned from browser cache.

> Applies `no-cache` to all AJAX calls.

Allows long-polling and tree navigation to work properly when re-visiting folders.


elmodaddyb:
Fixes #402

v00d00:
A couple of tests are failing?

elmodaddyb:
Yes, there is missing app config in 1 test... fixing now.

Issue #406, opened 2019-01-04
Playlist entries missing album name and artist


grb43:
My uPnP clients (e.g. TV, VLC) don't display the album name or artist when playing playlists with local audio entries.

They display when the files are played individually. And the album name and artist are displayed when the same playlists are served by MiniDLNA or Ampache.

Looking back I can see this was raised indirectly at Issue #31, and a new function, getCdsObject, was added to the playlist parser  (Commit 0182b60...). The intention was that getCdsObject enabled a playlist.js script to access an audio object's metadata to add it to the database; similar to the way import.js handles new audio objects.

To date, getCdsObject doesn't appear to have been documented and the playlist.js script hasn't been modified to allow it to fully populate the database.

Could this be added as enhancement? It would help to round out an otherwise great application.

elmodaddyb:
Hello @grb43 

#413 is merged via 7b10cb7d9bb0144ddba70c6ec59af4a915149aa6 to `origin/master`.

> Does `copyObject` and `getCdsObject` support address this issue?  I tested OK with sample playlist, but would like your confirmation regarding the issue.

Thanks,

Eamonn

grb43:
Hi @elmodaddyb 
Compiled and ran latest.
Tests
   -  New playlists (100 mp3) of files already in database. Result: all metadata available.
   -  New playlists (20 files) of files not in database. Result: files added to database and metadata available.

So all looks fine
Thank you
Graeme

elmodaddyb:
@grb43. That is great news. Thanks again for the great work!!

PR #407, opened 2019-01-05
Use getCdsObject to retrieve metadata for m3u playlist

grb43:
Follow up to Issue #406. I have messed around with the playlist.js script and can now see album and artist on my uPnP clients. I am very much a novice with js, but somebody may find it useful. I don't have enough knowledge to tackle pls playlists.

### DIFF
--- a/scripts/js/playlists.js
+++ b/scripts/js/playlists.js
@@ -40,19 +40,11 @@ function addPlaylistItem(location, title, playlistChain, order) {
     } else {
         if (location.substr(0,1) !== '/')
             location = playlistLocation + location;
-        var item  = {};
-        item.location = location;
-        if (title)
-            item.title = title;
-        else
-        {
-            var locationParts = location.split('/');
-            item.title = locationParts[locationParts.length - 1];
-            if (! item.title)
-                item.title = location;
-        }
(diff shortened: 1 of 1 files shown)


v00d00:
@elmodaddyb would probably be best to review this.

It also looks like this has broken a test?

elmodaddyb:
This looks cool.  The `getCdsObject` will create the playlist object, if it doesn't find it...

The tests fail because the `getCdsObject` duktape function needs to proxy the mock calls of the `getCdsObject` function within the test

https://github.com/gerbera/gerbera/blob/master/test/test_script/test_internal_m3u_playlist.cc#L87

> I will try to create the changes in next few days to support the testing of this....

elmodaddyb:
Good News, I was able to work out the pattern for testing this capability.  I will patch up the test suite and post back in a day or so.

elmodaddyb:
@grb43 - I was able to make the necessary test driven changes to support this behavior.

> All that remains is a real-time test of the playlist script to confirm it works OK

elmodaddyb:
@grb43 - I opened #413 to solve the build failures faced with this PR.

If you agree we can close this PR and merge #413

grb43:
@elmodaddyb Great, many thanks.

PR #408, opened 2019-01-07
gerbera doesn't generate the folder inside config

isramv:
some instructions necessary.


elmodaddyb:
Thanks for improving the docs!!

PR #409, opened 2019-01-08
Dockerfile created

nemoload:


v00d00:
I do think we need to improve our Docker experience.

@elmodaddyb Do you think it would make more sense to "officalise" your [repo](https://github.com/elmodaddyb/gerbera-docker)?

I dont think we need a Dockerfile in the main project, but we could move it into the Gerbera org and then link it from Docs/Readme?

nemoload:
I think maintaining an official Docker image associated with the project is a necessity because Gerbera has as it offers a fast, secure and clean way to deploy the service efficiently. The place of the Dockerfile is negotiable as long as it's actively, officially maintained.

While writing my Dockerfile I put into consideration that the image should require minimal user intervention (except mapping your multimedia directory to the `root` directory of the container) and the image size should be as small as it could be (my image is 772MB while @elmodaddyb image is 1.3 GB)

elmodaddyb:
Yea, I have a bunch of docker files in 
https://github.com/elmodaddyb/gerbera-docker
...and they definitely outdated

Using `FROM gerbera/gerbera-base-ci` is actually even less code.

Below is the latest and greatest dockerfile which runs with full containerized volumes, etc...

https://github.com/elmodaddyb/gerbera-integration/blob/master/gerbera-core/Dockerfile

> This is part of the integration test suite that is coming together nicely...

v00d00:
I'm not sure we should use `gerbera-base-ci` as the base for the end image as it contains 3 compiler toolchains and various build time stuff that isn't needed!

In my view ideally the flow would be that the build produces somekind of package/artifact that is then _those_ used by the end docker image to keep it minimal.

Up until now we have relied on 3rd parties to produce said packages/artifacts though. I guess the obvious choice would be an alpine package/repo and image.

nemoload:
We can use a multi-stage Dockerfile with Ubuntu image contains one of the building toolchain then we copy the produced executable file to an Alpine image which contains only the runtime dependencies.

elmodaddyb:
@nemoload - When you say _copy executable_ do you mean using Docker volume to transfer the published app?

> thanks for the help in understanding...

nemoload:
@elmodaddyb No. I meant something like that:
```Dockerfile
FROM golang:1.7.3
WORKDIR /go/src/github.com/alexellis/href-counter/
RUN go get -d -v golang.org/x/net/html  
COPY app.go .
RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .

FROM alpine:latest  
RUN apk --no-cache add ca-certificates
WORKDIR /root/
COPY --from=0 /go/src/github.com/alexellis/href-counter/app .
CMD ["./app"]  
```
https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

elmodaddyb:
@nemoload - Thanks for the followup ... Multi-stage builds sounds really neat.  This could eliminate a ton of development software needed to compile, but not needed for runtime

PR #429, opened 2019-03-08
MakeLists.txt: fix static build with curl and libidn2

ffontaine:
curl can be statically linked with libidn2, in this case, build fails:
/accts/mlweber1/rclinux/rc-buildroot-test/scripts/instance-0/output/host/m68k-buildroot-linux-uclibc/sysroot/usr/lib/libcurl.a(libcurl_la-url.o): In function `free_idnconverted_hostname.isra.1':
url.c:(.text+0xf4): undefined reference to `idn2_free'

To fix this issue, add a call to pkg_check_modules to retrieve any
needed dependencies

Fixes:
 - http://autobuild.buildroot.org/results/be5893b507d22a23951efeea20c18642742cef5a


v00d00:
Thanks for the patch.

My understanding was that CMake would [do it for us](https://github.com/Kitware/CMake/blob/master/Modules/FindCURL.cmake)?

ffontaine:
As you pointed out, cmake added a call to pkg-config in FindCurl with https://github.com/Kitware/CMake/commit/fc5afbe97094d741b853b045d51351f3a370e87e#diff-62c4a3bc6680cd4945de9749af697a2a

Howerber, this commit will only be a part of the upcoming 3.14.0.
All "official" versions (including 3.13.4) does not have this call.

v00d00:
Ahh, makes sense!

PR #436, opened 2019-03-21
cmake/FindFFMPEG.cmake: use pkgconfig

ffontaine:
ffmpeg has too many optional dependencies (opus, openssl, ...) instead
of trying to find them all one by one, use pkg-config to get them and
fallback on current mechanism

Fixes:
 - http://autobuild.buildroot.org/results/2b99fabd798db84a0fce26ad696c58e54c6ff626
 - http://autobuild.buildroot.org/results/95e410e5ab34c6d4626a58f97c0d2d5e6829a300


v00d00:
Thanks!

PR #437, opened 2019-03-22
Set file size for file handler

### LINKED ISSUE #430: Invalid thumbnail served to VLC
Gerbera 1.3.0 with libupnp-1.8.4, VLC 3.0.6

Thumbnails are not shown in VLC.
Wireshark trace shows that they are downloaded, but contents is not valid jpeg.
First, VLC sends request with Content-Range which Gerbera seems to not like:\
```
GET /content/media/object_id/1545/res_id/1/rh/6/ext/file.jpg HTTP/1.1
Host: 192.168.0.135:49152
Accept: */*
Accept-Language: en_US
User-Agent: VLC/3.0.6 LibVLC/3.0.6
Range: bytes=0-

HTTP/1.1 416 Requested Range Not Satisfiable
SERVER: Linux/4.20.10-gentoo, UPnP/1.0, Portable SDK for UPnP devices/1.8.4
CONNECTION: close
CONTENT-LENGTH: 70
Accept-Ranges: bytes
CONTENT-TYPE: text/html

<html><body><h1>416 Requested Range Not Satisfiable</h1></body></html>
```

Then it repeats same URL with HTTP/1.0:
```
GET /content/media/object_id/1545/res_id/1/rh/6/ext/file.jpg HTTP/1.0
Host: 192.168.0.135:49152
User-Agent: VLC/3.0.6 LibVLC/3.0.6
Icy-MetaData: 1

HTTP/1.0 200 OK
CONTENT-TYPE: image/jpeg
DATE: Sat, 09 Mar 2019 16:58:00 GMT
LAST-MODIFIED: Sat, 06 Feb 2016 12:59:09 GMT
SERVER: Linux/4.20.10-gentoo, UPnP/1.0, Portable SDK for UPnP devices/1.8.4
X-User-Agent: redsonic
Content-Disposition: attachment; filename="VID_20180103_143716.mp4"


......JFIF
```

The problem is with an extra newline following response headers. It gets saved into file resulting in a corrupted jpeg:
```
00000000  0d 0a ff d8 ff e0 00 10  4a 46 49 46 00 01 01 00  |........JFIF....|
```

wget downloads same url without any problems. And as I understand HTTP/1.0 also requires only one empty line between response headers and body

elmodaddyb:
Fixes #430 

* Does not set content length to `-1` and instead uses the file size calculated from the FileHandler

This tested OK, although it was definitely written hastily.....a second opinion is worth the time before merging.  I did not reconcile if setting `UpnpFileInfo_set_FileLength` explicitly has any consequences in underlying libupnp?

### DIFF
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -174,7 +174,7 @@ void FileRequestHandler::getInfo(IN const char* filename, OUT UpnpFileInfo* info
         if (!string_ok(mimeType))
             mimeType = h->getMimeType();
 
-        off_t size = 0;
+        off_t size = -1;
         h->serveContent(item, res_id, &(size));
         UpnpFileInfo_set_FileLength(info, size);
     } else if (!is_srt && string_ok(tr_profile)) {
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -157,10 +157,6 @@ void FileRequestHandler::getInfo(IN const char* filename, OUT UpnpFileInfo* info
     if (((res_id > 0) && (res_id < item->getResourceCount()))
         || ((res_id > 0) && string_ok(rh))) {
 
-        log_debug("setting content length to unknown\n");
-        /// \todo we could figure out the content length...
-        UpnpFileInfo_set_FileLength(info, -1);
-
         int res_handler;
         if (string_ok(rh))
             res_handler = rh.toInt();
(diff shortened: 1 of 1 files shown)


v00d00:
Looks sensible, should we initialise it to -1 in case the handler doesnt set the length?

elmodaddyb:
Yea, good idea.  I updated to default `off_t size = -1`

PR #443, opened 2019-04-02
cmake/FindLibUpnp.cmake: fix static linking

ffontaine:
Fix static linking with an openssl enabled upnp:
 - remove libupnp-1.8 from pkg_check_modules call otherwise test will
   fail if libupnp-1.8.pc is not found
 - add PC_UPNP_LIBRARIES to UPNP_LIBRARIES

Fixes:
 - http://autobuild.buildroot.org/results/10098c8972725d54b717ddc8ea41f4de5e5b066d

### DIFF
--- a/cmake/FindLibUpnp.cmake
+++ b/cmake/FindLibUpnp.cmake
@@ -8,7 +8,7 @@
 #  UPNP_HAS_REUSEADDR - If LinUPnP was built with SO_REUSEADDR support
 
 find_package(PkgConfig QUIET)
-pkg_check_modules (PC_UPNP QUIET libupnp-1.8 libupnp)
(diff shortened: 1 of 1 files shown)


cmake/FindLibUpnp.cmake
```diff
@@ -8,7 +8,7 @@
 #  UPNP_HAS_REUSEADDR - If LinUPnP was built with SO_REUSEADDR support
 
 find_package(PkgConfig QUIET)
-pkg_check_modules (PC_UPNP QUIET libupnp-1.8 libupnp)
+pkg_check_modules (PC_UPNP QUIET libupnp)
```
v00d00: My bad, this should have been `pkg_search_module` instead

ffontaine: OK, I'll update the PR.

PR #450, opened 2019-04-18
feat: provide a way to toggle DLNA-seeking with a config

rtm-ctrlz:
By default: seeking is enabled (as it was originally)
Location of switch in config: /server/protocolInfo/attribute::dlna-seek
Config generator and config tests: updated and passing


v00d00:
Thanks for the PR! I do wonder why you would want to disable seeking like that though?

rtm-ctrlz:
My case isn't very common: my son has autistic disorder, but he loves cartoons and he knows how to use tv remote. Of course there are some interesting moments, because of autistic disorder he seek-and-watch some 3-5 second moment over and over again for pretty long time and yes this makes him happy, but it's Ok for 5 times, may be Ok for a 5 minutes, but it definitely is not Ok for a longer period, unfortunately my son himself cannot break this endless loop "fun -> seek for 5s back -> repeat". Not watching cartoon or hiding remote is not an option, but disabling seek I break this loop once and for ever. Because now to watch one more time single moment he need to watch whole cartoon once again and can be too boring or to long for him to wait this moment, so he decides to watch next cartoon instead.

Also, with this PR we get named constants for "supported seek options" (none/time based/range based/both) which is more clear to read, understand and (I hope so) use in future, instead of hardcoded "01" and "10" strings.

Also there was a error (upnp_xml.cc:921) - for "disabling" seek (for transcoded content) was used "range based seek supported" ("10") instead of "no seek options" ("00") ;)

rtm-ctrlz:
Two more things:
 1. I'm not so sure about adding "dlna-seek" option to config generator - may be this kind of option can be left only on documentation.
 2. I don't know a proper way how to up xsd schema, which by the way pointing to http://mediatomb.cc/config

v00d00:
@elmodaddyb Whats your opinion on having this in the generated file? Is optional stuff best left in the docs?

elmodaddyb:
It sounds like a good use case to include in the generator.  We had a few scenarios where the `extended-options` were poorly documented (until recently 👍 ).  It would have been nice to have them out of the box in the generated file.

v00d00:
Thanks @rtm-ctrlz

PR #451, opened 2019-04-19
main: bugfix `interface` CLI option

edzius:
CLI `interface` option parsing is broken after changes in 6cf290d725a98b7d9b4f3d09db1e94ea66c35258. This fixes interface value assignment to correct variable.


v00d00:
Ouch, thanks.

PR #466, opened 2019-05-14
cmake/FindFFMPEG: do not quieten messages when using pkg-config

joerg-krause:
gerbera does print some configuration status messages, e.g. _"Found TagLib: ..."_, _"Found ZLIB: ..."_, but it does not so for FFmpeg. To make the status message behaviour more consistent, do not quieten the log messages from pkg-config module.

On my system, this prints:

```
-- Checking for modules 'libavformat;libavutil'
--   Found libavformat, version 57.83.100
--   Found libavutil, version 55.78.100
```


v00d00:
@joerg-krause Thanks, LGTM

joerg-krause:
Thanks for merging, there are more log messages quietened by `QUIET`.

I find is quite useful when configuring gerbera, so I can send a PR removing all `QUIET` entries for the dependencies?

v00d00:
Most of the modules should print their found versions, but in not sure about removing the quiets for all pkgconfig calls.

Fixing up the modules so they do print the versions would be good though.

PR #472, opened 2019-05-22
Convert Gerbera UI to ES6

elmodaddyb:
Upgrade the client-side code to ES6 modules.

* [ECMAScript 6](http://es6-features.org) Standard - Supported by all major browsers [can i use - es6-module](https://caniuse.com/#feat=es6-module)
* Use native Promise instead of jQuery promise

I created entirely new `-module.js` files to guard against a possible failure with ES6.  The different name allows for side-by-side loading of both ES5 & ES6 (_if required_)


elmodaddyb:
I found bug https://github.com/gerbera/gerbera/issues/473 while testing the system out.

> It appears to be a non-critical sequential processing failure.

Let me know if you would like to delay this merge until it is fixed (_likely a bug in current JS too_)

v00d00:
I am thinking of doing a `1.3.2` release, so will wait until that has been cut.

v00d00:
1.3.2 is tagged, so this is getting merged!

PR #477, opened 2019-05-27
Allow to set the manufacturer and the modelURL via config file

joerg-krause:
This allows to override the default values via the config file.


v00d00:
Thanks for the patch, the code looks sensible.

However I am not sure why you would want to advertise your Gerbera instance as something else?

joerg-krause:
> However I am not sure why you would want to advertise your Gerbera instance as something else?

For example when using Gerbera in audio systems, like the Moode Audio Player, or in commercial open-source products.

joerg-krause:
Thanks!

PR #482, opened 2019-07-10
Add DeviceDescriptionHandler

v00d00:
This is currently not used, but will be needed when we switch to the
newer `UpnpRegisterRootDeviceX` methods.

It serves the device description XML as generated by
`renderDeviceDescription()` at `<serverpath>/content/description.xml`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,8 @@ set(libgerberaFILES
         src/contrib/md5.h
         src/curl_io_handler.cc
         src/curl_io_handler.h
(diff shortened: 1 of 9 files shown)


v00d00:
In reality, this doesnt work. Because we have no link between `getInfo` and `open` we cant do anything upfront. I guess I need to chase upstream to merge the 2nd cookie patch.

v00d00:
Prodded upstream at https://github.com/mrjimenez/pupnp/issues/109

PR #506, opened 2019-09-17
Support inotify kernel driver on illumos/OmniOS

whorfin:
quick fixes to enable inotify in illumos/OmniOS builds


whyman:
Looks good, thanks!

Issue #30, opened 2017-01-10
meta-data mkv


alainstark:
This is an example of a mkv file with meta-data.

[metadata.mkv.zip](https://github.com/v00d00/mediatomb/files/697566/metadata.mkv.zip)

This is the [matroska documentation for tagging](https://matroska.org/technical/specs/tagging/index.html) files.
And this is the [cover art matroska documentation](https://matroska.org/technical/cover_art/index.html).

The goal is to enhance mediatomb metadata handler to use these mkv meta-data.

whyman:
Fixed with #540

PR #540, opened 2019-12-06
#30 add metadata mkv support via libmatroska

pamapa:


pamapa:
CI fails because it has not matroska installed. Should i by default disable matroska?

whyman:
I dont think it needs to be disabled by default, we just need to add it to the [base image](https://github.com/gerbera/gerbera-ci-base-docker).

pamapa:
cool, i added the needed dev libraries there:
https://github.com/gerbera/gerbera-ci-base-docker/pull/1

whyman:
Thanks, LGTM!

PR #549, opened 2019-12-14
replace zmm::String with std::string

pamapa:
I know, this is a huge change. This replacement touches a lot of code. It took more several hours to come up with this pull request. I just found that having a own String class make the whole project more complex. Its mostly easier to stick with what comes with C++. That is STL, which is nowadays quite good.

If this pull request gets in i am willing to work and replace more of zmm. Namely std::vector, std::map and std::shared_ptr are my friends.

I did not find a less invasive approach. The Ref/Object class system is very sticking to everything else.

- most code changes are trivial
- some are tricky:
  - zmm::String can be nullptr, where std::string can not
  - std::stoi does not auto convert empty string to 0
  - zmm::String::index(pos, ch) -> std::string::find(ch, pos)
  - zmm::String::index return -1 when not found, std::string::find return std::string::npos
  - Ref<Array<String>> has been replaced by std::vector<std::string> -> copied where needed, no refcounting needed
- all existing unit tests are working
- VLC as client is already working here
- found issues by using gdb + valgrind

**Future**:
- replacing Array, Dictionary, Ref with STL variants (std::vector, std::map, std::shared_ptr) 
- replacing mxml with standard lightweight xml library (e.g. pugixml)
- easier to share/use code with other projects
- more standard (at the time the code was written it made sense, but not these days)
- easier to develop and understand code
- easier debugging

It needs some fine tuning (remove strings.h too) and more testing.
What do you think?


whyman:
This is great and something I had been meaning to do for a while!

whyman:
Looks like the lasftm code needs converting also

pamapa:
cool that you like this too.
I have fixed LastFM, exiv2 and MySQL. strings.h is now gone too.
CI passes, ready for merging :-)

pamapa:
Fixed:
creating sqlite3 table from scratch works now

What i do not yet understand is why a journal file is added

src/tools.cc
```diff
@@ -136,17 +137,89 @@ String trim_string(String str)
...
             break;
         }
     }
-    return str.substring(start, end - start);
+    return str.substr(start, end - start);
+}
+
+bool startswith_string(std::string str, std::string check)
+{
+    return str.rfind(check) == 0;
```
whyman: Shouldnt this be `rfind( check, 0)`? Otherwise we search the whole string?

I feel like all the utility methods could lose the `_string`, the typing makes it pretty obvious, and the code tidier

pamapa: good catch, i have fixed this

pamapa:
The thing with the journal file does not seem to be new, when i close gerbera properly the journal vanishes. Its known that temporary such a file might exist until the DB buffers are written to disk.

So we get ready for merge :-)

pamapa:
Once this merge is through i plan to do:
- remove now obsolete: string_ok (is simpty a !str.empy()) + string_ok_ex (only one one caller)
- restructure:
  - move utils files into src/utils
  - move io_handler files into src/io_handler
- add some unit tests for important features in utils + io_handler

src/tools.cc
pamapa: - renaming tolower_string to tolower leads to problem as it clashes with existing symbols (same for stoi_string)

Maybe it makes sense to introduce a util/string.cc/h modules containing all string util functions. There we could stream line the functions like:
- X_str or str_X (e.g. X == tolower, ....)
- or you have a better idea

whyman:
I do want to merge it but I am wondering if I should release a 1.3.6 first 🤔

src/tools.cc
pamapa: or static class or namespace :-)

pamapa:
This stuff is definitely not bugfixing. You should go first with a 1.3.6 release. When do you plan to do so?

whyman:
This weekend is looking most likely

src/tools.cc
whyman: Yes, to any of those :)

whyman:
Hey, can you rebase this? Then will merge straight away

pamapa:
should be ok now

whyman:
Can you squash the commits? If I do it I will get set at the author which is unfair to your efforts!

pamapa:
i have combined the commits in a new feature branch....

PR #558, opened 2019-12-24
fix matroska build without js, taglib or atrailers

ffontaine:
i2i function is used in matroska_handler.cc but this function is defined
only if defined(HAVE_JS) || defined(HAVE_TAGLIB) || defined(ATRAILERS)
as a result compilation fails if HAVE_MATROSKA is set but HAVE_JS,
HAVE_TAGLIG or ATRAILERS are not.


whyman:
Nice spot, thanks!

PR #560, opened 2019-12-25
fix mkv on 32bit

pamapa:
i have checked other projects, which use libeml. Nobody is using StdIOCallback directly, they all have their own implementations, so lets do the same and use fseeko + ftello...


whyman:
That copyright is just wrong really. It should be a gerbera one as it was a new file...

PR #563, opened 2019-12-26
replace zmm::String with std::string

pamapa:
- most is trivial
- tricky:
  - zmm::String can be nullptr, where std::string can not
  - std::stoi does not auto convert empty string to 0
  - zmm::String::index(pos, ch) -> std::string::find(ch, pos)
  - zmm::String::index return -1 when not found, std::string::find return std::string::npos
  - Ref<Array<String>> has been replaced by std::vector<std::string> -> copied where needed, no refcounting needed

trivial: replace _(x) with x and remove strings.h
avoid null pointer to std::string in exception cases
replace String to std::string
remove strings.h from CMakeLists.txt too
fix: make sure correct quote function is taken for c string
fix escape function: < 0 vs. std::string::npos
rename startswith_string to startswith and fix implementation
drop left over includes of strrings.h
_() fix


pamapa:
i have combined my commits in a new feature branch

pamapa:
i am happy when this is merged. In the next days i have time to do some more stuff....

Issue #562, opened 2019-12-25
Linking error ...


count00zero:
Hi again!

Currently I am a bit confused about using whyman/gerbera or gerbera/gerbera.
When using the former I experience the following problem when Linking_
```

[100%] Linking CXX executable gerbera
/usr/bin/ld: -lupnp-1.8 not found
collect2: error: ld returned 1 exit status
```

The interesing part is, that upnp is found above:
`Found UPnP: /usr/local/lib/libupnp.so (found version "1.8.6")`

Do u have any suggestions?

Thanks in advance and BR,
count

whyman:
You should use the main `gerbera/gerbera` repo most likely, `whyman/gerbera` is my personal version.

Odd that we find the non suffixed library, but then the linker flag is wrong. I guess there is a mix up in pkg-config somwhere, files from an older libupnp around perhaps?

whyman:
So it looks like this is caused by `PC_UPNP_LIBRARIES` being included in the libraries, we find some libraries manually using paths from pkg-config.

But @ffontaine had issues with static linking ssl enabled upnp so included these extra libraries, I wonder if there is some concept of "static linking libraries" or "extra libraries" so we dont have both a full path and `-lupnp-1.8` -- but that said, they should in theory match anyway :man_shrugging:

count00zero:
OK, so I will use the gerbera/gerbera, thanks for the information.

When using the newly pulled sources, I get a new one: 
```

[  1%] Building CXX object CMakeFiles/libgerbera.dir/src/metadata/exiv2_handler.cc.o
/opt/_Build/gerbera/src/metadata/exiv2_handler.cc:36:10: fatal error: config_manager.h: Datei oder Verzeichnis nicht gefunden
   36 | #include "config_manager.h"
```
Meaning: config_manager.h file or folder not found ...

whyman:
@pamapa Can you sort this? And maybe check the other files...

count00zero:
Thank you guys!

Now the process gets a bit further (52%); when it comes to the mysql part it halts again:
```

/opt/_Build/gerbera/src/storage/mysql/mysql_storage.cc:40:10: fatal error: config_manager.h: Datei oder Verzeichnis nicht gefunden
   40 | #include "config_manager.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

When I disable MYSQL-Support via the cmake option, it stops at 67% with the following:
```

[...]

PR #570, opened 2019-12-30
part of #562: include was not adapted when moving config

pamapa:
We definitely should improve CI.

github itself supports since short also CI via workflows/actions. I have another project where i use it and i am quite happy (https://github.com/pamapa/callblocker/tree/master/.github/workflows). No need for additional account, docker based, perfect integrated into github...

in the meantime, i have installed exiv2 locally too


whyman:
The CI just needs adjusting to build all the optional deps, github or travis makes no difference.

Issue #567, opened 2019-12-28
replace zmm:Ref/zmm::Object with std::shared_ptr or std::unique_ptr


pamapa:
- [x] singleton.h/cc (replaced by injecting objects directly into depended classes via constructor, see #374)
- [x] content_manager.h/cc
- [x] config_manager.h/cc
- [x] storage/*
- [x] iohandler/*
- [x] metadata/*
- [x] request_handler.h/cc
- [x] subscription_request.h/cc
- [x] action_request.h/cc
- [x] cds_objects.h/cc
- [x] cds_resource.h/cc
- [x] autoscan*.h/cc
- [x] timer.h/cc
- [x] mxml/* (replaced by using https://pugixml.org)

pamapa:
you can assign me

pamapa:
@whyman:
i dont think its worth porting mxml to std. We should replace it with a standard xml library. What do you think? On the other hand mxml is not only used for the configuration part but also in the upnp code.
- possible candidate: https://pugixml.org
- others?

whyman:
We already link to `libexpat` so could use it? Though its its C so not that nice to use. If you are going to swap it out, make sure to replace the expat as well mxml.

pamapa:
done

PR #573, opened 2019-12-30
part of #567: remove singleton.h/cc

pamapa:
only part for #567, but it makes sense to merge it. Otherwise we will end up with a huge change. I tried to keep the invasive changes as small as possible. Testing this feature uncovered some previous std::string conversion bugs, for which i have made other pull requests.


src/server.cc
```diff
@@ -248,20 +244,26 @@ void Server::shutdown()
...
+    config->emptyBookmark();
     server_shutdown_flag = true;
 
     log_debug("Server shutting down\n");
 
     ret = UpnpUnRegisterRootDevice(deviceHandle);
     if (ret != UPNP_E_SUCCESS) {
-        throw _UpnpException(ret, "upnp_cleanup: UpnpUnRegisterRootDevice failed");
+        log_error("upnp_cleanup: UpnpUnRegisterRootDevice failed: %i", ret);
     }
```
whyman: Why no exception any more?

src/server.cc
```diff
@@ -332,12 +345,12 @@ int Server::handleUpnpEvent(Upnp_EventType eventtype, const void* event)
     return ret;
 }
 
-std::string Server::getIP() const
+std::string Server::getIP()
```
whyman: Did you mean to remove the const?

src/update_manager.cc
```diff
@@ -172,7 +174,7 @@ void UpdateManager::threadProc()
     struct timespec lastUpdate;
     getTimespecNow(&lastUpdate);
 
-    unique_lock<mutex_type> lock(mutex);
+    AutoLockU lock(mutex);
```
whyman: Have you tested this change? Is it needed? I remember we broke some stuff with a similar change some time ago.

src/server.cc
```diff
@@ -248,20 +244,26 @@ void Server::shutdown()
 {
     int ret = 0; // return code
 
-    /*
-    ContentManager::getInstance()->shutdown();
-    UpdateManager::getInstance()->shutdown();
-    Storage::getInstance()->shutdown();
-    */
+#if 0 // enable later
```
whyman: I guess you are addressing this shortly?

whyman:
Nice work, thats big change! Ive added some comments but overall LGTM

src/server.cc
pamapa: we are in the shutdown phase, i think its better to execute the following code in that function too instead of skipping...

src/server.cc
pamapa: good point. there is a place in the code where getIP and getPort is called. indead of passing through the server via shared_ptr, i simply made getIP and getPort static. but now "const" is no longer allowed...

src/update_manager.cc
pamapa: i hope i have not changed behavior here. AutoLockU should still resolve to unique_lock. (see using AutoLockU = std::unique_lock<decltype(mutex)>; in update_manager.h)

src/server.cc
pamapa: yes, i was first confused by the existing commented code and did not want to add something new. but then i realized that the singleton was always calling shutdown.
i will enable this.

src/server.cc
whyman: Yep, good spot, missed the context.

pamapa:
shutdown phase is in place now too. ready to merge.

PR #575, opened 2020-01-01
cmake/FindLibUpnp.cmake: drop libupnp-1.8

ffontaine:
-1.8 suffix has been dropped from pupnp since version 1.8.2 and commit
https://github.com/pupnp/pupnp/commit/07f504c61bd9e4d93eb3d373ffc8527cafe0b9af,
so drop search for libupnp-1.8 to avoid retrieving old or broken libupnp
versions


whyman:
Applied via #577

PR #585, opened 2020-01-02
#567: add const in subscription_request too

pamapa:


whyman:
Maybe combine some of these PRs? I can review the commits individually?

pamapa:
ok, i will combine in the future

PR #607, opened 2020-01-12
Feature/replace zmm ref mxml

pamapa:
pcdata " and ' escaping:
Solved (not needed for now) and if needed by a messy device, we can provide a patch with an extra option (`format_escape_quotes`) in pugixml.
Related pugixml issue https://github.com/zeux/pugixml/issues/320


pamapa:
@whyman:
I am struggeling with the pcdata field of `<Result>` from `<DIDL-Lite>`

Do you have a link to the upnp specification, where it is defined how to encode the `<Result>` field? I just can find only examples like [this](https://developer.sony.com/develop/audio-control-api/get-started/browse-dlna-file#tutorial-step-6) (see end of that file). All examples i find use `&quot;` instead of " in the pcdata. But pure XML specification allows " inside pcdata...

pamapa:
@whyman 
Found another bug, when i fixed that is also seems to work with " instead of `&quot;`. Nice! I am using VLC for my tests.

whyman:
The UPnP specs are at: https://openconnectivity.org/developer/specifications/upnp-resources/upnp/

There is a schema: http://www.upnp.org/schemas/av/didl-lite-v2.xsd too

pamapa:
pugixml pcdata is no issue, i will continue with this replacement...

pamapa:
ready for review

src/onlineservice/sopcast_content_handler.cc
```diff
@@ -39,193 +39,187 @@
...
-#define DATE_BUF_LEN 12
-    std::string temp;
-    struct timespec ts;
-
-    while (current_group_node_index < group_count) {
-        if (current_group == nullptr) {
-            Ref<Node> n = channels->getChild(current_group_node_index);
-            current_group_node_index++;
+    auto root = service_xml->document_element();
+    bool skipGroup = true;
```
pamapa: this is not correct for the first call of getNextObject, i need to fix this first.
like:

skipGroup = false
while (...)
  if (skipGroup || ...) {
     ...
     skipGoup = false; // we are on the next group
     continue;
  }
  skipGroup = true;
  ...

pamapa:
found an issue, i have commented it, want to fix it before merging...

pamapa:
fixed an rebased

whyman:
@pamapa, amazing work. Seriously, I thought that the zmm stuff would be around for another decade! Thanks for putting in the work!

PR #608, opened 2020-01-12
cleanup generated directory in tearDown

pamapa:


whyman:
Where does that `ftw.h` header come from?

pamapa:
Official "Unix" header file. See here https://pubs.opengroup.org/onlinepubs/7908799/xsh/ftw.h.html and here http://man7.org/linux/man-pages/man3/ftw.3.html

PR #613, opened 2020-01-14
Dockerfile: install newer spdlog

whyman:


whyman:
Thanks to @KarlStraussberger for the script :+1:

Issue #626, opened 2020-01-21
Fix CI


whyman:
Fun times in travis: https://travis-ci.com/gerbera/gerbera/builds/145440723

GCC 9: Passes
GCC 8: Unit Test segfault
Clang 8: Build failure

whyman:
Fixed in #267

PR #627, opened 2020-01-21
Drop libc++ with clang

whyman:
We need to link to (non-libcxx) system libs

Fixes Clang part of #626.

CI is still failing with GCC8 unit test


whyman:
Annoyingly I couldn't get the test to segfault with gcc 8.3.0 locally :/

PR #650, opened 2020-02-09
Support libupnp 1.12.0

whyman:


src/util/headers.cc
```diff
@@ -58,7 +57,7 @@ void Headers::addHeader(const std::string& header, const std::string& value)
     }
     log_debug("Adding header: '{}: {}'", cleanHeader.c_str(), cleanValue.c_str());
 
-    headers->insert({cleanHeader, cleanValue});
+    headers->push_back({cleanHeader, cleanValue});
```
neheb: should be emplace_back.

whyman: @neheb Thanks - and for the other clang-tidy PRs too. I will get those merged once Ive unborked the CI...

neheb: OK. I'll rebase those once done.

PR #696, opened 2020-02-23
fix #693: allow symbol link for scripts used in configuration

### LINKED ISSUE #693: error: File '/home/xxx/.gerbera/js' does not exist!
Hi there,

since one of the last commits (3 or so ago), I cant' get gerbera to run; I am always confronted with that: `error: File '/home/xxx/.gerbera/js' does not exist!`

Do you have an clue?
If I create an empty file, gerbera starts, but doesn't work properly (no file is imported).
If I create a symbolic link, it doesn't work (but older versions do).

I am clueless ....

pamapa:


whyman:
Nice!

PR #702, opened 2020-02-26
cxxopts: remove u8 usage

neheb:
From clang: warning: type of UTF-8 string literal will change from array
of const char to array of const char8_t in C++2a


whyman:
Going to leave this one alone as its an external project. Perhaps raise it with then?

PR #705, opened 2020-02-26
replace NULL with nullptr

neheb:
Found with clang's -Wzero-as-null-pointer-constant


whyman:
LGTM, thanks

PR #708, opened 2020-02-26
import script variable lost in copy&paste

KarlStraussberger:
Running a test I got the message "file /etc/gerbera/js does not exists"


whyman:
Oh nice, this should sort #693

Issue #692, opened 2020-02-18
Videos are shown in wrong path (Raspbian Buster)


ItsMeMarc:
Hi,

I noticed a strange behavior in my gerbera installation (Rasperry Pi 4, Raspbian Buster, 32 bit kernel):

The video /media/pi/raid/DirA/DirB/DirC/DirD/Video.avi was displayed correctly via gerbera in VLC in the path Video/Directories/DirA/DirB/DirC/DirD.
I deleted the video and copied a new one into the same path on my Rasperry Pi. The file name is the same, only the extension has been changed from avi to mkv (different container). Now the video is displayed in VLC in the path Video/Directories/DirD. The subdirectories DirA/DirB/DirC are skipped...

The whole thing can also be seen in the log file:

**Old avi file**
```
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:241] fillMetadata(): Running ffmpeg handler on /media/pi/raid/DirA/DirB/DirC/DirD/Video.avi
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:173] addFfmpegResourceFields(): Added duration: 00:22:50.5
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:181] addFfmpegResourceFields(): Added overall bitrate: 178248 kb/s
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:200] addFfmpegResourceFields(): FourCC: 44495658 = XVID
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:210] addFfmpegResourceFields(): Added resolution: 512x384 pixel
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:219] addFfmpegResourceFields(): Added sample frequency: 48000 Hz
2020-02-08 22:46:17   DEBUG: [src/metadata/ffmpeg_handler.cc:225] addFfmpegResourceFields(): Added number of audio channels: 2
2020-02-08 22:46:17   DEBUG: [src/metadata/fanart_handler.cc:82] fillMetadata(): Running fanart handler on /media/pi/raid/DirA/DirB/DirC/DirD/Video.avi
2020-02-08 22:46:17   DEBUG: [src/metadata/fanart_handler.cc:62] getFolderName(): Folder name: /media/pi/raid/DirA/DirB/DirC/DirD
[...]

PR #714, opened 2020-02-29
shortten virtual path to videos

pamapa:
saw this when fixing issue #692. I see no benefit of passing last subdir of added location to virtual path too.

**Example:**
config.xml:
```
...
    <autoscan use-inotify="auto">
        <directory location="/home/.../Videos" mode="inotify" level="full" recursive="yes" hidden-files="no"/>
    </autoscan>
...
```
Before virtual path: /Video/Directories/**Videos**/Action/...`
With change: /Video/Directories/Action/...

### DIFF
--- a/src/layout/fallback_layout.cc
+++ b/src/layout/fallback_layout.cc
@@ -138,8 +138,8 @@ void FallbackLayout::addImage(const std::shared_ptr<CdsObject>& obj, const fs::p
 
     std::string dir;
     if (!rootpath.empty()) {
-        // make location relative to rootpath: "/home/.../Photos/Action/a.mkv" with rootpath "/home/.../Photos" -> "Photos/Action"
-        dir = fs::relative(obj->getLocation().parent_path(), rootpath.parent_path());
+        // make location relative to rootpath: "/home/.../Photos/Action/a.mkv" with rootpath "/home/.../Photos" -> "Action"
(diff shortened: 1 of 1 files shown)


pamapa:
please do not merge yet, audio + images are also affected

pamapa:
done for photos too now, ready to merge

whyman:
Is this still needed? I can't remember the context.

pamapa:
The idea of that patch is to shorten the virtual path

Its optional and does change the virtual directory structure, by removing a parent path, which was probably not wanted anyway.

When you specify a directory location. I expect to get all inside that directory, without the directory name.
```
<autoscan use-inotify="auto">
        <directory location="/home/user/Bla" mode="inotify" level="full" recursive="yes" hidden-files="no"/>
    </autoscan>
```

Virtual directory structure:
- Without patch: /Video/Directories/**Bla**/Action/...
- With patch: /Video/Directories/Action/...

whyman:
Ah, yes, it's an improvement

count00zero:
just a short question regarding <autoscan ...
Where should it be inserted within the config file?

I put it after </scripting> but it isn't recognized when I don't close it with `</autoscan>`.
BUT if I do, I get the following error:
`Error parsing config file '~/.config/gerbera/config.xml': Start-end tags mismatch`

PR #718, opened 2020-03-03
spdlog build path improvement, fanart semantic and exception fix, and spdlog flushing

whorfin:
Greetings.  Lots has changed since last I peeked, fancy C++ idioms and long link times...
Three fixes in order of priority:

1.  This most recent master was completely broken for me, removing almost all my media.  I saw errors thrown, saying "/folder.jpg" was a filesystem exception.  The problem is twofold.  First, the check for fanart shouldn't throw an exception.  Second, the fanart path was bogus.  The `/` operator isn't just "+ for paths", it has semantics, and the leading "/" in the two `names` blows away the base path if it is rooted.  I've lost track of how "fanart" differs from associating images with albums; if this code is still important it could stand a more diverse set of names and some case-insensitive matching, but at least now it doesn't kill a scan.

2. The spdlog I'm using via Joyent is just a normal library, with a pkg-config file but no .cmake package definition.  This change allows that to work, only trying the hard-require .cmake config if spdlog isn't otherwise found.

3. While debugging, I noticed the logfile appeared empty, for long stretches of time.  Aggressive buffering made it unhelpful for diagnosing problems.  I've added a `flush_on` to ensure lines are written as they happen.

Finally, with a build of this current tree, I was observing
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Abort
```
which did _not_ get logged to the logfile

Running overnight, it also used up all system memory, populating a new database which should be about 1G.  So, there's some problems in master right now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -322,8 +322,16 @@ endif()
 find_package(fmt REQUIRED)
 target_link_libraries(gerbera fmt::fmt)
 
-find_package(spdlog REQUIRED)
-target_link_libraries(gerbera spdlog::spdlog)
+# See if it's just installed as a library first, as not all installs have .cmake files
+find_library(SPDLOG_LIBRARY spdlog)
+if(SPDLOG_LIBRARY)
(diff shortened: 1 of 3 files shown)


whyman:
Looks good to me! Thanks!

And thanks for trying a master build

PR #719, opened 2020-03-06
fix std::string with nullptr in MySQL code

wesleycoakley:
While booting this project for the first time, I ran into the following error:

```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

It appears this has been addressed before (8ac412a) but that commit did not discover these two areas in the configuration code where the same error arises.


whyman:
Awesome, thanks!

PR #721, opened 2020-03-06
fix autoscan timed

pamapa:


pamapa:
The whole autoscan implementation needs a refactoring, most part should not be in content_manager but in a seperate file. However its a lot of work...

whyman:
I did start on changing the autoscan stuff, buts on hold with the arrival of my newborn 😄🍼

PR #726, opened 2020-03-07
replace memory.cc/h with malloc/free, we can use valgrind for debug

pamapa:


pamapa:
@whyman 
Could not find your email somewhere, can you ping me via pammann@gmx.net?

whyman:
@pampa ian@gerbera.io

PR #732, opened 2020-03-09
fix compilation error

Rogach:
If building without FFmpeg support there was a compilation error -
MALLOC() instead of malloc() in one place. (see e373740)


whyman:
Thanks!

PR #736, opened 2020-03-10
switch several defines to constexpr

neheb:
Eliminates several clang-tidy warnings.

Fixed up new compile failures.


neheb:
This is not a good change. Increases filesize. Closing.

Issue #737, opened 2020-03-11
Video file is not recognized (Log says: Not a file)


ItsMeMarc:
Hi,

On my Raspberry Pi 4 (Raspbian Buster, 32 bit kernel) I recognized another strange behaviour.

In a directory there are four mkv files. I can play all of them in VLC without any problems.
```
pi@rpi4:/media/pi/raid/Videos/TEMP/Temp $ ls -lh --time-style=long-iso
insgesamt 8,0G
-rwxr--r-- 1 pi pi 2,5G 2014-03-29 09:27 1.mkv
-rwxr--r-- 1 pi pi 1,5G 2014-03-28 17:24 2.mkv
-rwxr--r-- 1 pi pi 1,7G 2014-03-29 10:36 3.mkv
-rwxr--r-- 1 pi pi 2,3G 2014-03-29 09:29 4.mkv
```
But gerbera only recognizes 2.mkv and 3.mkv as valid video files. For the other two files the debug log says `Not a file`. What does this mean? Is there a way to get some more debug information?
```
2020-03-11 17:35:34  debug: [content_manager.cc:731] addRecursive(): Checking path /media/pi/raid/Videos/TEMP/Temp
2020-03-11 17:35:34  debug: [content_manager.cc:748] addRecursive(): IS TASK VALID? [true], task path: [/media/pi/raid/Videos/TEMP/Temp]
2020-03-11 17:35:34 warning: [content_manager.cc:828] addRecursive(): skipping /media/pi/raid/Videos/TEMP/Temp/4.mkv : Not a file: /media/pi/raid/Videos/TEMP/Temp/4.mkv
2020-03-11 17:35:34 warning: [content_manager.cc:828] addRecursive(): skipping /media/pi/raid/Videos/TEMP/Temp/1.mkv : Not a file: /media/pi/raid/Videos/TEMP/Temp/1.mkv
```
Thanks.

kyak:
Hi!

Both 1.mkv and 4.mkv are bigger than 2 Gb. Might be related to this: https://github.com/gerbera/gerbera/issues/738

pamapa:
are you using master or 1.4.0?

ItsMeMarc:
@pamapa 
I'm using master:
```
* commit 0bda7840 (HEAD -> master, origin/master, origin/HEAD)
| Date:   2020-03-09 16:52:58 +0300
```
@kyak 
[...]

PR #743, opened 2020-03-15
workaround is_regular_file for file > 2GB on 32bit systems

pamapa:
workaround for #737


whyman:
Wow.

neheb:
Does this error have anything to do with _FILE_OFFSET_BITS?

whyman:
Probably this was fixed with GCC 9.3 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81091

neheb:
Ah that is unfortunate.

neheb:
Looks like this was backported to GCC 8.4.

I wonder if this was ever a problem with musl which always uses 64-bit APIs.

Issue #744, opened 2020-03-17
Please provide a full url to subtitle file in the CaptionInfoEx attribute


ytimenkov:
Apparently VLC doesn't resolve relative URL to subtitle (I'm not sure if this is strictly possible - what is the base since media URL is a full one).
So for it to load subtitle properly a full URL needs to be specified.
I usually patch Gerbera locally and it works fine:

```diff
diff --git a/src/upnp_xml.cc b/src/upnp_xml.cc
index 19a58fa7..6ac23724 100644
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -397,7 +397,7 @@ void UpnpXMLBuilder::renderCaptionInfo(const std::string& URL, pugi::xml_node* p
     // though it's necessary.
 
     size_t endp = URL.rfind('.');
-    cap.append_child(pugi::node_pcdata).set_value((URL.substr(0, endp) + ".srt").c_str());
+    cap.append_child(pugi::node_pcdata).set_value((virtualURL + URL.substr(0, endp) + ".srt").c_str());
     cap.append_attribute("sec:type") = "srt";
 }

diff --git a/src/upnp_xml.h b/src/upnp_xml.h
index b3d83910..dae71f05 100644
--- a/src/upnp_xml.h
+++ b/src/upnp_xml.h
@@ -90,7 +90,7 @@ public:
 
     /// \brief Renders a subtitle resource tag (Samsung proprietary extension)
     /// \param URL download location of the video item
-    static void renderCaptionInfo(const std::string& URL, pugi::xml_node* parent);
+    void renderCaptionInfo(const std::string& URL, pugi::xml_node* parent);
 
     static void renderCreator(const std::string& creator, pugi::xml_node* parent);
     static void renderAlbumArtURI(const std::string& uri, pugi::xml_node* parent);
 ```

I also don't think adding subtitle based on "SM Hack" option is necessary: I've compared to Kodi and it adds `sec:CaptionInfoEx`, `sec:CaptionInfo` (and many other) elements unconditionally.

```diff
diff --git a/src/upnp_xml.cc b/src/upnp_xml.cc
index 19a58fa7..6ac23724 100644
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -831,11 +831,11 @@ void UpnpXMLBuilder::addResources(const std::shared_ptr<CdsItem>& item, pugi::xm
[...]

PR #745, opened 2020-03-17
Return full URL of the subtitle (#744)

ytimenkov:
Apparently VLC requires a full URL to load subtitle properly
(treating as local path otherwise).

Add sec:CaptionInfoEx tag unconditionally as other UPNP servers do.


ytimenkov:
It's really cool that you've started to use modern C++ and clean up the code :+1:

PR #756, opened 2020-03-26
Warnings-free build with gcc 9.

ytimenkov:
Minor refactoring to build without warnings with Gcc 9 and latest
libraries.

Fixed undefined behavior in matroska_handler and taglib_handler.


whyman:
Nice! Thanks for the CONTRIBUTING updates too

PR #761, opened 2020-03-27
imitate same behaviour as std::filesystem::is_regular_file

pamapa:


src/util/tools.cc
```diff
@@ -176,9 +176,24 @@ bool isRegularFile(const fs::path& path)
     // unfortunately fs::is_regular_file(path, ec) does not to work for files >2GB on ARM 32bit systems (see #737)
     struct stat statbuf;
     int ret = stat(path.c_str(), &statbuf);
-    if (ret != 0)
+    if (ret != 0) {
+        throw std::runtime_error(mt_strerror(errno) + ": " + path.string());
```
whyman: Maybe we should include the function name in the error, just to make it 1000% clear?

pamapa: good point will do i follow up

pamapa: hmmm there multiple throws in tools.cc, etc

Should be do something like:
```
#define throw_runtime_error(msg) throw std::runtime_error(std::string(msg) \
  + " line: " + std::to_string(__LINE__) + " file: " + std::string(__FILE__))
```
in exceptions.h
?

whyman: Yes, nice idea!

Issue #752, opened 2020-03-22
add client detection code


pamapa:
After some debugging i have found a way to achieve this with the used upnp sdk. We need to register additionally a client callback via `UpnpRegisterClient` then we get `UpnpDiscovery` events where we can extract modelName, serialNumber and friendlyName. Those can be used to create a cache entry. When now a `UpnpActionRequest` event comes we can lookup the cache entry via `UpnpActionRequest_get_CtrlPtIPAddr`.

I will prepare a patch for this feature. This feature can then later be used for client specific quirks used for Samsung TV, XBox, etc...

PR #768, opened 2020-03-28
feat: #752 add client detection code

pamapa:
next steps would be:
- replace/remove <protocolInfo> from config.xml with hard coded quirk flags (samsung, etc...)
- replace/remove <mappings> from config.xml with hard coded quirk flags (samsung, etc...)
- move quirks to upnp_quirks.cc and apply them according to detected type automatically

### DIFF
--- a/src/util/upnp_clients.cc
+++ b/src/util/upnp_clients.cc
@@ -1,0 +1,166 @@
+/*GRB*
+
+    Gerbera - https://gerbera.io/
+    
+    upnp_clients.cc - this file is part of Gerbera.
+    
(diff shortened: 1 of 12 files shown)


src/util/upnp_clients.cc
```diff
@@ -0,0 +1,167 @@
+/*MT*
+    
+    MediaTomb - http://www.mediatomb.cc/
```
whyman: This should be a Gerbera licence header as its a new file.

pamapa: done + matroska handler files i added previously too

PR #774, opened 2020-04-05
part of #752 improve client detection

pamapa:
I have provided two patches in pupnp to make client detection possible.

- we need upnp 1.12.0 + patches or upcoming 1.12.1
- for upnp 1.8.x define UPNP_HAS_IPADDR_AND_OS_IN_FILEINFO is not set, client info will be 'Unknown'


pamapa:
can you please re-run docker repo to take latest version of install_pupnp.sh?

whyman:
Yeah, it would be nice if we can poke upstream to get that 1.12.1 release out!

pamapa:
theoretically if you rebuild the docker image in https://github.com/gerbera/gerbera-ci-base-docker i should pick up my patches as well...

whyman:
Yeah, but distros are not going to ship a patched upstream release just for us. Its fine as a stopgap until a new release, though.

CMakeLists.txt
```diff
@@ -313,6 +313,7 @@ endif()
 
 if (UPNP_VERSION_STRING VERSION_GREATER_EQUAL "1.12.0")
     add_definitions(-DUPNP_1_12_LIST)
+    add_definitions(-DUPNP_HAS_IPADDR_AND_OS_IN_FILEINFO)
```
whyman, line 316: This will need to moved to a 1.12.1 test now its released!

pamapa: or should we make a hard depend on 1.12.1?

pamapa:
this merge request depends on "part of #776 start to consolidate quirks". When this merge request is landed i can make use of the quirk class and do not need to inject CientInfo struct...

pamapa:
@whyman 
Or should we just drop supporting anything below 1.12.1. And hard require minimum 1.12.1?

whyman:
@pamapa I think drop anything older than 1.12.1 for 1.5 and tidy up the ifdefs. Can be done separately though.

Issue #776, opened 2020-04-05
consolidate client quirks in single file


pamapa:
- gives a better overview of what is handled specially

PR #777, opened 2020-04-05
part of #776 start to consolidate quirks

pamapa:


whyman:
Wow, this is so nice!

I wonder if its worth keeping `CFG_SERVER_EXTEND_PROTOCOLINFO` and `SM_HACK` etc as an option at all.

I think we should include the extended info all the time, so shouldnt be a "quirk" IMO, and quirks are applied by client detection, so `SM_HACK` etc shouldnt be needed.

pamapa:
I am planing to replace CFG_SERVER_EXTEND_PROTOCOLINFO* with client detection code, such that each client only gets what he needs. For that i will need the other merge request to go in :-)

Currently CFG_SERVER_EXTEND_PROTOCOLINFO is a pool of various quirks for different kind of devices. For example there is a notice that one if is for Bravia TV, etc... . I prefer to explicitly know what is needed by whom. The non quirks code path should be 100% upnp complainant.

pamapa:
Questions:

```
if (config->getBoolOption(CFG_SERVER_EXTEND_PROTOCOLINFO)) {
        // we do not do DLNA yet but this is needed for bravia tv (v5500)
...
```
-> quirk

```
if (config->getBoolOption(CFG_SERVER_EXTEND_PROTOCOLINFO)) {
                        /// \todo clean this up, make sure to check the mimetype and
                        /// provide the profile correctly
                        aa.append_attribute("xmlns:dlna") = "urn:schemas-dlna-org:metadata-1-0";
                        aa.append_attribute("dlna:profileID") = "JPEG_TN";
                    }
```
-> quirk for which client?

```
if (config->getBoolOption(CFG_SERVER_EXTEND_PROTOCOLINFO)) {
            std::string extend;
            if (contentType == CONTENT_TYPE_JPG) {
...
```
-> quirk for which client?

what about CFG_SERVER_EXTEND_PROTOCOLINFO_DLNA_SEEK, is that client specifc?

src/file_request_handler.cc
```diff
@@ -65,9 +66,11 @@ FileRequestHandler::FileRequestHandler(std::shared_ptr<ConfigManager> config,
 
 void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
 {
-    Headers headers;
     log_debug("start");
 
+    auto quirks = std::make_unique<Quirks>(config);
```
ytimenkov, line 71: I see you have sometimes `shared_ptr` for `Quirks`, `unique_ptr` here. I wonder if you really need to create one on heap all the time?

I can imagine if you had more generic name like "Request PostProcessor" with virtual interface which is injected for unit testing... But for example in this particular function it could be just on stack.

I also see through the code base that you toss `shared_ptrs` around quite much and this makes me really think "Is this possible that function does need to potentially take an ownership of the object or is it just "becase shared ptrs are everywhere"?

I think even Cpp Guidelines (for modern C++) encourage use of raw pointers as a "borrowed" object (that is if you don't have `delete` in your code you don't need to think twice whether to deallocate object or not).

I'm asking to understand and deduce coding style here a bit better :)

whyman:
> \<snip\>
> 
> -> quirk for which client?
> 
> what about `CFG_SERVER_EXTEND_PROTOCOLINFO_DLNA_SEEK`, is that client specifc? Something to do with the header `"TimeSeekRange.dlna.org"`?

All the ones you listed are not client specific, they are just part of DLNA, it applies to any DLNA device, of which most things are nowadays.

ytimenkov:
What about starting to make use of real encapsulation?

Main idea is that most of `if`s should be turned into different implementations of some interface. You may disagree, but to me this significantly simplifies the code.

Basically instead of `Quirks` class, come up with some abstract class, like `ResponseHooks` which defines extension points which are called through the processing of the request.

Then there should be a factory, like `ResponseHookProvider` which takes User Agent (or all headers) and returns an implementation of hooks for this particular client per request.

Then there will be at least 2 implementations: a default one (no op) and for Samsung.

Essentially this means moving all
```c++
  if (!config->getBoolOption(CFG_SERVER_EXTEND_PROTOCOLINFO_SM_HACK))
```

From each and every method in quirks into a single place in the factory with straight-forward implementations of particular classes.

Another benefit is that if `Provider` is injectable and Hooks are fairly isolated it is much simpler to write unit tests against them.

What do you think?

ytimenkov:
In addition to the previous comment I could imagine different processing for different clients (or options), for example for VLC (if I remember the code correctly) it should be possible to just add multiple `text/srt` resources as subtitle tracks instead of single `CaptionInfoEx` tag...

src/file_request_handler.cc
pamapa: its always a question of speed vs safety. Stack allocation is super fast. My first goal is to not leak any memory and to use an easy to use concept. Whenever speed is not critical do not loose time with optimizations. Focus on hot spots first.

If you have pointers allover in your code how to you know who is responsible for deleting it again? It gets quiet quite messy, specially if more than one person is coding in the same base.

Also consider this: e.g. std::string:
The string object itself is stored on the stack but it points to memory that is on the heap. Maybe some parts is optimized away by the compiler.

pamapa:
Most clients are very close i would like to avoid copy/past to much code just because of some variations.
So for quirks i do not see it.

But i was also think about something like this. But not for quirks but for different kind of provider:
- direct + transcoding via external program
- gstreamer (direct+ transcoding)
- ...

Inside of each the quirks can be applied as well.

pamapa:
@whyman
So i will just remove the if (CFG_SERVER_EXTEND_PROTOCOLINFO_DLNA_SEEK) and if (CFG_SERVER_EXTEND_PROTOCOLINFO) and keep the code where it is.

If a device popups, which does not work with that. We make for that specific device a "not if" quirk :-)

pamapa:
I will remove CFG_SERVER_EXTEND_PROTOCOLINFO+CFG_SERVER_EXTEND_PROTOCOLINFO_DLNA_SEEK in a different merge request. I need to first land the Samsung automatic detection code.

pamapa:
@whyman
ok like this?

whyman:
LGTM, feel free to merge it once the conflicts are sorted @pamapa

src/file_request_handler.cc
ytimenkov: @pamapa 

I totally agree with your first paragraph.

However I'm most worried not about efficiency but rather semantics and expressiveness: When I (as someone not familiar with the code, but just applying a common sense) read smart pointer I assume that ownership of the object can be transferred somewhere.

* Passing `shared_ptr` by const ref means it can be potentially stored somewhere
* Passing `unique_ptr` by ref means it *could be* stored (maybe under some circumstances).
* Passing by value or `&&` means "sink": it will be stored, but the caller doesn't need it.

The choice `unique_ptr` vs `shared_ptr` has also significant impact on thread-safety requirements on type: in case of `unique_ptr` one can assume "external synchronization" by owner (i.e. mutex or just by design). in case of `shared_ptr` most probably class needs to be thread-safe internally (or have only `const` methods which is an another way to express "thread safe") since it's hard to synchronize access for arbitrary number of owners (since anyone can pass it further to other objects).

That's said I got confused by mixed usage of `unique_ptr` and `shared_ptr` as handlers for a single class. I just can't easily tell guarantee that class must promise.
 Especially confusing looked local `unique_ptr` without passing it somewhere neither using any polymorphic behavior (In this case returning from a factory is a passing of ownership, at least once).

Most of this is based on C++ Core Guidelines, which also recommend usage of *raw* pointers. I really though: Did Bjarne got dementia? when I first read it, but after some thinking it makes perfect sense: If you ban:
1. Explicit calls to `new` and `delete` (which you do)
2. "Adopting" raw pointer into a smart pointer (i.e. allow only `make_*` functions to create objects) - which you also do.

There is just no question "Who needs to delete?". It's simple: no one! (And given above bans you just can't). Essentially raw pointers become just a reference which can be a null or rebound (where you need such behavior). 

I can take that this code base is insanely old and was full of raw pointers and manual memory management so it was plain dangerous to allow raw pointers, but now it seems ok, so maybe it's time to allow this.

Would be really much easier if most projects can just point at C++ Core Guidelines (well, maybe with some exceptions as usual).

PR #782, opened 2020-04-09
clang warnings

neheb:
Stuff I found with -Weverything.


whyman:
Looks good apart from the virtual destructors, which I am not confident enough with to merge blindly.

neheb:
:)

The third or last commit?

whyman:
`9dc8a77`

neheb:
OK. There are also -Wnon-virtual-dtor warnings, but I chose to ignore those.

Safest option would be to defer these two commits.

ytimenkov:
Does clang really force you to put virtual destructors in every derived class?
It looks terribly verbose, should be enough to do it only in the base class...

neheb:
Good call. Changed to only add default destructors for base classes.

I find it interesting that with that change, -Wnon-virtual-dtor now stops complaining all over the place. It complains about class Storage and SQLResult only.

ytimenkov:
> se classes.
> I find it interesting that with that change, -Wnon-virtual-dtor now stops complaining all over the place. It complains about class Storage and SQLResult only.

That is legitimate. Although with `Storage` is not strictly necessary: objects are created using `make_shared` which captures the destructor of the concrete most derived class, so no destruction via pointer to a base class happens.

Same seems to be true for `SqlResult`. But even then `Sqlite3Result` declares a virtual destructor...

whyman:
Travis is down for Maintenance
> Scheduled maintenance is currently in progress. We will provide updates as necessary.

neheb:
Looking more at it, I think the first commit should be redone.

I think adding deleted destructors is a better idea. Will need to do some testing.

neheb:
Bah never mind:

```
/usr/bin/../lib/gcc/x86_64-redhat-linux/9/../../../../include/c++/9/bits/unique_ptr.h:81:2: error: attempt to use a deleted function
        delete __ptr;
```

whyman:
@neheb Can you rebase this please?

neheb:
Done.

Issue #749, opened 2020-03-20
upnp:artist@role[AlbumArtist] breaks folder display on clients


michaelkl:
I came across the following problem: _some folders with MP3 files were displayed empty by DLNA client_.

I used debug mode to inspect the exact data transferred to a client and it turned out that items were actually sent by Gerbera. By analyzing the logs for folders which works correctly I finally tracked the problem down to a `upnp:artist@role[AlbumArtist]` tag.

I know nothing about DLNA protocol but quick googling revealed articles like [this](https://community.netgear.com/t5/Using-your-ReadyNAS-in-Business/ReadyDLNA-1-1-2-3-solved-my-albumartist-problems-but/td-p/952775).

So, I tried to fix the problem with a simple SQL:
```
UPDATE mt_metadata
SET property_name='upnp:albumArtist'
WHERE property_name='upnp:artist@role[AlbumArtist]'
```
**And it helped!**

The problem persists with the following clients:
- VLC (Android)
- BubbleUPnP (Android)
- LG Smart TV

The problem does not persist with:
- foobar2000 (Android)

BubbleUPnP even shows the error message when I was trying to enter this folder: "The Media Server sent invalid data".

Might also be related to #707

whyman:
Is this a master build?

michaelkl:
> Is this a master build?

Yes, but not the latest. I built it a month or two ago. But I am pretty sure [it is still here](https://github.com/gerbera/gerbera/blob/bd66bc847b0ca09e3571dddb20d0a7e0a2aedf95/src/metadata/metadata_handler.cc#L81).

count00zero:
I tried it with the latest master build: same behaviour; SQL fix helped
https://github.com/gerbera/gerbera/issues/707#issuecomment-602260613
https://github.com/gerbera/gerbera/issues/707#issuecomment-602263434

whyman:
Thanks for reporting and confirming both.

astylos:
Seeing same issue, submitting a PR

whyman:
Ive taken a look into this.

The issue here that the special code that handeld the album artist got lost during the removal of the legacy `mxml` stuff.

`upnp:artist@role[AlbumArtist]` should not end up in the XML. It needs to be rendered as `<upnp:artist role="AlbumArtist">`

Here is the original commit: https://github.com/gerbera/gerbera/commit/b31f659235b9fb328abb1404f9e7e07bdfdcb18d where the special case was added.

[...]

Issue #707, opened 2020-02-26
No files are exposed via DLNA/uPNP


count00zero:
The following happens, when using master with an empty databse, when adding files:

Gerbera starts and runs and imports (status info on web-gui), but unfortunately it doesn't really work, because no audio-(or other)-file is presented via DLNA/uPNP although they are visible via the web-gui

The path structure is also visible when accessing DLNA, but no files inside the folders are visible/exposed!

Attached the difference between vanilla & master with a nearly identical config.

![Master vs Vanilla](https://user-images.githubusercontent.com/14225157/75330895-1d5f5080-5882-11ea-9661-c04ff0c19922.jpg)

whyman:
Are you using the JS layout?

count00zero:
nope: in this szenario (although I tested them all) i use "disabled"

pamapa:
I have found an issue recently introduce in upnp_xml.cc. Maybe this pull request helps you:
https://github.com/gerbera/gerbera/pull/712

count00zero:
unfortunately: that didn't do it :-(

whyman:
Remember you may need to clear your DB and reimport

count00zero:
yeah, of course. 
Did that (that was initially the reason I detected the problem => cleared the DB -> added some directories -> was happy because they showed up in the GUI -> was unhappy because they were not exposed via upnp/dlna ...

pamapa:
i am not able to reproduce locally. Since you are running from master you could bisect (https://www.metaltoad.com/blog/beginners-guide-git-bisect-process-elimination) and find the bad commit?

Simplest way could also be a binary search using `git cehckout [commit hash].`. Where you find first a working base and then you take a version in the middle, ...

pamapa:
ok, i may have found the reason, will provide a patch soon...

count00zero:
tried that, I seem to be not able to choose a right commit hash without any other problems for me ...

Nevertheless: with the latest commit, at least images seem to work (though I didn't specifically test that before) ...

pamapa:
I made the same patch for video too. But for Audio, it is different. There is no '/Audio/Directories/..."' in the builtin layout.

count00zero:
[...]

PR #783, opened 2020-04-10
Albumartist

astylos:
Fix for #749


astylos:
Why do all the updates I've pulled into my for show here?

ytimenkov:
> Why do all the updates I've pulled into my for show here?

That's how git works... They normally should be empty ones (just having 2 parents).

If you want to remove those (usually a good idea before submitting a PR) you can do rebase on `upstream/master`.
(If you didn't have any conflicts it's the easiest way, otherwise I would go with `merge --squash` into a new branch started from master and then either force-pushing to the same branch or creating a new PR).

whyman:
I'm not certain that this fix is right, something else must be broken as it has worked fine since `1.1.0` 3 years ago. It has some special chars, so wondering if its different behaviour with the new XML lib?

astylos:
I've been having this issue for quite some time, but never bothered to look into it. It's only reading #749 that i decided to check if it was same issue i was seeing. But agree it could also be an issue with xml and how it gets encoded. 

How did current tag mapping come about? Looks like gerbera isn't the only one having issues with this: [***](url). The xml in the final comment looks like what gerbera should be emitting if using role is the way to go.

astylos:
> > Why do all the updates I've pulled into my for show here?
> 
> That's how git works... They normally should be empty ones (just having 2 parents).
> 
> If you want to remove those (usually a good idea before submitting a PR) you can do rebase on `upstream/master`.
Thanks

pamapa:
i work a lot with rebase, it keeps the changes lime and nice. You do not get any virtual merge commits, etc...

1. make sure your local master is up to date
  - git checkout master
  - git pull master upstream
  - git push

2. rebase feature branch
  - git checkout <feature/branch>
  - git rebase master # sometimes needs fixing conflics
  - git push --force origin <feature/branch>

3. if you want to change you last commit
  - git add <changed files>
  - git commit --amend
  - git push --force origin <feature/branch>

pamapa:
@astylos 
That fix will work for all new read in metadata. Additional to that a db migration step is needed for all users which still have the wrong string in their db. Like described here: https://github.com/gerbera/gerbera/issues/707#issuecomment-602263434

whyman:
Closing this as per https://github.com/gerbera/gerbera/issues/749#issuecomment-615246006

> 
> Ive taken a look into this.
> 
> The issue here that the special code that handeld the album artist got lost during the removal of the legacy mxml stuff.
> 
> upnp:artist@role[AlbumArtist] should not end up in the XML. It needs to be rendered as <upnp:artist role="AlbumArtist">
> 
> Here is the original commit: b31f659 where the special case was added.
> 
> This should be pretty easy to restore with the nice new XML library.

Issue #691, opened 2020-02-18
Should not override interface flags for spdlog


marillat:
Hi,
I reported this bug https://bugs.debian.org/951474

The Debian maintainer tell me the problem is in  CMakeLists.txt added in commit fe81e5f
This line should be removed:

```add_compile_definitions("SPDLOG_FMT_EXTERNAL")```

This fix the build for several architectures (armel, armhf, arm64 and powerpc).

whyman:
Ah, well, we just need to improve the CMake so that it complains if the fmt is not >=6.

marillat:
/usr/lib/x86_64-linux-gnu/cmake/fmt/fmt-config-version.cmake
seems to be here for that.

ytimenkov:
OOC: which feature of v6 is required?

whyman:
Nothing that we are using directly, but for compatibility with spdlog

pamapa:
should be fixed now with above commit

ytimenkov:
That was not merged. I'm cooking another one 😐

PR #791, opened 2020-04-13
Don't override interface flags for spdlog.

ytimenkov:
Finding fmt explicitly is not needed since it's handled by spdlog's
CMake files.

Fixes: #691.

What I checked:
1. Built spdlog 1.5 without `SPDLOG_FMT_EXTERNAL`: Gerbera compiles as expected (all fmt stuff comes from `bundled/` folder of spdlog).
2. Built fmt and spdlog with `SPDLOG_FMT_EXTERNAL=ON`. Both were installed into different paths.
When configuring gerbera I (expectedly) got this:
```
CMake Error at /usr/share/cmake-3.13/Modules/CMakeFindDependencyMacro.cmake:48 (find_package):
  Could not find a package configuration file provided by "fmt" with any of
  the following names:

    fmtConfig.cmake
    fmt-config.cmake

  Add the installation prefix of "fmt" to CMAKE_PREFIX_PATH or set "fmt_DIR"
  to a directory containing one of the above files.  If "fmt" provides a
  separate development package or SDK, be sure it has been installed.
  /home/dev/install-spdlog/usr/local/lib/cmake/spdlog/spdlogConfig.cmake:11 (find_dependency)
  CMakeLists.txt:328 (find_package)
```
to verify that spdlog indeed looks for fmt itself. After fixing search paths I've got Gerbera compiling with flags: `-isystem /home/dev/install-spdlog/usr/local/include -isystem /home/dev/install-fmt/usr/local/include`.

There are 2 things which a bit unclear:
1. Whether to add check for fmt >= 6. It is possible to check for version since both `SPDLOG_FMT_EXTERNAL` and `fmt_VERSION_MAJOR` CMake variables are available (given that both spdlog and fmt are found using Package Config mode), but I think this check should be a part of spdlog (to not violate encapsulation).
2. Gerbera searches for `SPDLOG_LIBRARY` using `find_library` and this will not bring fmt into the scope. However I don't know what is this use-case and it doesn't handle include paths at all and there is not way to check whether fmt was bundled or not. I think this use-case is already broken.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,20 +317,17 @@ else()
     add_definitions(-DUPNP_HAS_EXTRA_HEADERS_LIST)
 endif()
 
-find_package(fmt REQUIRED)
-target_link_libraries(gerbera fmt::fmt)
 
 # See if it's just installed as a library first, as not all installs have .cmake files
 find_library(SPDLOG_LIBRARY spdlog)
 if(SPDLOG_LIBRARY)
     set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} ${SPDLOG_LIBRARY})
-    target_link_libraries(gerbera ${SPDLOG_LIBRARY})
+    target_link_libraries(libgerbera ${SPDLOG_LIBRARY})
 else()
     # Didn't find it, so try the .cmake file, fail if not found
     find_package(spdlog REQUIRED)
-    target_link_libraries(gerbera spdlog::spdlog)
+    target_link_libraries(libgerbera spdlog::spdlog)
 endif()
-add_compile_definitions("SPDLOG_FMT_EXTERNAL")
 
 find_package (Pugixml REQUIRED)
 if (PUGIXML_FOUND)


CMakeLists.txt
```diff
@@ -317,20 +317,17 @@ else()
...
 find_library(SPDLOG_LIBRARY spdlog)
 if(SPDLOG_LIBRARY)
     set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} ${SPDLOG_LIBRARY})
-    target_link_libraries(gerbera ${SPDLOG_LIBRARY})
+    target_link_libraries(libgerbera ${SPDLOG_LIBRARY})
 else()
     # Didn't find it, so try the .cmake file, fail if not found
     find_package(spdlog REQUIRED)
-    target_link_libraries(gerbera spdlog::spdlog)
+    target_link_libraries(libgerbera spdlog::spdlog)
```
ytimenkov, line 329: spdlog needs to be linked to `libgerbera` because it includes its headers. It compiles if you dump all headers into a single system path, but if it's a really separate path, Gerbera won't build.

ytimenkov:
I've also noticed that `Dockerfile` installs fmt-dev, but then `install-spdlog.sh` builds with bundled fmt. In which way this inconsistency should be resolved? (if...)

ytimenkov:
Actually `find_library(SPDLOG_LIBRARY spdlog)` even messes things up when I try to consume spdlog from build tree (i.e. check out, build in subdir and then point cmake at it with `cmake -DCMAKE_PREFIX_PATH=~/spdlog/build/ .`): it just finds a library and skips looking for package with proper defines and include paths.

Not sure however if this is the case you worry about...

whyman:
My thinking for this was: as we using `fmt` we may as well depend on it in CMake rather than relying on `spdlog`s bundled version.

And if we are depending on it anyway, then we may as well use it also for `spdlog`.

That said, I am not against removing `SPDLOG_FMT_EXTERNAL`, but then we will effectively have two versions of fmt and that feels like asking for trouble.

ytimenkov:
> That said, I am not against removing SPDLOG_FMT_EXTERNAL, but then we will effectively have two versions of fmt and that feels like asking for trouble.

That I 100% agree with.

The question is whether Gerbera can dictate requirements on its upstream dependencies? That's what apps shipped as containers usually free to do, while so far I assumed that Gerbera should be a part of a distro (I personally use it with Gentoo which is probably easiest to tweak).

If you don't want to rely on spdlog's bundled fmt (although I don't see practical issues right now with this) I could change scripts to look for fmt *and* fail if `SPDLOG_FMT_EXTERNAL` is false. At least this will be explicit.

Unfortunately after reading original bug report couple times it feels like they use bundled fmt but have (had?) only fmt5 installed. This means either we do this fix or ask them later to change spdlog to have external fmt.

Personally I'm fine with the latter and I consider all those "and by the way we bundle library X, y and Z so we're dependency-free and header-only" as ugly hacks. Dependencies are dependencies and should be explicit and now are quite well handled with packet managers (had amazing experience with Conan at work).
/imho :D

ytimenkov:
> This means either we do this fix or ask them later to change spdlog to have external fmt.

To be clear: I'm fine with both options, it's not my decision to take :)

whyman:
@ytimenkov Okay, then I think lets close this one, and make our dependency on `fmt` >=6 explicit in CMake with a nice message if its not found.

ytimenkov:
Okay, but what to do with external_fmt?

Even if we just start requiring fmt>=6 we still get 2 versions of fmt library in Debian: one coming from spdlog (compiled into it) and one in gerbera (will be compiled in). It will probably work for now and `ld` will handle it even in ringtone, but you won't know what version of fmt being executed (say, if some bug is discovered in fmt).

PR #800, opened 2020-04-17
make use of client detection code

pamapa:


pamapa:
i have added now a hard depends on libupnp 1.12.1, as otherwise it might be a step backwards for Samsung TV owners, if distribution are building against older libupnp libs

pamapa:
ready for merging

PR #804, opened 2020-04-17
Don't override interface flags for spdlog.

ytimenkov:
Finding fmt explicitly is not needed since it's handled by spdlog's
CMake files.

Fixes: #691.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -310,19 +310,31 @@ if (NOT UPNP_HAS_REUSEADDR)
 endif()
 
 find_package(fmt REQUIRED)
-target_link_libraries(gerbera fmt::fmt)
+target_link_libraries(libgerbera fmt::fmt)
 
 # See if it's just installed as a library first, as not all installs have .cmake files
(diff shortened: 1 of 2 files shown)


ytimenkov:
Attempt number 2 :)

I looked at spdlog history and it seems they switched to fmt 6.0 in version 1.4.

I also checked on my ubuntu 19.10 with libspdlog-dev=1:1.3.1-1 and libfmt-dev=5.2.1+ds-2, seems to build (although I've got my warning about `SPDLOG_FMT_EXTERNAL`).

However I can remove it if you don't want to have it.

whyman:
Docker build failed with `spdlog/fmt/fmt.h:22:10: fatal error: spdlog/fmt/bundled/core.h: No such file or directory`

ytimenkov:
I think I have an idea why this is happening. This could then be the reason for adding define of `SPDLOG_FMT_EXTERNAL`...

whyman:
LGTM! Thanks

PR #810, opened 2020-04-19
introduce Config interface, which will allow easier Mocking in unit-tests

pamapa:


src/config/config.h
```diff
@@ -0,0 +1,198 @@
...
+} config_option_t;
+
+class Config {
+public:
+    /// \brief Returns the name of the config file that was used to launch the server.
+    virtual fs::path getConfigFilename() const = 0;
+
+    /// \brief returns a config option of type std::string
+    /// \param option option to retrieve.
+    virtual std::string getOption(config_option_t option) = 0;
```
ytimenkov, line 171: Do you want to take an opportunity and make them `const`? So that all other methods can take `shared_ptr<const Config>`? No one really changes them, right?

src/config/config_manager.h
```diff
@@ -185,45 +60,40 @@ class ConfigManager {
     virtual ~ConfigManager();
 
     /// \brief Returns the name of the config file that was used to launch the server.
```
ytimenkov, line 62: IMHO copying comment from an interface usually leads to inconsistency: base class already contains info what this method should do, therefore `override` is sufficient, especially in this case.

src/main.cc
```diff
@@ -393,6 +394,7 @@ int main(int argc, char** argv, char** envp)
                     }
 
                     ///  \todo fix this for SIGHUP
+                    auto config = std::static_pointer_cast<Config>(configManager);
```
ytimenkov, line 397: This should not be needed: `shared_ptr` can be constructed from a derived class so `make_shared<Server>(configManager)` should work.

~I guess you ran into issues because `Server` accepts a reference to `shared_ptr` which is not the same thing. I would change `Server` constructor to take `shared_ptr` by value and then move into a member. It will also do one less copy,..~

Apparently it does accept by value, so I see no reason why it won't work....

pamapa:
@ytimenkov 
Good input! Please provide followup a patch.

PR #812, opened 2020-04-19
Feature/simplify storage interface

pamapa:


pamapa:
this is something we might should merge after 1.5 is release?

Issue #807, opened 2020-04-18
Browsing playlists not working with some clients


zimj0:
I build gerbera from source and am having issues browsing playlists when using the vlc player or BubbleUPnP as clients.
Vlc simply jumps to the next element.
BubbleUPnP shows a message: "The media server sent invalid data."

Both clients worked without issue with gerbera v1.4.0

[gerbera.log](https://github.com/gerbera/gerbera/files/4496266/gerbera.log)

pamapa:
What sticks out is this:
```
&lt;upnp:album&gt;Hallo Frühling&lt;/upnp:album&gt;
```
The **ü** is not encoded...

whyman:
This looks like #749 to me

pamapa:
i am able to reproduce when i use **ü** in title

pamapa:
I am trying to get a fix upstream:
https://github.com/zeux/pugixml/issues/342

pamapa:
if upstream takes too long i can provide a workaround solution

pamapa:
does not fully work

pamapa:
it has nothing to do with the ä, i was looking into the wrong direction.

@whyman  
Your guess is correct #749, when i temporary skip `p:artist@role[AlbumArtist]` it start working again. So we need to fix this.

pamapa:
@zimj0
should be fixed, can you try with latest master?

zimj0:
Can confirm

PR #815, opened 2020-04-19
correctly handle upnp:artist@role[AlbumArtist]

### LINKED ISSUE #749: upnp:artist@role[AlbumArtist] breaks folder display on clients

pamapa:
fixes #749 and #807

### DIFF
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -80,12 +80,10 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, bool re
         auto item = std::static_pointer_cast<CdsItem>(obj);
 
         auto meta = obj->getMetadata();
-
-        std::string key;
         std::string upnp_class = obj->getClass();
 
         for (const auto& it : meta) {
-            key = it.first;
+            std::string key = it.first;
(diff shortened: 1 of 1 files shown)


whyman:
Yay, I was meaning to do this patch, but time was not my friend this weekend.

Issue #763, opened 2020-03-27
Add item OR Add Autoscan Item


hornetster:
Can anyone explain the difference between the above items?
I have found that, with the top directory selected on the left (ie Music), that selecting "Add Autoscan Item" seems to be hit and miss on what it Caches, whereas, selecting "Add Item" seems to add (at least) most items.
Previously I have used the "Add Autoscan Item" and found that a lot of files are missed, but thought that this should be the correct way to add a permanent item that will always be updated when new/different files are added???
Thanks.

whyman:
Add autoscan item will do an initial scan in the exact same way as add item (same code).

Then it will also register either a timed rescan or add an inotify watch on the directory as well.

hornetster:
But, as I said, the Add Autoscan, never seemed to complete a full scan, whereas the Add Item, DID do a full scan...? Should I do the Add Item, followed by Add Autoscan?

whyman:
There should be no difference, as I say, its the same code.

Can you post a debug log comparing the two?

hornetster:
To do the debug, would have to reset to defaults, I would think?? Easiest way to reset to default?
Thanks.

runderwo:
I also find that adding a directory as autoscan sets the notify watches but precludes it being initially scanned.  Looking at the code now.

runderwo:
Looks like there are two problems with inotify autoscan:
1. If the autoscan entry didn't previously exist in the database, an initial scan would be necessary, but there is no mechanism to trigger an initial scan
2. Items which were added while Gerbera was down or restarting will never be picked up because of the lack of an initial scan.
I am going to try treating inotify autoscan dirs as timed ones for the first run at startup.

PR #817, opened 2020-04-20
Always scan inotify autoscan locations after startup

runderwo:
This avoids gaps and handles the initial state as well.  Fixes https://github.com/gerbera/gerbera/issues/763


whyman:
Hi @runderwo, thanks for the patch!

Issue #683, opened 2020-02-16
Binding to specific interfaces


johnlane:
I have a server with twi nics in it. I would like to bind the Gerbera DLNA service to both of these interfaces but have the web server only on one of them.

Can I bind to multiple interfaces and can I specify different interfaces for the webserver ?

I've tried using `0.0.0.0` unsuccessfully.

whyman:
You can only use a single interface. Both SSDP (Broadcast) and the webserver will both use that interface.

whyman:
As for a feature request I am unsure if libupnp supports it.

What is your use case for the multi interface setup out of interest?

johnlane:
I have a network on which I put stuff I don't have control over (basically anything I don't have root access to). I have another network with trusted stuff. I want to serve content to both networks but only have the webserver on the trusted network.

whyman:
Its a problem for libupnp really. We use the same mechanism to serve the web-ui as the we do the media content so not sure its possible without adding our own http server.

neheb:
During my cleanup of the code, I noticed clang throwing warnings with the libupnp headers. Would it make sense to replace libupnp with a C++ alternative like https://www.lesbonscomptes.com/upmpdcli/libnpupnp.html ?

whyman:
Thanks for telling about `libnpupnp`. Certainly sounds interesting to, I would be open to switching.

I dont understand why @medoc decides to use `opensourceprojects.eu` for things instead of Github, it worries me no one will find this library.

neheb:
When fixing several warnings, I noticed in gerbera that pupnp was being compiled as C++. Thus stupid warnings like NULL instead of nullptr...

johnlane:
I have just been comparing with Mediatomb 0.12.1 and can confirm that it binds to `0.0.0.0`, this is with no `<interface>` setting in its config.
```
tcp        0      0 0.0.0.0:50500        
```
Something must have changed in Gerbera to restrict that, possibly an underlying library ?

johnlane:
I have just been comparing with Mediatomb 0.12.1 and can confirm that it binds to `0.0.0.0`, this is with no `<interface>` setting in its config.
```
tcp        0      0 0.0.0.0:50500        
```
Something must have changed in Gerbera to restrict that, possibly an underlying library ?

whyman:
Old versions of libupnp (as bundled in mediatomb) would *always* bind to 0.0.0.0 regardless of configuration 🤦‍♂️

[...]

PR #818, opened 2020-04-21
Quick changes for working with NPUPNP

medoc92:
As mentioned on framagit, a branch supporting building with libnpupnp. The CMakeLists.txt is hacked, the rest is hopefully preserving the existing code when USING_NPUPNP is not set. Still this is more a proof of concept (and a test of libnpupnp) than anything else.

You will need the npupnp git code, the last release was missing a few fields in the API structures.


whyman:
The interface is much nicer!

medoc92:
The main goal was to keep maximum compatibility with the existing API. Almost all the exceptions are places where ixml was exposed, and it's not hard to improve on passing DOM trees around :)

pamapa:
The main questions for me are:
- how big is the community of NPUPNP
- how responsive are they regarding changes
- which other projects are using it
- what about the specific user agent, IP address i did for npupnp

medoc92:
> The main questions for me are:
> 
>     * how big is the community of NPUPNP

It's currently made of exactly one person. This is a relatively small library, which I don't expect to need a lot of changes. Of course you should also count the people maintaining libmicrohttpd, libcurl and expat as these replace function which was previously implemented inside the UPnP base library.

>     * how responsive are they regarding changes

Now is the right moment to express the need for changes, because there is little client code and the library is not packaged.  I'd be quite interested by testing and comments about the interfaces used by application code for customising behaviour for specific remote end points (either devices or control points), as this has been the main area of changes and issues with pupnp lately. It would be nice to be able to implement what is needed now so that it can be kept stable further on.

>     * which other projects are using it

It's currently only used by upmpdcli which is a quite widely used mpd-based renderer, and upplay, which is a minor control point which largely started as a validating tool for upmpdcli. By the way, this is the main motivation for the gerbera/npupnp branch right now: I wanted to validate how the library would work with another device application. This has been very useful in wrapping up the webserver  functions, which upmpdcli does not use much.

>     * what about the specific user agent, IP address i did for npupnp

I'm not sure that I understand this part, but as far as I can see, the lib now implements every function  needed by the current Gerbera.

To be clear, I am not suggesting that it would  be a good idea for Gerbera to switch to npupnp at this point. This is more a request for comments, so that I can finalize the interfaces and make sure that the lib remains a possible option further on.

neheb:
Somewhat related: https://github.com/gerbera/gerbera/issues/683

I personally noticed upmpdcli using this after I updated libupnp in OpenWrt to 1.12 and it not working.

My comment about libupnp being compiled as C++ instead of C (missing extern "C" maybe) still stands. No idea if it's an issue here.

npupnp uses autotools, making it somewhat hard to run modern LLVM based tools.

medoc92:
> Somewhat related: #683

The current versions of libupnp and libnpupnp  only support binding to a single interface. Changing this has been on my todo for some time. It is encouraging to know that an older version of the lib would somewhat work with multiple interfaces, but I think that this was restricted for a reason (my guess would be incorrect SSDP behaviour), and fixing it is probably not trivial (the bound ip address is currently stored as a global).

> I personally noticed upmpdcli using this after I updated libupnp in OpenWrt to 1.12 and it not working.

I never tried upmpdcli with anything older than upnp 1.6, and this is single-interface to the best of my knowledge.

> My comment about libupnp being compiled as C++ instead of C (missing extern "C" maybe) still stands. No idea if it's an issue here.

I am not sure if I understand the issue. Do you mean libnpupnp ? This is definitely C++, not "compiled as C++".

> npupnp uses autotools, making it somewhat hard to run modern LLVM based tools.

Could you elaborate on this ?

neheb:
> 
> 
> > Somewhat related: #683
> 
> The current versions of libupnp and libnpupnp only support binding to a single interface. Changing this has been on my todo for some time. It is encouraging to know that an older version of the lib would somewhat work with multiple interfaces, but I think that this was restricted for a reason (my guess would be incorrect SSDP behaviour), and fixing it is probably not trivial (the bound ip address is currently stored as a global).
> 
> > I personally noticed upmpdcli using this after I updated libupnp in OpenWrt to 1.12 and it not working.
> 
> I never tried upmpdcli with anything older than upnp 1.6, and this is single-interface to the best of my knowledge.
> 
> > My comment about libupnp being compiled as C++ instead of C (missing extern "C" maybe) still stands. No idea if it's an issue here.
> 
> I am not sure if I understand the issue. Do you mean libnpupnp ? This is definitely C++, not "compiled as C++".
libupnp as currently used by gerbera. Not your library.
> 
> > npupnp uses autotools, making it somewhat hard to run modern LLVM based tools.
> 
> Could you elaborate on this ?

As an example, clang-tidy. meson generates a compile_commands.json file that is used by the run-clang-tidy python script (among other things). This allows easy static analysis with clang-tidy.

gerbera uses CMake which uses a different though somewhat similar mechanism: cmake -DCMAKE_CXX_CLANG_TIDY clang-tidy ...

medoc92:
> > I am not sure if I understand the issue. Do you mean libnpupnp ? This is definitely C++, not "compiled as C++".
> 
> libupnp as currently used by gerbera. Not your library.

Understood.

> > > npupnp uses autotools, making it somewhat hard to run modern LLVM based tools.
> > 
> > 
> > Could you elaborate on this ?
> 
> As an example, clang-tidy. meson generates a compile_commands.json file that is used by the run-clang-tidy python script (among other things). This allows easy static analysis with clang-tidy.
> 
> gerbera uses CMake which uses a different though somewhat similar mechanism: cmake -DCMAKE_CXX_CLANG_TIDY clang-tidy ...

Ok. There is an app for this :) https://github.com/rizsotto/Bear

medoc92:
I have updated the npupnp code for supporting multiple network interfaces. This is still very much a work in progress, but it seems to work for me in simple IPV4 situations (e.g. the device has separate eth and wifi interfaces, on disjoint networks). If someone wants to give it a try, it needs more testing.

Use a space-separated parameter to UpnpInit2(), or "*" for all interfaces.

whyman:
Maybe that should be a Upnpinit3 that takes a vector to make it explicit?

medoc92:
I have added an overloaded UpnpInit2() which takes a vector input.

As far as I can see the multi-interfaces support is working fine with IPV4, including:
- Binding to a single address when the old UpnpInit() is called
- Rejecting webserver connections outside of the configured addresses  (the webserver binds to INADDR_ANY).

The IPV6 support is supposedly present, but untested.

There is no support for interfaces appearing or disappearing at the moment. Maybe one day.
Eagerly wishing for testers :)

neheb:
This is probably a long shot but...

Any chance of getting rid of libexpat and using pugixml? It seems gerbera 1.5.0 switched to the latter. Having gerbera depend on two XML libraries is not quite good for embedded.

medoc92:
I pushed an update which uses a small internal parser that I had around, and does not need libexpat. This seems to work fine, the base UPnP protocol XML elements do not need a fancy parser. This retains the capability to build with expat with configure --enable-expat
pugixml is a very different beast because it's a DOM, not SAX parser, and using it, while not impossible of course, would be a much bigger job.

neheb:
Ah OK. Thanks for the clarification.

neheb:
Tested under Docker. IPv6 does not work.

neheb:
I see this:

```
2020-06-05 02:46:38  debug: [server.cc:119] run(): Initialising libupnp with interface: '', port: 49152
```

Hmmm...

medoc92:
How can I reproduce this ? I'm not too surprised that IPV6 does not work because I did not really test it. I guess it's time to debug this.

neheb:
https://stackify.com/docker-build-a-beginners-guide-to-building-docker-images/ - go down to building image. There's already a Dockerfile file in the gerbera repo.

I use portainer to set everything up but the instructions should be the same.

neheb:
Here's the exact branch I used: https://github.com/neheb/gerbera/tree/npupnp

ytimenkov:
@medoc92 
I understand that you're trying to maintain compatibility with pupnp, but for C++ I appreciate composability. For example:
1. Ability to use another XML serialization library (as mentioned)
2. Ability to plug a custom web server.
3. Integration with external event loops / thread pools (like asio, GCD, libuv)..

(Although at the same time it's nice to have some default implementation).

medoc92:
Ip v6 should be working in libupnp now, single interface for now.

neheb:
I’ll test when I can.

medoc92:
> @medoc92
> I understand that you're trying to maintain compatibility with pupnp, but for C++ I appreciate composability. For example:
> 
>     1. Ability to use another XML serialization library (as mentioned)

Using another SAX handler would certainly be feasible. DOM would need a lot of rewrite for no good reason, so no.

>     2. Ability to plug a custom web server.

You can already use an external web server for streaming or serving images etc. Replacing libmicrohttpd for the internal uses (SOAP and GENA) would be quite a lot of work. Do you think of any specific candidate ?

>     3. Integration with external event loops / thread pools (like asio, GCD, libuv)..

I guess that plugging another thread pool implementation would be feasible.

> (Although at the same time it's nice to have some default implementation).

neheb:
Please no boost. That thing is huge.

ytimenkov:
> Please no boost. That thing is huge.

Yeah, from 30 years ago when it was the easiest way to ship libraries: you just get boost and use what you want/need. Now things tend to be more modularized and independent. Especially when package managers get more traction.

But why "huge" is a problem? Are you limited in space on your dev machine?

neheb:
I package gerbera as a replacement for miniDLNA for OpenWrt. Gerbera is already 10x bigger as miniDLNA suffers from multiple parsing bugs as well as being relatively inactive. I’d like for gerbera to not get bigger.

ytimenkov:
I think it's quite common to think that a "huge" library results in "huge" binary. Sure, there could be some things implemented not efficiently, but not all of them.

And moreover, if you need some functionality, you get increase in the size, regardless whether you implement it by hand or pull from a library.

One example is I find boost unit test library in recent versions the best, it definitely beats gtest (it less verbose and has universal test macro which splits a==b into pieces, so no need to use ugly _eq, _ne, etc) and catch2 started catching up only recently (they didn't have data tests).
This would have zero effect on the binary size.

As for boost.asio for example, I don't know how much current tasks in gerbera take in term of binary size, but there is quite a bit of copy paste and iirc a couple similar implementations.

neheb:
In OpenWrt, libraries are typically linked dynamically. For gerbera, I made sure most of them get linked statically for size reasons.

Having shared libraries results in a significant size increase. -flto helps slim the size down with static linking.

ytimenkov:
👍 have similar thoughts. I feel that it more relevant for C++ libraries and less for C with narrow interface.

I [work on integration](https://github.com/ytimenkov/gerbera/blob/feature/conan/doc/conan.rst) with Conan package manager, it should make it one-liner to rebuild everything static with lto and even cross compile. If you want me to check some cases file an issue there.

neheb:
From the container:

```
bash-5.0# netstat -an
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       
tcp        0      0 :::49152                :::*                    LISTEN      
udp        0      0 127.0.0.1:46931         0.0.0.0:*                           
udp        0      0 0.0.0.0:1900            0.0.0.0:*                           
Active UNIX domain sockets (servers and established)
Proto RefCnt Flags       Type       State         I-Node Path
```

The way docker is set up, 1900 needs to be done on IPv6 as well. No idea why it's not doing that.

medoc92:
> From the container:
> 
> ```
>[...]
> ```
> 
> The way docker is set up, 1900 needs to be done on IPv6 as well. No idea why it's not doing that.

Are you using the latest npupnp code ? 

Also I'd like to try, but it seems I can't build on ubuntu bionic:

````
jftp$ cmake -v CMakeLists.txt 
CMake Error at cmake/FindFilesystem.cmake:227 (message):
  Cannot Compile simple program using std::filesystem
````

**Comment by neheb** [comment]
> > From the container:
> > ```
> > [...]
> > ```
> > 
> > 
> > The way docker is set up, 1900 needs to be done on IPv6 as well. No idea why it's not doing that.
> 
> Are you using the latest npupnp code ?
yes
> 
> Also I'd like to try, but it seems I can't build on ubuntu bionic:
> 
> ```
> jftp$ cmake -v CMakeLists.txt 
> CMake Error at cmake/FindFilesystem.cmake:227 (message):
>   Cannot Compile simple program using std::filesystem
> ```

try with cmake -DCXX_FILESYSTEM_NO_LINK_NEEDED=OFF

**Comment by medoc92** [comment]
I installed gcc-8 and tried with this. Still no luck with std::filesystem:
With CXX=g++-8 CXXFLAGS=-std=c++17 here is what I get:
````
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED - Success
CMake Error at cmake/FindFilesystem.cmake:209 (target_compile_features):
  Cannot specify compile features for imported target "std::filesystem".
  CMakeLists.txt:272 (find_package)


CMake Error at cmake/FindFilesystem.cmake:215 (target_link_libraries):
  Cannot specify link libraries for target "std::filesystem" which is not
  built by this project.
  CMakeLists.txt:272 (find_package)


-- Configuring incomplete, errors occurred!
````

Bailing out. Meanwhile, upmpdcli is happily advertising on ipv6 (2 instances):

````
netstat -an | grep 1900
udp        0      0 0.0.0.0:1900            0.0.0.0:*                          
udp        0      0 0.0.0.0:1900            0.0.0.0:*                          
udp6       0      0 :::1900                 :::*                               
udp6       0      0 :::1900                 :::*             
````

**Comment by neheb** [comment]
Weird. No idea why CMake fails.

**Comment by ytimenkov** [comment]
Check your cmake version. Probably it doesn't support compile features on imported libraries.
Best shot is 3.14. however gerbera has minimal requirement 3.8... should have failed sooner...

**Comment by medoc92** [comment]
cmake is 3.10.2

I finally managed to build after working around spdlog version issues, std::filesystem detection and some others. 

Gerbera listens on tcp6 just fine, so the issue looks to be docker-related. If you are indeed running the latest npupnp commit "Test and fix IPV6 operation. Single interface only for now. ", maybe it's a docker-related issue.

**Comment by neheb** [comment]
> cmake is 3.10.2
> 
> I finally managed to build after working around spdlog version issues, std::filesystem detection and some others.
> 
> Gerbera listens on tcp6 just fine. If you are indeed running the latest npupnp commit "Test and fix IPV6 operation. Single interface only for now. ", maybe it's a docker-related issue.

Could be. I'll stick to running gerbera locally for now.

**Comment by medoc92** [comment]
Rosen Penev writes:
 >     cmake is 3.10.2
 >    
 >     I finally managed to build after working around spdlog version issues,
 >     std::filesystem detection and some others.
 >    
 >     Gerbera listens on tcp6 just fine. If you are indeed running the latest
 >     npupnp commit "Test and fix IPV6 operation. Single interface only for now.
 >     ", maybe it's a docker-related issue.
 >    
 > Could be. I'll stick to running gerbera locally for now.

I meant a libnpupnp issue related to running in Docker, not a Docker issue
of course.

**Comment by neheb** [comment]
I'm considering using this for OpenWrt as libupnp does not compile with GCC10. Anything to watch out for?

**Comment by neheb** [comment]
```
root@OpenWrt:/# netstat -an | grep 1900
udp        0      0 0.0.0.0:1900            0.0.0.0:*                           
udp        0      0 :::1900                 :::*                                
root@OpenWrt:/# netstat -an | grep 49152
tcp        0      0 0.0.0.0:49152           0.0.0.0:*               LISTEN 
```

Wonder if that's normal...

I'll try npupnp 4.0.6 next.

**Comment by medoc92** [comment]
> I'm considering using this for OpenWrt as libupnp does not compile with GCC10. Anything to watch out for?

Yes, wait until I've fixed this: https://framagit.org/medoc92/npupnp/-/issues/4

About the netstat output: I think that the single listen on :::49152 is normal because microhttpd uses a single socket for ipv4/6. It looks mostly the same on my system (except that it's listed as tcp6), and things work as expected.

**Comment by neheb** [comment]
> 
> 
> > I'm considering using this for OpenWrt as libupnp does not compile with GCC10. Anything to watch out for?
> 
> Yes, wait until I've fixed this: https://framagit.org/medoc92/npupnp/-/issues/4
> 
> About the netstat output: I think that the single listen on :::49152 is normal because microhttpd uses a single socket for ipv4/6. It looks mostly the same on my system (except that it's listed as tcp6), and things work as expected.

The upgrade to 4.0.6 from .5 also seems to have upgraded the 49152 socket to ipv6. Note that the gerbera message mentions an IPv4 address.

**Comment by medoc92** [comment]
4.0.6 was meant to fix the ipv6 function, so this seems logical. I'm not sure of the meaning of udp/tcp being listed instead of udp6/tcp6 with an ipv6 address.

I just released libupnpp 0.19.2 and libnpupnp 4.0.7. The latter hopefully fixes the above vulnerability. I just hope that I did not break anything while doing this...

**Comment by neheb** [comment]
Sounds good.

**Comment by neheb** [comment]
Well I just discovered something unfortunate...

```
root@OpenWrt:/mnt# netstat -an | grep 1900
udp        0      0 0.0.0.0:1900            0.0.0.0:*                           
udp        0      0 0.0.0.0:1900            0.0.0.0:*                           
udp        0      0 :::1900                 :::*                                
udp        0      0 :::1900                 :::*
```

It seems gerbera conflicts with miniupnpd...

**Comment by neheb** [comment]
> 4.0.6 was meant to fix the ipv6 function, so this seems logical. I'm not sure of the meaning of udp/tcp being listed instead of udp6/tcp6 with an ipv6 address.

busybox

**Comment by medoc92** [comment]
> Well I just discovered something unfortunate...
> 
> ```
> root@OpenWrt:/mnt# netstat -an | grep 1900
> udp        0      0 0.0.0.0:1900            0.0.0.0:*                           
> udp        0      0 0.0.0.0:1900            0.0.0.0:*                           
> udp        0      0 :::1900                 :::*                                
> udp        0      0 :::1900                 :::*
> ```
> 
> It seems gerbera conflicts with miniupnpd...

I don't think it's a conflict, they're just both listening to the multicast address.

**Comment by neheb** [comment]
Seems this no longer builds with libnpupnp 4.0.8:

```
/home/mangix/devstuff/openwrt/build_dir/target-mips_74kc+dsp2_musl/gerbera-1.5.0/src/upnp_cds.cc: In member function 'void ContentDirectoryService::doBrowse(const std::__1::unique_ptr<ActionRequest>&)':
/home/mangix/devstuff/openwrt/build_dir/target-mips_74kc+dsp2_musl/gerbera-1.5.0/src/upnp_cds.cc:86:29: error: 'UPNP_SOAP_E_INVALID_ARGS' was not declared in this scope; did you mean 'UPNP_E_INVALID_PARAM'?
   86 |         throw UpnpException(UPNP_SOAP_E_INVALID_ARGS,
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~
      |                             UPNP_E_INVALID_PARAM
make[5]: *** [CMakeFiles/libgerbera.dir/build.make:795: CMakeFiles/libgerbera.dir/src/upnp_cds.cc.o] Error 1

medoc92:
Oops, I restored the definitions and issued 4.0.9.


Rosen Penev writes:

neheb:
So 1.6 got released.

What's the status on this?

whyman:
Closing in favour of #952

PR #820, opened 2020-04-21
Ensure all xml-namespaces are declared (fixes #819)

### LINKED ISSUE #819: Some clients do not show videos
VLC and Windows Explorer do not show videos if [need some tests for the exact reason]
- caption file does not exist or 
- sec:CaptionInfoEx is present in the BrowseResponse.

No proper solution other than commenting out line 813 in upnp_xml.cc, yet.

May also be the reason for #707 .

KarlStraussberger:
Namespace is still used in UpnpXMLBuilder::renderCaptionInfo for videos.

### DIFF
--- a/src/upnp_cds.cc
+++ b/src/upnp_cds.cc
@@ -166,7 +166,7 @@ void ContentDirectoryService::doSearch(const std::unique_ptr<ActionRequest>& req
     didl_lite_root.append_attribute(XML_NAMESPACE_ATTR) = XML_DIDL_LITE_NAMESPACE;
     didl_lite_root.append_attribute(XML_DC_NAMESPACE_ATTR) = XML_DC_NAMESPACE;
     didl_lite_root.append_attribute(XML_UPNP_NAMESPACE_ATTR) = XML_UPNP_NAMESPACE;
-    request->GetQuirks()->appendSpecialNamespace(&didl_lite_root);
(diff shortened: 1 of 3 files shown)


whyman:
@pampa what do you think is going on here?

whyman:
If you don't comment out the quirk but add the extra namespace, are players still happy?

pamapa:
The changed code was previously within a if SAMSUNG_HACK. So it was only enabled for those with that flag on.

@KarlStraussberger
For what kind of device are you enabling that? And are you sure it wont bite other players?

BTW: When looking at other server code (minidlna), i can see, that they enable it always, independed of a specific client, so i think this should be ok :-)

Please do not only comment the quirk, do completely remove that function and call. Lets keep the code clean.

pamapa:
I can confirm that for VLC it help me too

KarlStraussberger:
Left the old code for discussion. Now it's gone

KarlStraussberger:
Have been testing with my LG TV, Kodi, VLC, Windows 👎 , Foobar2000, Sony DVD player, HEOS sound system so far

pamapa:
cool, nice work

whyman:
> Have been testing with my LG TV, Kodi, VLC, Windows -1 , Foobar2000, Sony BluRay player, HEOS sound system so far

Awesome work, very much appreciated!

PR #824, opened 2020-04-22
Avoid server abort when tracknumber starts with non-digit character

KarlStraussberger:
If tracknumber starts with an non-digit character std::stoi throws an std::invalid_argument exception.

Moved the check to the existing tools function "stoi_string" making it available for other places with same issues.


whyman:
Good catch, thanks!

PR #828, opened 2020-04-23
avoid exceptions

KarlStraussberger:
replaced more std::stoi by stoi_string to avoid exception when reading unformatted data. May be a bit slower but is much safer


whyman:
I think I prefer this to just blow up. The DB schema should enforce good data and if not then we probably can't handle that case anyway...

KarlStraussberger:
I encountered a crash with malformed track numbers. So we should at least handle track number safely and implement try-catch-blocks on the outer levels (search, browse, import) for proper logging of errors.

whyman:
Perhaps we should include the `setTrackNumber` change as the data is externally sourced, but the IDs are internal, so should be safe...

PR #829, opened 2020-04-25
clang-tidy cleanups

neheb:
Suggested by useStlAlgorithm

I'm not sure these are correct. Would love someone to check these.

### DIFF
--- a/src/util/upnp_clients.cc
+++ b/src/util/upnp_clients.cc
@@ -181,7 +181,7 @@ void Clients::getInfo(const struct sockaddr_storage* addr, const std::string& us
     }
 
     *ppInfo = info;
-    log_debug("client info: {} '{}' -> '{}'", sockAddrGetNameInfo((struct sockaddr*)addr), userAgent, (*ppInfo)->name);
+    log_debug("client info: {} '{}' -> '{}'", sockAddrGetNameInfo(reinterpret_cast<const struct sockaddr*>(addr)), userAgent, (*ppInfo)->name);
 }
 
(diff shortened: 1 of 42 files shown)


neheb:
Removed algorithm commit and added others. That commit was totally broken anyway...

src/util/upnp_clients.cc
```diff
@@ -156,7 +156,7 @@ void Clients::getInfo(const struct sockaddr_storage* addr, const std::string& us
     // by addClientByDiscovery
     if (!found) {
         // always return something, 'Unknown' if we do not know better
-        assert(clientInfo[0].type == ClientType::Unknown);
+        static_assert(clientInfo[0].type == ClientType::Unknown);
```
whyman, line 159: I think we want to assert at runtime still

neheb: Why?

neheb:
Rebased.

src/util/upnp_clients.cc
whyman: Because the check is a runtime one and makes no sense at compile time?

neheb: The way I read this is that it asserts that the first struct member at 0 is Unknown, which it is. Either that or the check is wrong.

whyman: Yes, but why would we want to check that at compile time?

neheb: It’s a const struct...

PR #830, opened 2020-04-25
Check the other dynamic_casts in matroska_handler

runderwo:
More resiliency for badly constructed files.


whyman:
LGTM, thanks!

PR #831, opened 2020-04-25
Add title param to external transcoder

runderwo:
This allows us to pass metadata from the import script to the transcoder via the object title.  It's not much, but it's enough to e.g. signal a resume offset for otherwise non-seekable content.

Also, fix a string length bug that was corrupting the range parameter


whyman:
Nice! Thanks!

PR #833, opened 2020-04-27
 Update documentation for PS4 disconnects (closes #821)

### LINKED ISSUE #821: media player stops after 30min
My PS4 will always stops playing after 30 min.
This is because default value of gerbera setting `alive` is `1800`.

I changed this value to `86400` and all movies play nicely.

KarlStraussberger:


whyman:
Perhaps add a section on the compatible devices page for PS4 too?

Issue #775, opened 2020-04-05
remove content type/mime mapping


pamapa:
Directly working with mime types makes everything must easier. In case a specific client needs a quirk we can make use of this #752 feature.

- [ ] libmagic needs to become from optional to a hard build requirement.

- [ ] removed configuration properties:
  - CFG_IMPORT_MAPPINGS_IGNORE_UNKNOWN_EXTENSIONS
  - CFG_IMPORT_MAPPINGS_EXTENSION_TO_MIMETYPE_CASE_SENSITIVE
  - CFG_IMPORT_MAPPINGS_EXTENSION_TO_MIMETYPE_LIST
  - CFG_IMPORT_MAPPINGS_MIMETYPE_TO_UPNP_CLASS_LIST
  - CFG_IMPORT_MAPPINGS_MIMETYPE_TO_CONTENTTYPE_LIST

@whyman
If libmagic can not be a hard build requirement, an alternativewould be to simply use the file extension and work with that. Like this
```
bool isImage(const fs::path& filename)
{
  static std::vector<std::string> v{".jpg",  ".jpeg"};
  return std::find(v.begin(), v.end(), filename.extension()) != v.end();
}
```

whyman:
@pamapa I think its fine to have magic as build requirement, lets keep our code-paths as simple as possible.

PR #834, opened 2020-04-29
Extend configuration for autodetected clients

KarlStraussberger:
**Each automatic mechanism is only as good as its manual switch.**

This patch add a `<clients>`-section to config.xml, so autodetected clients can be configured based on current quirks. It was inspired by similar mechanisms in other MediaServers.
Currently, clients can be detected by IP and/or UserAgent and given the same properties as autodetection does. So this can be used if a any device is not detected correctly. IP+UserAgent configuration is especially helpful if more than one media player is used on a PC with different behaviour.

It may become more useful when the number of available Quirks increases. At the moment it can be used to bring back special hacks done via protocolInfo in Gerbera 1.4 to individual devices.


KarlStraussberger:
I will add some documentation as soon as the patch comes close to merging into master. For the time beeing, here's some example config

```
<clients enabled="yes">
		<client ip="192.168.1.12" userAgent="foobar2000" clientType="StandardUPnP"></client>
		<client ip="192.168.1.12" userAgent="Microsoft-DLNA" clientType="SamsungBDP"></client>
		<client ip="192.168.1.35" clientType="StandardUPnP"></client>
</clients>
```

pamapa:
@KarlStraussberger 
The whole idea about the quirks, is to get ride of user specific configurations. The main idea is to use the detection of the user agent to make gerbera work the same way for all users. This also makes it very easy for new users of gerbera.

I agree, that this support is not in fully place yet and that way maybe not visible. So i explain what my plan is :-)

Instead of a config section, there is already `src/util/upnp_clients.cc/h` in place. In there we are collection a list of known/supported devices. Each device entry can use quirk flags. Currently i have only added one quirk flag: `QUIRK_FLAG_SAMSUNG`. But more will come.

once #775 is done, we can add client specific mime-mappings (e.g. old Samsung TVs need matroska mime quirk "x-matroska" -> "mkv") simply by adding a new quirk flag for this behavior.


e.g. client entry:
```
// User-Agent: DLNADOC/1.50 SEC_HHP_[TV] Samsung Q7 Series (49)/1.0
    {
        "Samsung Series [Q] TVs",
        ClientType::SamsungSeriesQ,
        QUIRK_FLAG_SAMSUNG,
        ClientMatchType::UserAgent,
        "SEC_HHP_[TV] Samsung Q" },
```

If we now later find out that the device needs a special treatment, we can simply add a new flag and handle this the same way for all users.

So please add your devices in `src/util/upnp_clients.cc/h` iff they will need later a special treatment. Also note that all devices that work out of the box, do not needed to be added, also it would not harm if they were. They will match with the first (unknown) or the last device (StandardUPnP) in the list.

pamapa:
In short: I wont merge this. Sorry. But many thanks for your efforts in all other areas! It would be awesome if you could help up building a working client table in `upnp_clients.cc/h` after 1.5.0 is released.

KarlStraussberger:
I fully understand your plan and apreciate the idea that everything works out of the box. As you said it is not fully done and from my experience never will. I doubt you want to release gerbere for each new or undetected device.

That's the idea of my patch. Add support for any device that is not **yet** known - there's a lot out there we will never hear about. With each release of gerbera the list of supported devices can be extended. In the meantime we can support it by manually overriding the autodetection.

It's not a feature for the masses but for the extraordinary and they would have loved it!

pamapa:
ok, i start to understand your motivation.
At the end we also must make sure we can handle all that code paths. The less `if` the better.

What about this idea:
We allow the users to change `static const struct ClientInfo clientInfo[] = {` by configuration. The configuration enties are added to that list. Of course we would also accept an override (configuration wins). Like this the main code paths remain quite the same.

For this we need something like this
<clients enabled="yes">
		<client userAgent="foobar2000" flags=0x00010003></client>
</clients>

Flags is uint32_t containing all needed flags.

The cool thing about this is, that advanced users can try out flags without re-compiling gebera and if it works they could tell us to add additional quirks to the table in the source code.

The `ClientType` is really taught for internal usage only. It should match only devices we have in `upnp_clients.cc/h`. For devices, which will come from the configuration, we should make a special enum value. e.g. `ConfigUPnP`. The important things for future behavior will be quirk flags.

As we have the userAgent information, i do not see the point to add an IP to the configuration...

src/util/upnp_clients.cc
```diff
@@ -146,12 +147,16 @@ void Clients::addClientByDiscovery(const struct sockaddr_storage* addr, const st
 #endif
 }
 
-void Clients::getInfo(const struct sockaddr_storage* addr, const std::string& userAgent, const ClientInfo** ppInfo)
+void Clients::getInfo(std::shared_ptr<Config> config, const struct sockaddr_storage* addr, const std::string& userAgent, const ClientInfo** ppInfo)
```
pamapa, line 150: instead of injecting here the config:

Add an additional function `addClienInfo(ClientInfo* add)` to manipulate `ClientInfo clientInfo[]`.
Probably in server.cc get the configuration and call this function.

src/util/upnp_clients.cc
```diff
@@ -181,7 +186,66 @@ void Clients::getInfo(const struct sockaddr_storage* addr, const std::string& us
     }
 
     *ppInfo = info;
-    log_debug("client info: {} '{}' -> '{}'", sockAddrGetNameInfo((struct sockaddr*)addr), userAgent, (*ppInfo)->name);
+    log_debug("client info: {} '{}' -> '{}' as {}", sockAddrGetNameInfo((struct sockaddr*)addr), userAgent, (*ppInfo)->name, ClientConfig::mapClientType((*ppInfo)->type));
+}
+
+bool Clients::getInfoByConfig(const struct sockaddr_storage* addr, const std::string& userAgent, std::shared_ptr<Config> config, const ClientInfo** ppInfo)
```
pamapa, line 192: far too complex

src/config/client_config.cc
```diff
@@ -0,0 +1,159 @@
...
+    if (id >= list.size()) {
+        log_debug("No such ID {}!", id);
+        return;
+    }
+
+    list.erase(list.begin() + id);
+    log_debug("ID {} removed!", id);
+}
+
+std::string ClientConfig::mapClientType(ClientType clientType)
```
pamapa, line 93: dont use ClientType, please use quirk flags

src/config/client_config.cc
```diff
@@ -0,0 +1,159 @@
...
+        break;
+    default:
+        throw_std_runtime_error("illegal clientType given to mapClientType()");
+    }
+    return clientType_str;
+}
+
+ClientType ClientConfig::remapClientType(const std::string& clientType)
+{
+    if (clientType == "BubbleUPnP") {
```
pamapa, line 129: dont use ClientType, please use quirk flags

src/config/config.h
```diff
@@ -190,6 +194,10 @@ class Config {
     /// \param option to retrieve
     virtual std::shared_ptr<AutoscanList> getAutoscanListOption(config_option_t option) = 0;
 
+    /// \brief returns a config option of type AutoscanList
```
pamapa, line 197: copy/past: please adapt AutoscanList

whyman:
@pamapa Nice idea, I think a bitmask is a bit hard for a normal user, maybe let them specify the quirks by name as an array and then build the mask from that.

pamapa:
yes its makes sense with providing names instead of bit masks:

something like this
```
<clients enabled="yes">
		<client userAgent="foobar2000" flags="SAMSUNG|OTHER1|OTHER2"></client>
</clients>
```
this name can be resolved in `src/config/client_config.cc` and `src/util/upnp_clients.cc` still gets an uint32_t.. if name is not recognized: log warning + assume 0

KarlStraussberger:
I'll see what I can make of it.

src/config/client_config.cc
KarlStraussberger: Kept the conversion function for completeness

src/config/client_config.cc
KarlStraussberger: Kept the parsing function for completeness

KarlStraussberger:
Left the option to filter with IP because it allows the user to force flags to a device which otherwise would have been autodected (we add all new clients to the end of the list)

KarlStraussberger:
Left the option to filter with IP because it allows the user to force flags to a device which otherwise would have been autodected (we're addin all new clientInfos to the end of the list).

whyman:
LGTM

As a follow up - I imagine we will need to expose MAC address info for dynamically addressed devices at some point in the future.

KarlStraussberger:
The sooner the better.
I'm thinking about a page on the web interface contaning all known client information. No big deal, inspired by minidlna.

whyman:
> I'm thinking about a page on the web interface contaning all known client information. No big deal, inspired by minidlna.

Sounds nice!

PR #841, opened 2020-05-02
Fix cds external url

Juergen-W:
With mediatomb i was able to import and play cue-files using scripting and transcoding. In gerbera the behavior of handling the object type "external url" seems to be different or has bugs. I fixed these and add a short description for handling cue-files. On mediatomb.cc there were some examples like this.
IMHO using the type external_url is a little dirty hack - the best way should be to have an new object type for transcoding scripts which use parameters to read the source itself...

### DIFF
--- a/doc/compile.rst
+++ b/doc/compile.rst
@@ -108,10 +108,10 @@ The following packages are too old in 16.04 and must be installed from source:
 Build On Ubuntu 18.04
 ~~~~~~~~~~~~~~~~~~~~~
 
-::
 To build gerbera on Ubuntu 18.04 you have to install a newer version of the gcc++ compiler and clang++:
 
 :: 
+
   sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   sudo apt update
   sudo apt upgrade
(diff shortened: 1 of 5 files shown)


whyman:
Not sure whats going on with this PR, it looks like you really only wanted this commit https://github.com/gerbera/gerbera/pull/841/commits/c5b75b360822cc20ad04af11e9fb18767fee8d34 correct?

Juergen-W:
Sorry, i had some trouble with the code checks... (github newbie). Yes, the commit 236cd65 corrects c5b75b3. 
This patch makes an old mediatomb hack (https://sourceforge.net/p/mediatomb/discussion/440750/thread/ef12d37a/) to run. But i'm not sure if this way (a kind of misuse of the object type ext.url) is the best solution. IMHO a cleaner solution is to introduce a new object type that represents a transcoder which accepts a parameterstring as input instead of a media file...

whyman:
@KarlStraussberger You have incorporated the main fix for this issue already I believe? Can we close this or are there other bits?

KarlStraussberger:
Merged scripting cue-files as example into the scripting doc

PR #844, opened 2020-05-04
Call to getnameinfo needs to use the size of the actual address struct

medoc92:
sizeof(struct sockaddr_in6) > sizeof(struct sockaddr)


whyman:
LGTM!, please fix the clang-format lint failure.

PR #846, opened 2020-05-04
Fixed use of uninitialized pthread_mutex.

ytimenkov:
Had spontaneous crashes / assertions in libc. Fortunately was not a low hanging fruit.


whyman:
LGTM, thanks!

PR #847, opened 2020-05-04
remove iconv casting

neheb:
iconv_t is sometimes a pointer and other times an int. Remove casting
to make it work with the latter.


whyman:
Woop, thanks!

PR #850, opened 2020-05-07
Propagate layout init errors

runderwo:
If we ignore these errors, the following can happen:
1. The layout type fails initialization
2. An object is added
3. The object is not added to the virtual layout

After this happens, the object will never again have an opportunity
to be added to the virtual layout unless it is manually deleted
from the database.  This could be observed while developing the
JS import script, for example.

By propagating failure, we ensure that adding the object will fail
and will thus be reattempted later.


whyman:
Thanks, looks sensible.

PR #851, opened 2020-05-07
Avoid nullref exception

KarlStraussberger:
Obviously some change broke this


whyman:
Do we know what is causing a null pointer to get passed to this method?

KarlStraussberger:
No, but I just had a crash

whyman:
Fixed via #858

PR #858, opened 2020-05-10
getAutoscanDirectory may return nullptr

ytimenkov:
`gsl::not_null` anyone? :D

Looks like handling of this `adir` is just patched up whenever a problem comes... For example, why can `CMRescanDirectoryTask` be constructed with `nullptr` if it will do nothing in `run()`? I don't see anyone touches it...

### DIFF
--- a/src/content_manager.cc
+++ b/src/content_manager.cc
@@ -680,7 +680,9 @@ void ContentManager::_rescanDirectory(std::shared_ptr<AutoscanDirectory> adir, c
                     list->erase(objectID);
                 // add a task to rescan the directory that was found
                 auto adir2 = getAutoscanDirectory(objectID);
-                rescanDirectory(adir2, newPath, task->isCancellable());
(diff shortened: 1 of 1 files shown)


whyman:
@KarlStraussberger does this fix the crash you were hitting?

ytimenkov:
The observable effect is that start with fresh db is ok while the second one crashes. I also had paths in the config rather than configured via UI. Haven't looked at what matters.

ytimenkov:
Oh, looks we have 2 PRs fixing the issue 🤓

ytimenkov:
I didn't think much about it but comparing those 2 I like mine more 😊 from function contract standpoint: here we take data from one place and pass it to another. We must ensure that post- and preconditions match.
The other fix (in `rescanDirectory`) means "throw me any crap and I'll try my best to figure out way to handle it".
🤓

PR #853, opened 2020-05-08
Clientlist on ui

KarlStraussberger:
Adds Clients Page on UI. Each client detected is displayed on the page with its knwon details

PR #859, opened 2020-05-11
Client list on ui

KarlStraussberger:
Had to set up the branch again after messing up with my local branch. It replaces #853 and contains #834


gerbera-web/test/client/fixtures/clients-data.json
```diff
@@ -0,0 +1,67 @@
+{
```
whyman, line 1: Looks like this clients-data.json is in two separate places?

gerbera-web/test/client/fixtures/index.html
```diff
@@ -55,6 +58,11 @@ <h4><i class="fa fa-plus"></i> Add some files</h4>
                 <h4><i class="fa fa-files-o"></i> View indexed files</h4>
                 <p>View, update and remove files from Gerbera.</p>
             </a>
+
+            <a id="clientlink" href="javascript:;" onclick="$('#nav-clients').click(); // Replace with hashchange handling">
```
whyman, line 62: I guess you are working on this still?

whyman:
Ive left a couple of comments, but LGTM otherwise.

The clients view image in the doc still shows a number for the quirks?

gerbera-web/test/client/fixtures/index.html
KarlStraussberger: Copy and paste, It's been in the source as well

gerbera-web/test/client/fixtures/clients-data.json
KarlStraussberger: Using the same data for client, e2e (incl screenshots) at the moment. As tests are separate I prefer copies of the same file.

KarlStraussberger:
> 
> 
> Ive left a couple of comments, but LGTM otherwise.
> 
> The clients view image in the doc still shows a number for the quirks?

The text in the image is generated from the mock data which was not yet in shape. chrome-81 stopped working on remote-x sessions :-( Had to downgrade to chrome-80.

Added hostname as well to the UI and image

whyman:
If there are no Flags it should just say "None" I think, instead of 0. Same for if its using "StandardUPnP", it should just say None.

KarlStraussberger:
> 
> 
> If there are no Flags it should just say "None" I think, instead of 0. Same for if its using "StandardUPnP", it should just say None.

I agree with "None" for the flags. But "StandardUPnP" is fine because its what we do then - comply with the standard, except in the manual case where "None" is the better choice.

whyman:
@pamapa Any comment?

pamapa:
@KarlStraussberger 
wow looks really nice, very comfortable with support UI too

pamapa:
qurik flags i see we will sometime need (as found in minidlna):
- FLAG_MIME_AVI_DIVX (video/x-msvideo -> divx or avi)
- FLAG_MIME_AVI_AVI (video/x-msvideo -> avi)
- FLAG_MIME_MATROSKA_MKV (x-matroska -> mkv) (older Samsung TVs)
- FLAG_TITLE_DOT - LG hack: subtitles won't get used unless dc:title contains a dot
- FLAG_TITLE_MAX_23 Asus OPlay reboots with titles longer than 23 characters with some file types
- FLAG_TITLE_WITH_EXT Hyundai hack: Only titles with a media extension get recognized
- ...

Those flags can then be applied one by one by configuration or web ui.
The FLAG_MIME_x flags will be needed when doing #775

whyman:
@KarlStraussberger is this ready to merge?

KarlStraussberger:
Yes, IT US.

src/util/tools.cc
```diff
@@ -1283,6 +1283,19 @@ std::string getAVIFourCC(const fs::path& avi_filename)
 }
 #endif
 
+std::string getHostName(const struct sockaddr* addr)
+{
+    char hoststr[NI_MAXHOST];
+    char portstr[NI_MAXSERV];
+    int len = addr->sa_family == AF_INET6 ? sizeof(struct sockaddr_in6) : sizeof(struct sockaddr_in);
+    int ret = getnameinfo(addr, len, hoststr, sizeof(hoststr), portstr, sizeof(portstr), NI_NOFQDN);
```
ytimenkov, line 1291: @KarlStraussberger 
This may look like a nice feature but `getnameinfo` is blocking and in case of DNS service unavailable this can freeze the thread for a few minutes.

I would be really cautious about using such innocent-looking non-async APIs all over the place...

ytimenkov: Oh, and in case of failure you just return garbage :(

PR #866, opened 2020-05-14
several fixes

neheb:


src/web/directories.cc
```diff
@@ -96,13 +96,7 @@ void web::directories::process()
             || (exclude_config_dirs && startswith(filepath.filename(), ".")))
             continue;
 
-        bool hasContent = false;
-        for (auto& subIt : fs::directory_iterator(path)) {
-            if (!subIt.is_directory(ec) && !isRegularFile(subIt.path(), ec))
-                continue;
-            hasContent = true;
-            break;
-        }
```
whyman, line 105: Have you not lost a subdirectory iteration here?

neheb: Removed.

PR #876, opened 2020-05-18
no inline + size savings

neheb:
inline doesn't seem useful.

std::string is default allocated to 0. Not needed in many cases.

### DIFF
--- a/src/config/config_manager.cc
+++ b/src/config/config_manager.cc
@@ -1653,18 +1653,12 @@ std::shared_ptr<AutoscanList> ConfigManager::createAutoscanListFromNode(const st
             continue;
         }
 
-        ScanMode mode;
         std::string temp = child.attribute("mode").as_string();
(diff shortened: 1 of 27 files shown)


src/config/config_manager.cc
```diff
@@ -348,13 +348,7 @@ void ConfigManager::load(const fs::path& filename, const fs::path& userHome)
...
 #endif // SQLITE3
 
-    std::string dbDriver;
-    if (sqlite3_en == "yes")
-        dbDriver = "sqlite3";
-
-    if (mysql_en == "yes")
-        dbDriver = "mysql";
-
+    std::string dbDriver = (sqlite3_en == "yes") ? "sqlite3" : "mysql";
```
whyman, line 351: This is not the same behaviour

neheb: Removed.

neheb:
Added a fix for the previous std::copy changes.

PR #882, opened 2020-05-23
std::algorithm conversions

neheb:

Did these manually. Not sure how many are correct.

### DIFF
--- a/src/autoscan.cc
+++ b/src/autoscan.cc
@@ -93,11 +93,7 @@ int AutoscanList::add(const std::shared_ptr<AutoscanDirectory>& dir)
 
 int AutoscanList::_add(const std::shared_ptr<AutoscanDirectory>& dir)
 {
-    std::string loc = dir->getLocation();
-
-    bool err = std::any_of(list.begin(), list.end(), [&](const auto& i) { return loc == i->getLocation(); });
-
-    if (err) {
+    if (std::any_of(list.begin(), list.end(), [loc = dir->getLocation()](const auto& item) { return loc == item->getLocation(); })) {
         throw_std_runtime_error("Attempted to add same autoscan path twice");
     }
 
@@ -137,21 +133,16 @@ std::shared_ptr<AutoscanDirectory> AutoscanList::getByObjectID(int objectID)
 {
     AutoLock lock(mutex);
 
-    for (const auto& i : list) {
-        if (objectID == i->getObjectID())
-            return i;
-    }
-    return nullptr;
+    auto it = std::find_if(list.begin(), list.end(), [=](const auto& item) { return objectID == item->getObjectID(); });
+    return it != list.end() ? *it : nullptr;
 }
 
 std::shared_ptr<AutoscanDirectory> AutoscanList::get(const fs::path& location)
 {
     AutoLock lock(mutex);
-    for (const auto& i : list) {
-        if (location == i->getLocation())
-            return i;
-    }
-    return nullptr;
+
+    auto it = std::find_if(list.begin(), list.end(), [=](const auto& item) { return location == item->getLocation(); });
+    return it != list.end() ? *it : nullptr;
 }
 
 void AutoscanList::remove(size_t id)
@@ -168,36 +159,6 @@ void AutoscanList::remove(size_t id)
 
     list.erase(list.begin() + id);
     log_debug("ID {} removed!", id);
-}
-
-int AutoscanList::removeByObjectID(int objectID)
-{
-    AutoLock lock(mutex);
-
-    for (size_t i = 0; i < list.size(); i++) {
-        if (objectID == list[i]->getObjectID()) {
-            auto dir = list[i];
-            dir->setScanID(INVALID_SCAN_ID);
-            list.erase(list.begin() + i);
-            return i;
-        }
-    }
-    return INVALID_SCAN_ID;
-}
-
-int AutoscanList::remove(const fs::path& location)
-{
-    AutoLock lock(mutex);
-
-    for (size_t i = 0; i < list.size(); i++) {
-        if (location == list[i]->getLocation()) {
-            auto dir = list[i];
-            dir->setScanID(INVALID_SCAN_ID);
-            list.erase(list.begin() + i);
-            return i;
-        }
-    }
-    return INVALID_SCAN_ID;
 }
 
 std::shared_ptr<AutoscanList> AutoscanList::removeIfSubdir(const fs::path& parent, bool persistent)
--- a/src/autoscan.h
+++ b/src/autoscan.h
@@ -77,13 +77,6 @@ public:
     /// \brief removes the AutoscanDirectory given by its scan ID
     void remove(size_t id);
 
-    int removeByObjectID(int objectID);
-
-    /// \brief removes the AutoscanDirectory with the given location
-    /// \param location the location to remove
-    /// \return the scanID, that was removed; if nothing removed: INVALID_SCAN_ID
-    int remove(const fs::path& location);
-
     /// \brief removes the AutoscanDirectory if it is a subdirectory of a given location.
     /// \param parent parent directory.
     /// \param persistent also remove persistent directories.
--- a/src/cds_objects.cc
+++ b/src/cds_objects.cc
@@ -101,11 +101,8 @@ int CdsObject::resourcesEqual(const std::shared_ptr<CdsObject>& obj)
         return 0;
 
     // compare all resources
-    for (size_t i = 0; i < resources.size(); i++) {
-        if (!resources[i]->equals(obj->resources[i]))
-            return 0;
-    }
-    return 1;
+    return std::equal(resources.begin(), resources.end(), obj->resources.begin(),
+        [](const auto& r1, const auto& r2) { return r1->equals(r2); });
 }
 
 void CdsObject::validate()
--- a/src/iohandler/process_io_handler.cc
+++ b/src/iohandler/process_io_handler.cc
@@ -65,21 +65,9 @@ bool ProcListItem::abortOnDeath() const
 
 bool ProcessIOHandler::abort() const
 {
-    bool abort = false;
-
-    if (procList.empty())
-        return abort;
-
-    for (const auto& i : procList) {
-        auto exec = i->getExecutor();
-        if ((exec != nullptr) && (!exec->isAlive())) {
-            if (i->abortOnDeath())
-                abort = true;
-            break;
-        }
-    }
-
-    return abort;
+    return std::any_of(procList.begin(), procList.end(),
+        [=](const auto& proc) { auto exec = proc->getExecutor();
+            return exec != nullptr && !exec->isAlive() && proc->abortOnDeath(); });
 }
 
 void ProcessIOHandler::killAll() const
--- a/src/util/tools.cc
+++ b/src/util/tools.cc
@@ -462,11 +462,9 @@ void dict_decode_simple(const std::string& url, std::map<std::string, std::strin
 std::string mime_types_to_CSV(const std::vector<std::string>& mimeTypes)
 {
     std::ostringstream buf;
-    for (size_t i = 0; i < mimeTypes.size(); i++) {
-        if (i > 0)
-            buf << ",";
-        std::string mimeType = mimeTypes[i];
-        buf << "http-get:*:" << mimeType << ":*";
+    for (const auto& mimeType : mimeTypes) {
+        buf << "http-get:*:" << mimeType << ":*"
+            << ",";
     }
 
     return buf.str();
--- a/src/web/session_manager.cc
+++ b/src/web/session_manager.cc
@@ -163,29 +163,31 @@ std::shared_ptr<Session> SessionManager::createSession(long timeout)
 
 std::shared_ptr<Session> SessionManager::getSession(const std::string& sessionID, bool doLock)
 {
-    std::unique_lock<decltype(mutex)> lock(mutex, std::defer_lock);
+    if (sessions.empty()) {
+        return nullptr;
+    }
+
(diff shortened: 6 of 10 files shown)


src/autoscan.cc
```diff
@@ -174,13 +168,12 @@ int AutoscanList::removeByObjectID(int objectID)
...
 
-    for (size_t i = 0; i < list.size(); i++) {
-        if (objectID == list[i]->getObjectID()) {
-            auto dir = list[i];
-            dir->setScanID(INVALID_SCAN_ID);
-            list.erase(list.begin() + i);
-            return i;
-        }
+    auto it = std::find_if(list.begin(), list.end(), [=](const auto& item) { return objectID == item->getObjectID(); });
+    if (it != list.end()) {
```
ytimenkov, line 172: I'd like (and some agree) it's better to fail fast: no item, return an error rather than doing work in the `if` clause.

Another thing: do you really need to copy `dir`?

neheb: src/autoscan.cc:174:14: error: no member named 'setScanID' in 'std::shared_ptr<AutoscanDirectory>'
        *it->setScanID(INVALID_SCAN_ID);

Failing fast is a good idea.

neheb:
I made some change and now it fails. Oh boy...

src/autoscan.cc
ytimenkov: `(*it)->...` i think it's operator precedence trick.

src/cds_objects.cc
```diff
@@ -101,11 +101,8 @@ int CdsObject::resourcesEqual(const std::shared_ptr<CdsObject>& obj)
         return 0;
 
     // compare all resources
-    for (size_t i = 0; i < resources.size(); i++) {
-        if (!resources[i]->equals(obj->resources[i]))
-            return 0;
-    }
-    return 1;
+    return std::all_of(resources.begin(), resources.end(), [&](const auto& res) //
+        { return res == obj->resources[std::distance(resources.begin(), resources.end())]; });
```
ytimenkov, line 105: Also tip: `distance` is constant but you evaluate it in every iteration and it's O(n). Better do it just once in the beginning:
`[r_to_find = obj->resources [distance(...)]](auto & res) {return res->equals(r_to_find);}`
(If I don't miss anything...)

Also I noticed that you dropped `equals` and now looks like you just compare pointers themselves...

src/util/tools.cc
```diff
@@ -447,11 +443,9 @@ void dict_decode_simple(const std::string& url, std::map<std::string, std::strin
 std::string mime_types_to_CSV(std::vector<std::string> mimeTypes)
 {
     std::ostringstream buf;
-    for (size_t i = 0; i < mimeTypes.size(); i++) {
-        if (i > 0)
-            buf << ",";
-        std::string mimeType = mimeTypes[i];
-        buf << "http-get:*:" << mimeType << ":*";
+    for (const auto& mimeType : mimeTypes) {
```
ytimenkov, line 446: Another hint is to modernize for fmt:
https://fmt.dev/latest/api.html#ranges-and-tuple-formatting

something like
`return fmt::format("{}", fmt::join(mimetypes, ","));`

Also consider accepting argument by const &

src/web/session_manager.cc
```diff
@@ -163,30 +163,33 @@ std::shared_ptr<Session> SessionManager::createSession(long timeout)
 
 std::shared_ptr<Session> SessionManager::getSession(const std::string& sessionID, bool doLock)
 {
+    if (sessions.empty()) {
+        return nullptr;
+    }
+
     std::unique_lock<decltype(mutex)> lock(mutex, std::defer_lock);
```
ytimenkov, line 170: decltype is not needed in C++17. just unique_lock, ctad should do the work.

src/util/tools.cc
ytimenkov: Or maybe not since it will just decorate the first item in the list...

neheb:
Passes now.

neheb:
rebased

whyman:
Travis CI :facepalm:

neheb:
Travis got fixed looks like.

src/cds_objects.cc
```diff
@@ -101,11 +101,8 @@ int CdsObject::resourcesEqual(const std::shared_ptr<CdsObject>& obj)
         return 0;
 
     // compare all resources
-    for (size_t i = 0; i < resources.size(); i++) {
-        if (!resources[i]->equals(obj->resources[i]))
-            return 0;
-    }
-    return 1;
+    return std::all_of(resources.begin(), resources.end(),
```
ytimenkov, line 104: Looks fishy: isn't that r always a one past last element in obj->resources?

Original loop indexed elements of both vectors, while with this change r is fixed, ie captured at lambda creation.

`std::equal` with custom predicate is better algorithm in this case.

src/autoscan.cc
```diff
@@ -174,30 +168,29 @@ int AutoscanList::removeByObjectID(int objectID)
...
-        }
+    auto it = std::find_if(list.begin(), list.end(), [=](const auto& item) { return objectID == item->getObjectID(); });
+    if (it == list.end()) {
+        return INVALID_SCAN_ID;
     }
-    return INVALID_SCAN_ID;
+
+    auto dir = *it;
+    dir->setScanID(INVALID_SCAN_ID);
+    list.erase(list.begin() + std::distance(list.begin(), it));
```
ytimenkov, line 178: This is a really strange math... Shouldn't be just `list.erase(it);`?
Also what is the type of `list`? because `erase` on `vector` invalidates iterators (in any case `it` should point at deleted element). If return value is used to preserve the original behavior you need to compute distance first and then return. If it's a `bool` in disguise just return `true`/`false`.

I would also probably just do `(*it)->setScanID(INVALID_SCAN_ID)` without creating a temporary.

src/iohandler/process_io_handler.cc
```diff
@@ -65,20 +65,11 @@ bool ProcListItem::abortOnDeath() const
 
 bool ProcessIOHandler::abort() const
 {
-    bool abort = false;
-
     if (procList.empty())
```
ytimenkov, line 68: This check is excessive. just `return std::any_of(...);`

neheb: Removed.

src/autoscan.cc
neheb: I think I adjusted it right.

src/cds_objects.cc
neheb: Correct. I took the code from your earlier comment.

ytimenkov: Oh, I found the one you training about ☺️

Looks like I didn't pay enough attention to the change add a whole and suggested only mechanical transformation of your change😞 Happens when reviewing on a phone (where screen is tiny and no access to all code...).

src/autoscan.cc
```diff
@@ -95,8 +95,7 @@ int AutoscanList::_add(const std::shared_ptr<AutoscanDirectory>& dir)
 {
     std::string loc = dir->getLocation();
```
ytimenkov, line 96:
```suggestion
    if (std::any_of(list.begin(), list.end(),
            [loc = dir->getLocation()](const auto& i) { return loc == i->getLocation(); })) {
        throw_std_runtime_error(fmt::format("Path {} is added twice", dir->getLocation().native()));
    }
```

src/autoscan.cc
```diff
@@ -95,8 +95,7 @@ int AutoscanList::_add(const std::shared_ptr<AutoscanDirectory>& dir)
 {
```
ytimenkov, line 95: Please make it `void`. its return value is not used anywhere.

src/autoscan.cc
```diff
@@ -174,30 +168,28 @@ int AutoscanList::removeByObjectID(int objectID)
...
```
ytimenkov, line 168: Whole function `removeObjectById` is not used. nuke it.

src/autoscan.cc
```diff
@@ -174,30 +168,28 @@ int AutoscanList::removeByObjectID(int objectID)
...
+        return INVALID_SCAN_ID;
     }
-    return INVALID_SCAN_ID;
+
+    (*it)->setScanID(INVALID_SCAN_ID);
+    list.erase(it);
+    return std::distance(list.begin(), it);
 }
 
 int AutoscanList::remove(const fs::path& location)
```
ytimenkov, line 181: Ditto. Not used anywhere.

src/util/tools.cc
```diff
@@ -229,13 +229,9 @@ fs::path find_in_path(const fs::path& exec)
 
     std::error_code ec;
     auto pathAr = split_string(PATH, ':');
-    for (auto& path : pathAr) {
-        fs::path check = fs::path(path) / exec;
-        if (isRegularFile(check, ec))
-            return check;
-    }
+    auto it = std::find_if(pathAr.begin(), pathAr.end(), [&](const auto& path) { fs::path check = fs::path(path) / exec; return isRegularFile(check, ec); });
```
ytimenkov, line 232: This is not an equivalent transformation: before the full path was returned, now only the base path.

Honestly I'm not convinced that converting this loop worth the effort.

src/web/session_manager.cc
```diff
@@ -163,29 +163,31 @@ std::shared_ptr<Session> SessionManager::createSession(long timeout)
...
     if (doLock)
         lock.lock();
-    for (const auto& s : sessions) {
-        if (s->getID() == sessionID)
-            return s;
-    }
-    return nullptr;
+
+    auto it = std::find_if(sessions.begin(), sessions.end(), [=](const auto& s) { return s->getID() == sessionID; });
```
ytimenkov, line 174:
```suggestion
    auto it = std::find_if(sessions.begin(), sessions.end(), [&](const auto& s) { return s->getID() == sessionID; });
```

src/autoscan.cc
neheb: It's used in add above.

src/autoscan.cc
neheb: Nuked.

src/autoscan.cc
neheb: Nuked.

src/web/session_manager.cc
neheb: Yikes good catch.

src/autoscan.cc
neheb: Fixed.

src/util/tools.cc
neheb: Good catch. Reverted.

src/cds_objects.cc
neheb: Hmmm I'm not sure about this change. Converting this to a range loop gives

```
    for (auto res : resources) {
        if (!res->equals(obj->resources[std::distance(resources.begin(), resources.end())]))
            return 0;
    }
```

That's quite verbose.

neheb: Meh reverted.

src/web/session_manager.cc
neheb: Removed.

ytimenkov:
Something like this (after checking for size):
```
return std::equal(resources.begin(), resources.end(), obj->resources.begin(),
    [](const auto& r1, const auto& r2) {return r1->equals(r2); });
```

src/autoscan.cc
```diff
@@ -93,11 +93,7 @@ int AutoscanList::add(const std::shared_ptr<AutoscanDirectory>& dir)
 
 int AutoscanList::_add(const std::shared_ptr<AutoscanDirectory>& dir)
 {
```
ytimenkov, line 95: Yes, you need convert that one too.

src/autoscan.h
```diff
@@ -77,8 +77,6 @@ class AutoscanList {
     /// \brief removes the AutoscanDirectory given by its scan ID
     void remove(size_t id);
 
-    int removeByObjectID(int objectID);
-
     /// \brief removes the AutoscanDirectory with the given location
     /// \param location the location to remove
     /// \return the scanID, that was removed; if nothing removed: INVALID_SCAN_ID
```
ytimenkov, line 82: `int remove(fs::path...)` should be gone from the header too.

neheb:
Curious. Why a lambda?

ytimenkov:
> Curious. Why a lambda?

Because `std::equal` calls `operator ==` and it's not what we want, we want to compare what pointers point to. And using special function for this 😐

Ps I'm experimenting with different github apps. Apparently vs code extension didn't allow replying on reverted change.

neheb:
Ah got it.

src/autoscan.h
neheb: Done.

src/autoscan.cc
neheb: Done.

neheb:
As you say that, some change broke travis :).

neheb:
add functions need to stay as int.

whyman:
LGTM, does make me think we need more tests

neheb:
Rebased and squashed that revert to int commit.

neheb:
A few more in the upnp client code. I also changed one since I felt uneasy about it.

neheb:
Rebased.

whyman:
I am quite conflicted on this PR, it seems better to be using these std:alogos, but looking at the PRs they are harder to understand what is going on, at least for me.

Thoughts from other members of the community?

KarlStraussberger:
I know similar structures from C# which looked unfamiliar in the beginning. Some still do!

It's like regular expression, some get as weird that you're not able to modify them after a week. 

As long everything works as before, I rely on the language mechanisms to be a bit more efficient than my own doing - as long as the code stays legible for me - which changes over time.

After four months with this project and rediscovering C++, I'm fine with the changes.

PR #888, opened 2020-05-24
Add friendly messages when finding spdlog as library

whorfin:
When find_package fails, the error spew makes it look like a build failure.
That's why I put find_library first in #718 
Since 72f7d5b8c281f9591e2a89a321f6aa06d57afc8d changed that, for systems without a .cmake file for spdlog, that spew looks like a problem.  This patch adds some friendly status messages to make the configure process understandable.

### DIFF
--- a/src/autoscan_inotify.cc
+++ b/src/autoscan_inotify.cc
@@ -50,7 +50,7 @@ AutoscanInotify::AutoscanInotify(std::shared_ptr<Storage> storage, std::shared_p
     std::error_code ec;
     if (isRegularFile(INOTIFY_MAX_USER_WATCHES_FILE, ec)) {
         try {
-            int max_watches = std::stoi(trim_string(readTextFile(INOTIFY_MAX_USER_WATCHES_FILE)));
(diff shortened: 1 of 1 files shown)


whyman:
@whorfin Thanks, much nicer! Can you rebase the branch, seems like we have some extra commits in there

ytimenkov:
May I ask ooc what are the systems which have spdlog without cmakefiles?

I also think that since spdlog requires a bit more than just a library (include dirs and extra defines) in my opinion they are just broken and should be fixed. There is a need for a proper way to communicate all this (cmake or package config or whatever, but not just a `find_library`).

ytimenkov:
May I ask ooc what are the systems which have spdlog without cmakefiles?

I also think that since spdlog requires a bit more than just a library (include dirs and extra defines) in my opinion they are just broken and should be fixed. There is a need for a proper way to communicate all this (cmake or package config or whatever, but not just a `find_library`).

whorfin:
Well there we are, git is all messed up --- the squash was problematic
I won't be able to pay any attention to this for another day, this wasn't time well spent

In answer to the other question about "why you need this",  #718 explains.  That's the way this package is distributed by SmartOS, and it works just fine.  There's no mandate for .cmake

PR #890, opened 2020-05-26
Add friendly messages when finding spdlog as library

whorfin:
let's try that again... git rebase farts fixed

When find_package fails, the error spew makes it look like a build failure.                                                                                                             That's why I put find_library first in #718                                                                                                                                             Since 72f7d5b changed that, for systems without a .cmake file for spdlog, that spew looks like a problem. This patch adds some friendly status messages to make the configure process understandable.


ytimenkov:
First I suggest change `find_package` to `find_package(spdlog CONFIG)` (Add `CONFIG`). We're quite certain that spdlog exports its CMake files which are read in config mode. So we don't need to rely on find modules.
In this case error message is far more comprehensible, just:
```
-- Could NOT find spdlog (missing: spdlog_DIR)
```

Then your following "found spdlog library" will make much sense.

ytimenkov:
Few other thoughts:

1.Googling for "joyent spdlog" gives Gerbera github as first 3 hits 🤣 
2. I'm mostly concerned about embeded vs external fmt and that these versions match. And you need to explicitly give spdlog's define if you use external fmt.

That said If you say that there is a pkgconfig for spdlog then it's probably a good idea to import it into CMake (via [`pkg_check_modules`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html
))  instead of plain `find_library`.

whorfin:
I went with the most general solution - "can you find the library" - rather than add another specialized requirement
Since I don't have cmake packages for spdlog, which is the whole reason we're here, I didn't touch that code as I can't test it and don't want to break it.

PR #891, opened 2020-05-30
What do you think: Adding Conan package manager

ytimenkov:
This is a fairly quick prototype on how things can look like and ask for feedback.

Major objective is to simplify development: Instead of spending day(s) on setting up a dev environment to contribute, it boils down to 2-3 simple commands to get a working build.

Some convincing slide is here :D https://github.com/CppCon/CppCon2018/blob/master/Presentations/applied_best_practices/applied_best_practices__jason_turner__cppcon_2018.pdf (slide 18.2 and on).

This is not complete and if you agree that this is a way to go we (me, I guess) need to port remaining libraries (like punpn and duktape, to begin with) to Conan.

I added only libraries which are already in Conan-Center repo and presented different techniques which can be used (useful?).

Of course it should be possible to build Gerbera "the old way", without Conan.


whyman:
Definitely interested.

Is there any way to avoid all the ifs for conan support and still have it work as expected?

If we can get all the libraries packaged into conan, that would be amazing

ytimenkov:
1. I chose "targets" approach because it is the most "modern": you explicitly specify that since target needs to consume a package and inherits all include paths, compiler definitions and libraries in one go.

Regardless of whether Conan is used or not I prefer it over obscure variables defined in find modules which you can get only by reading the file and hope that doc section doesn't contradict the actual code. Also error-prone if module changes and, say, starts requiring you to add compile definitions as well.

2. Conan can create `Find*` modules for packages it provides (has `cmake_paths` and `cmake_find_package` generators) which should provide drop-in replacement without any change to `CMakeFiles.txt`.

However this is a theory 🙂. This feature is fairly new and I haven't seen it used yet (the recipes for open source projects still use `conan_basic_setup` macro).

My another concern is that Find modules have extra logic or checks for certain features of a library. Or just used inconsistently. This may also make it more difficult to have `CMakeLists.txt` without `if` s... See also the first item.

3. The `cmake_paths` Conan generator allows also package creator to provide a package config file for CMake to be used via `find_package`. This way you can get very good integration. Unfortunately I don't know which packages started to use this feature.

ytimenkov:
That's said I would prefer to wrap existing libraries into imported targets with includes, defines and other stuff and use same name in the main CMakeLists.txt (so only one if at the top). And if libraries catch up and start providing package configuration files all that complexity goes away.

Besides, Conan folks realized that isolating dependencies via `CONAN_PKG::` is the opposite to what users need and added support for "canonical" package names so that imported target has same name as found via `find_package`. Unfortunately I checked gerberas dependencies and only zlib has updated recipe (provides `zlib::zlib` target).

But I can definitely experiment more with what Conan provides.

ytimenkov:
Closing this one.
I will open a new one when ready to merge.

PR #900, opened 2020-06-05
Add Hint to set property type of JSON field to string, bool or number (fixes #879)

### LINKED ISSUE #879: Browsing files error
I'm trying to access via GUI some directories named after dates and years but I can't as Gerbera returns this error:

`{"success":true,"containers":{"parent_id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c6174657374","select_it":0,"type":"filesystem","container":[{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303133","child_count":1,"title":2013},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303139","child_count":1,"title":2019},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f30313331","child_count":1,"title":0131},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f30333135","child_count":1,"title":0315},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f30323239","child_count":1,"title":0229},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f30353135","child_count":1,"title":0515},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303136","child_count":1,"title":2016},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303138","child_count":1,"title":2018},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303134","child_count":1,"title":2014},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303135","child_count":1,"title":2015},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f30343330","child_count":1,"title":0430},{"id":"2f7372762f6465762d6469736b2d62792d6c6162656c2d46726565636f6d2f626172657269612f6d70332f6c61746573742f32303137","child_count":1,"title":2017}]}}×`

1 digit directory name is ok, 2 or more digits directory name result in errors.

is it me? did I make some error compiling?

KarlStraussberger:
… instead of automatic detection of property type. Hint is set to name and type of property.

### DIFF
--- a/src/util/xml_to_json.cc
+++ b/src/util/xml_to_json.cc
@@ -117,7 +117,7 @@ std::string Xml2Json::getAsString(const char* str)
 std::string Xml2Json::getValue(const char* name, const char* text, const Hints& hints)
 {
     std::string str = text;
-    auto hintsType = hints.asType;
+    auto& hintsType = hints.asType;
 
     auto hint = std::find_if(hintsType.begin(), hintsType.end(), [=](const auto& entry) { return entry.first == name; });
 
@@ -144,7 +144,7 @@ std::string Xml2Json::getValue(const char* name, const char* text, const Hints& 
 
 bool Xml2Json::isArray(const pugi::xml_node& node, const Hints& hints, std::string* arrayName)
 {
-    auto hintsArray = hints.asArray;
+    auto& hintsArray = hints.asArray;
     auto hint = std::find_if(hintsArray.begin(), hintsArray.end(), [=](const auto& entry) { return entry.first == node; });
 
     if (hint == hintsArray.end()) {
--- a/src/util/xml_to_json.cc
+++ b/src/util/xml_to_json.cc
@@ -59,7 +59,7 @@ void Xml2Json::handleElement(std::ostringstream& buf, const pugi::xml_node& node
                 buf << ',';
             else
                 firstChild = false;
-            buf << getAsString(at.name()) << ':' << getValue(at.as_string());
+            buf << getAsString(at.name()) << ':' << getValue(at.name(), at.as_string(), hints);
         }
     }
 
@@ -91,15 +91,16 @@ void Xml2Json::handleElement(std::ostringstream& buf, const pugi::xml_node& node
             if (array) {
                 if (nodeName != child.name())
                     throw_std_runtime_error("if an element is of arrayType, all children have to have the same name");
-            } else
+            } else {
                 buf << getAsString(child.name()) << ':';
+            }
 
             if (childAttributeCount > 0 || childElementCount > 0 || isArray(child, hints, nullptr)) {
                 buf << '{';
                 handleElement(buf, child, hints);
                 buf << '}';
             } else {
-                buf << getValue(child.text().as_string());
+                buf << getValue(child.name(), child.text().as_string(), hints);
             }
         }
     }
@@ -113,10 +114,22 @@ std::string Xml2Json::getAsString(const char* str)
     return "\"" + escape(str, '\\', '"') + '"';
 }
 
-std::string Xml2Json::getValue(const char* text)
+std::string Xml2Json::getValue(const char* name, const char* text, const Hints* hints)
 {
-    // type information is gone, we need to guess...
+    std::string item = name;
     std::string str = text;
+
+    for (const auto& hint : hints->asType) {
+        if (hint.first == item) {
+            if (hint.second == "string") {
+                return getAsString(text);
+            } else if (hint.second == "bool") {
+                return str;
+            } else if (hint.second == "number") {
+                return str;
+            }
+        }
+    }
 
     // boolean
     if (str == "true" || str == "false")
--- a/src/util/xml_to_json.cc
+++ b/src/util/xml_to_json.cc
@@ -38,7 +38,7 @@
 
 #include "util/tools.h"
 
-std::string Xml2Json::getJson(const pugi::xml_node& root, const Hints* hints)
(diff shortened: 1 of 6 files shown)


src/util/xml_to_json.cc
```diff
@@ -113,11 +114,23 @@ std::string Xml2Json::getAsString(const char* str)
...
 }
 
-std::string Xml2Json::getValue(const char* text)
+std::string Xml2Json::getValue(const char* name, const char* text, const Hints* hints)
 {
-    // type information is gone, we need to guess...
+    std::string item = name;
     std::string str = text;
 
+    for (const auto& hint : hints->asType) {
```
ytimenkov, line 122: If hints can't be null and it's not resigned, maybe make it a reference?

Also nice opportunity to try c++17's structured bindings `const auto& [name, type] : hints->asType`
(You don't need to create a string `item`. sts::string can compare with char*).

whyman: Agreed on the reference. Indifferent on the bind

ytimenkov: I was sleepy yesterday... If this is a `std::map`, there is no need to iterate over it linearly. Use `find`.

ytimenkov: find_if is not what i meant. std::map has `find` method which is logarithmic. The search is by the key, right?

src/util/xml_to_json.cc
```diff
@@ -113,10 +114,22 @@ std::string Xml2Json::getAsString(const char* str)
     return "\"" + escape(str, '\\', '"') + '"';
 }
 
-std::string Xml2Json::getValue(const char* text)
+std::string Xml2Json::getValue(const char* name, const char* text, const Hints& hints)
 {
-    // type information is gone, we need to guess...
     std::string str = text;
+    auto hintsType = hints.asType;
```
ytimenkov, line 120: This makes a copy ☺️ shouldn't be needed here.

KarlStraussberger:
HInts is a shared_ptr handed in, but can be changed to reference in the first place so we need not ot worry about nullptr. This change breaks PR #882 for `xml_to_json.cc` but does the same

src/util/xml_to_json.cc
KarlStraussberger: Well I just copy what PR #882 did. Changed it to reference

src/util/xml_to_json.cc
KarlStraussberger: Well, as I said, I copied. I am releaning C++ after 20 years of coding C# 👍

ytimenkov:
I like the last version except that it doesn't work as (I, me) expected.
`std::map` got a `find` overload which takes something which is comparable to its key. `const char*` in our case. I tried it in the debugger and it creates a temporary string 🤦 
Digging into [cppreference.com](https://en.cppreference.com/w/cpp/container/map/find
)... It says something about `Compare::is_transparent`, but most importantly gives an example for `FatKey`.

Long story short, you need to explicitly add comparison function to `asType`:
`std::map<std::string, std::string, std::less<>> asType;`

Apparently (and cppreference seems to confirm it) `std::map` defines comparator `less<Key>` which can compare only to `Key`, while `less<>` just deduces types for `operator <`. So `find` can compare it's key directly to C-string.

I feel bad about so much noise for almost nothing, but I've learned something new :)

src/util/xml_to_json.cc
ytimenkov: Learn 20 then. Ranges (not linq, but close), coroutines (async, await, although only language feature, no support on standard library), concepts (close to C# generics in some way...). 😂

KarlStraussberger:
I always love to learn new things.

It should be sufficient to upgrade `name` to `std::string` (coming from `const char*`). because we have a `map<string,string>`.

ytimenkov:
Well... in C++ when there is a tradeoff between correctness and performance it's almost always the correctness which wins.

In this case when you've changed argument conversion `const char*` -> `std::string` still takes place (and thus a memory allocation which I wanted to avoid).

Thing is, those functions are called with data from PugiXML which returns C-strings. This was also my first thought: do we really need C-string here? Unfortunately, yes, if we don't want a malloc for a temporary.

KarlStraussberger:
I just looked through the code here and there's an abundance of conversions from `char*` to string already. So I prefer avoiding extra comparison code to one more malloc (we indeed reduced the automatic conversions to exactly one - less code less errors).

whyman:
@KarlStraussberger Thanks for your work on this, is it ready to go?

KarlStraussberger:
From my point of view, it is.

Issue #897, opened 2020-06-03
InternetRadio not available on upnp-client


franconianmetal:
Hi,
I added some Internet-Radio-urls which work fine within gerbera.
Browser opens the stream-player.
But trying to find on my upnp-client (i.e. vlc) there is no entry for these radio-urls to play via vlc (bubbleupnp behaves the same)
Other media is working fine

What can I do to make these urls to be visible in the upnp-clients and to start streaming

![Bildschirmfoto vom 2020-06-03 19-52-24](https://user-images.githubusercontent.com/18643293/83670805-02645100-a5d4-11ea-956f-f5c30e560e67.png)
![Bildschirmfoto vom 2020-06-03 19-52-41](https://user-images.githubusercontent.com/18643293/83670814-06906e80-a5d4-11ea-869c-92cc4b20b8c1.png)

whyman:
Can you paste a debug log?

whyman:
Also, please make sure you are using the latest 1.5 version!

franconianmetal:
thanks, solved the problem

PR #907, opened 2020-06-10
Metadata comparison does not work

KarlStraussberger:
Changing description after object creation was not possible anymore (another issue discover alongside #897). 
== operatior for map should do what we want:
`Checks if the contents of lhs and rhs are equal, that is, they have the same number of elements and each element in lhs compares equal with the element in rhs at the same position`

### DIFF
--- a/src/cds_objects.cc
+++ b/src/cds_objects.cc
@@ -81,7 +81,7 @@ int CdsObject::equals(const std::shared_ptr<CdsObject>& obj, bool exactly)
     if (!resourcesEqual(obj))
         return 0;
 
-    if (!std::equal(metadata.begin(), metadata.end(), obj->getMetadata().begin()))
+    if (metadata != obj->getMetadata())
         return 0;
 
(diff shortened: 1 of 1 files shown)


whyman:
Isnt that also what `std::equals` does?

KarlStraussberger:
Well, the ==operator seems to look into the values also. The effect is that I could not change description.

ytimenkov:
I suspect a deeper problem.

The thing is `getMetadata()` returns a map by value 🤦 which becomes a **temporary**. Then `begin()` takes iteraor on that temporary. Temporaries should be destroyed after the whole expression is evaluated (and it's so in my setup), but who knows...

If possible, would be interesting to take the original code and run under valgrind, see if it complains on accessing freed memory.

ytimenkov:
Another issue with `std::equals` is that no one checks for the matching sizes, so if one map is longer than another the results will be wrong.

PR #912, opened 2020-06-12
Fixed use of freed memory

ytimenkov:
What a bummer. 2 persons working on the same fix late night...

#911

### DIFF
--- a/src/web/directories.cc
+++ b/src/web/directories.cc
@@ -31,6 +31,7 @@
 
 #include "pages.h" // API
 
+#include <set>
 #include <utility>
 
 #include "storage/storage.h"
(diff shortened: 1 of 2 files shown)


ytimenkov:
That has some other improvement... One bug I found is that `directory_iterator` should take error_code otherwise it throws and exception (permission denied to read `/root`) and just shows nothing.

ytimenkov:
We need a 3rd version of a fix 🤣

whyman:
Darn! Thanks for your efforts though

ytimenkov:
Shall we close this one? (or apply on top? :D)

I think I may consider myself an algorithm nerd 😊 If something can be done more efficiently for free why not do it?..

ytimenkov:
Another thought... Sorting by binary id may be not desired. I made tuple to explicitly contain file name first (comparison is faster than lexicographically traverse long paths with same prefix). But most importantly or can be easily turned into case-insensitive one. Not sure how all binary id sort non-English unicode code points and characters...

ytimenkov:
Actually (besides copying a string just to take its length) map version encodes "native" path. Hope no one uses KOI8 nowadays 🙂

whyman:
Lets close this one, if you have any parts you want improve, open new a PR :+1:

PR #914, opened 2020-06-15
Store current item in localstorage and reload it on startup or refresh

### LINKED ISSUE #856: How to make the UI maintain the navigation path
Switching between Database and File System view always resets the current navigation path which is quite inconvenient. Is there a way to configure this behavior? It's the same on Win10 / Chrome and Linux / Firefox.

KarlStraussberger:
Implemented storage of current item based on localStorage in browser.

Fixes #856

### DIFF
--- a/gerbera-web/mock-api/auth/config.json
+++ b/gerbera-web/mock-api/auth/config.json
@@ -21,6 +21,6 @@
       "action": []
     },
     "friendlyName" : "Gerbera Media Server",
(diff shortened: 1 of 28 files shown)


whyman:
This is fantastic work and again something I have been meaning to do for ages.

Would it be nicer if the state was stored in the History API/URL instead?

KarlStraussberger:
I did what I know best.

PR #918, opened 2020-06-17
Fix autoscan when renaming a directory

KarlStraussberger:
When renaming a directory the old files where not deleted from the database anymore. Issue was introduced with fs::path. Line 205 was older but makes no sense.
Also cleaned up the log messages (path to string instead of c_str)


whyman:
The loggers want c_str(), if you pass a string, the loggers will end up calling c_str() on them anyway.

whyman:
Awesome, thanks!

PR #922, opened 2020-06-20
Remove SQLite ifdefs as its not optional now

whyman:


ytimenkov:
Awesome! I really wondered why the define is still there...

PR #923, opened 2020-06-20
Added support for Conan package manager.

ytimenkov:
* Cleaned up and modernized CMakeLists: Use targets for most of
  dependencies.
* Bumped CMake minimum required version to have proper support for
  properties on imported targets as well as some Find modules.
* Build on FreeBSD: removed logging of pthread_self. It is a pointer
  and fmt refuses to format it. Moreover it's hard to correlate it with
  something outside the program (like top or ps):
  gettid() should be used instead (or thr_self on FreeBSD).
* Fixed some clang warnings.


whyman:
Generally looks good.

It does look like we have lost user friendly `message(FATAL_ERROR "Taglib not found!")` messages though.

ytimenkov:
It is handled by `REQUIRED` keyword to `find_package`. I get this message:

```
CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find pupnp (missing: UPNP_UPNP_LIBRARY UPNP_INCLUDE_DIR)
  (Required is at least version "1.12.1")
  /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
  cmake/Findpupnp.cmake:33 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:306 (find_package)
```

Message is somewhat longer, but at least no logic duplicated in CMakeLists.txt.

Another perk is that CMake won't allow to link to a target containing `::` if there is no such target (in contrast to libraries where it just keeps going).

ytimenkov:
Another question is whether we want to switch docker to use one of Conan-provided images (https://github.com/conan-io/conan-docker-tools)?

Easier to test multiple options combinations (e.g. minimal and all-in), but they're based on Ubuntu.

For "release" images Alpine is probably ok, but I would expect multi-stage build anyways without dev tools installed into the final image.

whyman:
I think using the conan images for CI is better, less for us to maintain. I want to get rid of Travis as its been unreliable lately, so putting it into actions is a nice idea. I would prefer to keep the "release" image as Alpine.

Issue #887, opened 2020-05-24
UPNP.Truncate the whole directory structure.


Alezzzander:
Hello,Devs!
Could you make option in web gui or config file to change root folder to browse?
Currently clients must browse full path from / to rearch desired folder. Is it possible to change it after import?
I know,that there is a script in archwiki mediatomb page,but i dont know what should i do with it :)

KarlStraussberger:
@whyman: What do you think about a mapping of the generated virtual structure to new paths with some sort of pattern. The perfect place would be ContentManager::addContainerChain.

This would solve serveral issues/feature requests currently open and enable the user to simply define its own layout without further scripting.

KarlStraussberger:
fixed

michaelbagrov:
Hello devs. I'm trying to achieve the behaviour expected to be possible after this fix, but it seems that something is not working as expected. Or I'm doing something wrong.
My configuration:
`
    <autoscan use-inotify="yes">
        <directory location="/cloud/shared" mode="inotify" recursive="yes" hidden-files="no"/>
    </autoscan>

    <layout parent-path=“no”>
        <path from="cloud/shared“ to=“Content”/>
    </layout>
`
I'm expecting, that all the contents of the /cloud/shared directory will be under PC Directory/Content. Instead of that I'm getting everything in PC Directory/cloud/shared/

Please help to understand where the issue is.

KarlStraussberger:
Thank you for asking, but the mapping only works for virtual layouts which in some places also reflect the physical structure.

Everything under PC Directory does. It holds all files regardless of their types and cannot be manipulated. You may hide PC Directory and use virtual folders only, which typically are focused on media type and meta data.

michaelbagrov:
Thank You @KarlStraussberger for speedy reply. Now the behavior is crystal clear.
Wouldn't it be too much to ask for an advice a configuration sample to fulfill the following need:
To have a virtual folder, which would aggregate content of different physical directories, including paths to the files in these directories, but omitting pats to these directories.
For instance:
Physical directories:
/cloud/shared/photo/**/*
/mnt/media/video/**/*
Virtual folder with name "Contents" to have:
Contents/photo/**/*
Contents/video/**/*

KarlStraussberger:
The virtual layout we create has Photo and Video right below root like PC Directory.

[...]

Issue #825, opened 2020-04-22
Not respecting folder hierarchy


Neocazen:
Here's what's going on.
    If I have a folder named Animes, this folder does not appear in directories but what has inside it, regardless of the settings of autoscan
For example,
Directories/folder 1 folder 2... and not Directories/Animes/ folder 1 folder 2 ...

pamapa:
Is it related to this? https://github.com/gerbera/gerbera/pull/714

In that case add the subfolder of Animes...

Neocazen:
> Is it related to this? #714
> 
> In that case add the parent folder of Animes...

Understand. The way to choose folders now, in a way, changed then. Because with version 1.4 it worked as I said.
![Captura de tela de 2020-04-22 17-01-20](https://user-images.githubusercontent.com/15347187/80028723-f016d380-84bb-11ea-9e4c-221c280227a1.png)
![Captura de tela de 2020-04-22 17-01-36](https://user-images.githubusercontent.com/15347187/80028728-f0af6a00-84bb-11ea-97fb-e7443a7ce3c3.png)

Neocazen:
It's just weird. I have a Movies folder in the same place and if I choose it, the Movies folder is created in the Video directory (even with the name '.') and looks like in previous versions. The problem is having subfolders as you can see earlier.
![Captura de tela de 2020-04-24 07-11-56](https://user-images.githubusercontent.com/15347187/80201760-fa8bb700-85fa-11ea-8811-d22e5c2da7d9.png)
![Captura de tela de 2020-04-24 07-12-27](https://user-images.githubusercontent.com/15347187/80201764-fb244d80-85fa-11ea-8f31-32434e1f26b3.png)

PR #925, opened 2020-06-22
Map virtual path

KarlStraussberger:
After issues (#825, #887) being raised and further discussions inbetween, this patch returns the pre-1.5 behaviour by option and also allows mapping paths without scripting - like translation of path items. It can even be used in conjunction with scripts and simplify the import scripts.


src/content_manager.cc
```diff
@@ -984,8 +985,13 @@ int ContentManager::addContainerChain(const std::string& chain, const std::strin
     if (chain.empty())
         throw_std_runtime_error("addContainerChain() called with empty chain parameter");
 
-    log_debug("received chain: {} ({}) [{}]", chain.c_str(), lastClass.c_str(), dict_encode_simple(lastMetadata).c_str());
-    storage->addContainerChain(chain, lastClass, lastRefID, &containerID, &updateID, lastMetadata);
+    std::string newChain = chain;
+    for (const auto& pattern : config->getDictionaryOption(CFG_IMPORT_LAYOUT_MAPPING)) {
+        newChain = std::regex_replace(newChain, std::regex(pattern.first), pattern.second);
```
ytimenkov, line 990: I'm concerned about 2 things:
1. It's a regex. Docs don't mention it so it could become a surprise if used with square brackets or point. Regexes are easy to get wrong.
2. This affects also js layout. Probably not what user expects (however quite unlikely to both use js layout and have mapping...).

KarlStraussberger: 1. is  good point. Added info
2. Very likely if you have separate media source. Easier than adding all maps in your script

ytimenkov: > Easier than adding all maps in your script

That is just the same regex replace, no?

But as I said, it's likely won't be a problem in practice.

whyman:
I like this idea, but I worry about how it will interact with scripted layouts

KarlStraussberger:
As long as there is no mapping there is no effect. If you have scripts and mapping, mapping superceeds the script output. Anyway, it is same admin who writes the script and sets the mapping.

ytimenkov:
Another idea is to map only roots: I.e. `/data/vol1/Video` -> `Video`, `/data/vol2/Video` -> `Video`, or by default map to `/`. This way the first switch is not needed.
To avoid duplication the `virtualRoot` element may be added to autoscan directories (in XML config), and as a stretch configurable via UI.
It can also be propagate to JS engine via, say `virtualRoot` variable.

To me this looks consistent and not surprising. However much more work (even without tests).

KarlStraussberger:
> Another idea is to map only roots: I.e. `/data/vol1/Video` -> `Video`, `/data/vol2/Video` -> `Video`, or by default map to `/`. 

My `Images` source folder contains pictures and videos take by our digital cameras, so mapping roots is not feasible (I was thinking about this before). But this brings up the idea to implement a metadata base virtual path creator to poplulate entries like "Camera Model" without scripting.

whyman:
LGTM, can we add a note to the docs to say that the mapping happens before it gets passed to the JS?

ytimenkov:
It happens after, or? `addContainerChain` is called from js code.

whyman:
Oh yes, you are right. That could be confusing, but lets assume people are smart enough :grin:

KarlStraussberger:
I won't back your assumption but take the risk :-)

ytimenkov:
As a side note, I have an idea to decouple js processing from adding to db. Inspired by dod and somewhat reflection in c++.

I.e. js should not add object immediately in addCdsObject call but rather locally stash it and only after script finished pass it to db backend.  Main reason is to avoid lock / sync call to db thread. This way that stashed data can be passed to db for further processing, or even whole import palletized.

ytimenkov:
So xsd schema is not updated? That would be convenient for me 🤔

PR #926, opened 2020-06-22
 use {} for cd

neheb:
iconv_t can be a pointer or an int. Normally the former but sometimes
the latter.

Silences warnings about using nullptr.

### DIFF
--- a/src/util/string_converter.cc
+++ b/src/util/string_converter.cc
@@ -42,7 +42,7 @@ StringConverter::StringConverter(const std::string& from, const std::string& to)
     dirty = false;
 
     if (!cd) {
-        cd = 0;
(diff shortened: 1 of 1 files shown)


src/util/string_converter.cc
```diff
@@ -42,7 +42,7 @@ StringConverter::StringConverter(const std::string& from, const std::string& to)
     dirty = false;
 
     if (!cd) {
-        cd = 0;
+        cd = 0; // NOLINT: iconv_t can be an int instead of a pointer.
```
ytimenkov, line 45: Does `cd = {};` do the trick?

neheb: Looks like it.

PR #927, opened 2020-06-22
Some conan fixes and GitHub actions

ytimenkov:
Example: https://github.com/ytimenkov/gerbera/actions/runs/144089972

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,23 @@ jobs:
           pip install -U --no-cache-dir conan
           conan profile new default --detect
           conan profile update settings.compiler.libcxx=libstdc++11 default
+      - name: Cache conan dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.conan/data
+          key: ${{ runner.os }}-conan-${{ matrix.compiler.name }}-${{ hashFiles('conanfile.py') }}-${{ hashFiles('conan/*') }}
+          restore-keys: |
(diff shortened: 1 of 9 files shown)


ytimenkov:
1. Problem (?) with current setup of GitHub actions is that they're triggered also on forks.
2. Travis can be removed (if it's the plan).
3. Proposed GitHub actions don't use any docker image and run natively on VM. Thankfully everything is set up there. However build times are slightly longer compared to Travis because most of C++ packages need to be rebuilt (since Gerbera uses C++17 and default binary packages on Conan-Center are for C++14). I plan to experiment with cache.

whyman:
Looks sensible, I wouldnt mind keeping the alpine build for PRs too, even if its a minor duplication, checks the non Conan build is OK.

ytimenkov:
> Looks sensible, I wouldnt mind keeping the alpine build for PRs too, even if its a minor duplication, checks the non Conan build is OK.

Alpine build or docker? I mean like it was in Travis: commands ran inside prebuilt container or line it's now: do everything in `docker build`?

whyman:
Sorry I meant the docker(alpine) image build.

I dont really care about the VM vs docker for the conan builds. Whatever is easier. We only went with Docker on travis as their VM images are ancient.

goettl79:
Hi, this actually prevented the automatic v1.6.0 release. I'd remove the limit on the master branch or just take branches/tags that start with v.

Issue #920, opened 2020-06-19
Why all my videos are marked that they have subtitles?


Hunsu:
I'm using Gerbera with BubbleUPnP Server and the android app. I compiled Gerbera from sources. In the android app all my videos are showing that they have subtitles stored in the server but it is not right.
How can I debug and fix this issue?

whyman:
Hi @Hunsu , thanks for reporting!

Can you provide us with a debug log? Ideally both importing the files, and then trying to play one

Hunsu:
I started gerbera in debug mode and added a test file that does not have any subtitles.
Here are the logs [logs.LOG](https://github.com/gerbera/gerbera/files/4807591/logs.LOG)

whyman:
Thanks!

It looks like we add a subtitle element to all video content currently...
```c++
        if (startswith(mimeType, "video")) {
            renderCaptionInfo(url, parent);
        }
```

ytimenkov:
Do you feel it can be easily fixed properly? I found ".srt" in 5 different places, mostly related...

@whyman Do you know why some things (like poster) are processed during scan while some others (like subtitles and thumbnails) dynamically when serving content (in `UpnpXMLBuilder`)?

I have (or had) some thoughts, but would like to confirm.

whyman:
Ideally we would do everything at scan, rather that serving time.

Obviously we cannot do some things at import time (virtual resources, transcoding, client specifics)

Art processing should be moved to import time, same with subs.

Both art and subs have "Internal"  and "Embedded" types and want serving in a similar fashion, so some mechanism to handle "associated" resources is needed that could be used for both cases.

ytimenkov:
Right, IIRC major problem with populating data at scan time was that some resources may be added afterwards and thus will be just ignored.

whyman:
Yes, but assuming you run timed scans often enough (or inotify) then your content should get reindexed and included.

It does mean that we have to be smart and when we see art or a poster get created, associate it correctly with the existing resource, and not treat it as a resource itself.

ytimenkov:
Current code only looks for "satellite" resources during the main scan (i.e. get path and check for `folder.jpg`).

I don't have a good idea how to do the reverse mapping: how to find the "main" resource given random file changed, but not the "main" resource.

KarlStraussberger:
Currently, CaptionInfo is added in a loop with each resource!

[...]

PR #930, opened 2020-06-25
Object details on UI

KarlStraussberger:
While investigating on the subtitles issue (#920) I found it very interersting to see all the details (metadata, aux data and resources) on the UI. Therefore I implemented a small addition.
edit-item-details

Along the way I updated the screenshots for all element to show the current version number and the new buttons.


src/web/edit_load.cc
```diff
@@ -90,6 +92,65 @@ void web::edit_load::process()
...
+
+        int resIndex = 0;
+        for (auto& resItem : objItem->getResources()) {
+            auto resEntry = resources.append_child("resources");
+            resEntry.append_attribute("resname") = "----RESOURCE----";
+            resEntry.append_attribute("resvalue") = fmt::format("{}", resIndex).c_str();
+            resEntry.append_attribute("editable") = false;
+
+            resEntry = resources.append_child("resources");
+            resEntry.append_attribute("resname") = fmt::format("handlerType").c_str();
```
ytimenkov, line 129: Why do you format just a plain string?

src/web/edit_load.cc
```diff
@@ -90,6 +92,65 @@ void web::edit_load::process()
...
+            resEntry.append_attribute("resvalue") = fmt::format("{}", MetadataHandler::mapContentHandler2String(resItem->getHandlerType()).c_str()).c_str();
```
ytimenkov, line 130: Please make `mapContentHandler2String` return `const char*` which you can assign to attribute directly. No need to format and `c_str()`.

src/web/edit_load.cc
```diff
@@ -90,6 +92,65 @@ void web::edit_load::process()
...
+            resEntry.append_attribute("editable") = false;
+
+            for (auto& resPar : resItem->getParameters()) {
+                auto resEntry = resources.append_child("resources");
+                resEntry.append_attribute("resname") = fmt::format(".{}", resPar.first.c_str()).c_str();
```
ytimenkov, line 135: This is parameter name, not a resName, right?

src/web/edit_load.cc
```diff
@@ -90,6 +92,65 @@ void web::edit_load::process()
...
+                resEntry.append_attribute("resvalue") = resPar.second.c_str();
+                resEntry.append_attribute("editable") = false;
+            }
+            for (auto& resAtt : resItem->getAttributes()) {
+                auto resEntry = resources.append_child("resources");
+                resEntry.append_attribute("resname") = fmt::format(" {}", resAtt.first.c_str()).c_str();
```
ytimenkov, line 141: Ditto, but resAttr?

web/js/jquery.gerbera.editor.js
```diff
@@ -110,6 +110,47 @@
     modal.find('#editState').val('').prop('disabled', false);
     modal.find('#editProtocol').val('').prop('disabled', false);
     modal.find('#editSave').text('Save Item');
+
+    modal.find('#metadata').empty();
+    modal.find('#auxdata').empty();
+    modal.find('#resdata').empty();
```
ytimenkov, line 116: I'm not an expert in html or js but I hope modern web technologies can do better than such logic scattered through multiple methods and html tags in strings...
Something like templates and bindings xaml in c# had for ages.

src/web/edit_load.cc
KarlStraussberger: It's the name of the xml attribute.

src/web/edit_load.cc
KarlStraussberger: ditto

web/js/jquery.gerbera.editor.js
KarlStraussberger: That's the way it works with jquery which is rather lowlevel. If we had stuff like Angular running it would look different.

src/web/edit_load.cc
KarlStraussberger: changed

src/web/edit_load.cc
KarlStraussberger: leftover

src/web/edit_load.cc
ytimenkov: Yes, I mean it contains resource parameter name, not resource name.

src/metadata/metadata_handler.cc
```diff
@@ -205,3 +205,30 @@ std::string MetadataHandler::getMimeType()
...
 }
+
+const char* MetadataHandler::mapContentHandler2String(int ch)
+{
+    switch (ch) {
+    case CH_DEFAULT:
+        return "Default";
+    case CH_LIBEXIF:
+        return "LibExif";
+    case CH_ID3:
```
whyman, line 216: CH_ID3 == Taglib now

whyman:
Looks good!

src/web/edit_load.cc
ytimenkov: I see now... But still, why is everything is called "resource" and not separate attrs, params and options? (However they all are like synonims to me and distinction is quite internal :|)

KarlStraussberger: The goal is to display everything in a simple table on UI. The chosen way makes that quite easy. If we store everything in a complex object the js-code is blown up. There are three lists now: metadata, auxdata and resources which all work the same way.

whyman:
Ready for merge?

KarlStraussberger:
Yes

PR #932, opened 2020-06-26
Reflect recent config changes in xsd

KarlStraussberger:
Forgot to do so in #859 and #925

### DIFF
--- a/config/config.xsd
+++ b/config/config.xsd
@@ -43,6 +43,7 @@
         <xs:complexType>
             <xs:all>
                 <xs:element ref="server"/>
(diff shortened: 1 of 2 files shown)


ytimenkov:
@KarlStraussberger What did you use for schema editing? Xml became so rare, maybe only hardcore enterprise software still uses it...

KarlStraussberger:
Well I'm using Notepad++ on a windows machine (there are XML extension), save to a Samba path and compile on Linux.

I've been already around when XML had it's great times! And furthermore: I still remember what to do.

PR #935, opened 2020-07-03
FIxed comparison to wrong object.

ytimenkov:
When exploring how metadata is stored and used found that in a751fea028a025f50a18e658d455e4d49a1a0128 code was torn apart and essentially metadata was compared with aux data.
Also calling `std::equal` without range check is dangerous and just not needed.


whyman:
Nice spot, thanks!

Issue #759, opened 2020-03-26
Separate multiple values of imported ID3v2.4 compliant mp3 tags


ghlupe2:
According to http://id3.org/id3v2.4.0-frames ID3v2.4 uses the null character to separate multiple values, which are allowed for all text information frames (tags beginning with T like TEP1 and TCOM).

TIPL (involved people list) and TMCL (musician credits list) are special cases. Every odd value is a role (instrument, voice, function) and every even is a person or a comma delimited list of persons.

Currently Gerbera replaces the null character with a space character which makes it hard to generate a list of values (which frequently contain blanks) using import.js. 

Following elmodaddyb https://github.com/gerbera/gerbera/issues/401#issuecomment-450454867 I inserted into the addaudio function of import.js this code for printing all imported data:
```

 for (var metaProp in obj.meta) {
    print(metaProp + '--->' + obj.meta[metaProp]);
  }
 for (var auxProp in obj.aux) {
    print(auxProp + '--->' + obj.aux[auxProp]);
  }
 for (var resProp in obj.res) {
    print(resProp + '--->' + obj.res[resProp]);
  }

```
for the attached [sample mp3 file](https://github.com/gerbera/gerbera/files/4388850/Come.Together.sample._ID3v2.4.zip) resulting in:

```
JS: dc:title--->Come Together
JS: upnp:artist--->The Beatles
JS: upnp:album--->Abbey Road
JS: dc:date--->1969-09-26
JS: upnp:originalTrackNumber--->1
JS: upnp:artist@role[AlbumArtist]--->The Beatles
JS: TMCL--->electric guitar George Harrison electric guitar, electric piano and handclaps John Lennon electric bass guitar Paul McCartney drums (drum set) and maracas Ringo Starr background vocals and lead vocals John Lennon lead vocals John Lennon background vocals Paul McCartney
JS: TXXX:WORK--->WORK Come Together
JS: TXXX:Writer--->Writer John Lennon Paul McCartney
JS: TPUB--->Apple Records
JS: TXXX:MusicBrainz Album Type--->MusicBrainz Album Type album
JS: protocolInfo--->http-get:*:audio/mpeg:*
JS: size--->628930
JS: bitrate--->20935
JS: duration--->00:00:30.0
JS: sampleFrequency--->44100
JS: nrAudioChannels--->2

```
[...]

PR #941, opened 2020-07-08
Load multivalue tags if supported instead of first item

### LINKED ISSUE #758: Allow duplicate tags to be imported from flac audio files
According to https://www.xiph.org/vorbis/doc/v-comment.htm 
> Field names are not required to be unique (occur once) within a comment header. As an example, assume a track was recorded by three well know artists; the following is permissible, and encouraged:
>                ARTIST=Dizzy Gillespie 
>                ARTIST=Sonny Rollins 
>                ARTIST=Sonny Stitt

Other important fields include WRITER, COMPOSER, PERFORMER, etc.
Currently Gerbera imports the first occurence of a field's value only. 

Following elmodaddyb's hint in his [comment ](https://github.com/gerbera/gerbera/issues/401#issuecomment-450454867) on issue #401 I inserted into the addaudio function of import.js this code for printing all imported data:

```
for (var metaProp in obj.meta) {
    print(metaProp + '--->' + obj.meta[metaProp]);
  }
 for (var auxProp in obj.aux) {
    print(auxProp + '--->' + obj.aux[auxProp]);
  }
 for (var resProp in obj.res) {
    print(resProp + '--->' + obj.res[resProp]);
  }
```

for the attached [sample flac file](https://github.com/gerbera/gerbera/files/4388413/Come.Together.sample.zip) resulting in:
```

JS: dc:title--->Come Together
JS: upnp:artist--->The Beatles
JS: upnp:album--->Abbey Road
JS: dc:date--->1969-01-01
JS: upnp:originalTrackNumber--->1
JS: upnp:artist@role[AlbumArtist]--->The Beatles
JS: PERFORMER--->George Harrison (electric guitar)
JS: WORK--->Come Together
JS: WRITER--->John Lennon
JS: LABEL--->Apple Records
JS: RELEASETYPE--->album
JS: protocolInfo--->http-get:*:audio/x-flac:*
JS: size--->2851998
JS: bitrate--->95066
JS: duration--->00:00:30.0
JS: sampleFrequency--->44100
JS: nrAudioChannels--->2

```
PERFORMER however should be: **George Harrison (electric guitar); John Lennon (electric guitar, electric piano and handclaps); Paul McCartney (electric bass guitar); Ringo Starr (drums (drum set) and maracas); John Lennon (background vocals and lead vocals); John Lennon (lead vocals); Paul McCartney (background vocals)** and WRITER should be: **John Lennon; Paul McCartney**

I'd like to suggest a new content import option in config.xml, like

`<duplicate-tags allow-duplicates="no|yes" dupe-separator="[separator string]"/>`

in order to maintain compatibility with the current behaviour. As a result multiple values of one property should be written by using the defined separator string, e.g. `ARTIST---Dizzy Gillespie;Sonny Rollins;Sonny Stitt`

Audio file downloaded from [archive.org](https://archive.org/details/cd_abbey-road_the-beatles_1/disc1/01.+The+Beatles+-+Come+Together.flac), tagged with MusicBrainz Picard, using [this](https://musicbrainz.org/release/d6010be3-98f8-422c-a6c9-787e2e491e58) release.

This partially covers issue [#418](https://github.com/gerbera/gerbera/issues/418)

KarlStraussberger:
Use toString() method from TagLib to get all item instead of first.

Also contains a fix for url_encode which failed with multi-byte UTF-8 characters.

This should fix #758, #759

### DIFF
--- a/src/metadata/taglib_handler.cc
+++ b/src/metadata/taglib_handler.cc
@@ -124,28 +124,28 @@ void TagLibHandler::addField(metadata_fields_t field, const TagLib::File& file, 
         // https://mail.kde.org/pipermail/taglib-devel/2015-May/002729.html
         list = file.properties()["ALBUMARTIST"];
         if (!list.isEmpty())
-            val = list[0];
+            val = list.toString(", ");
         else
             return;
         break;
     case M_COMPOSER:
         list = file.properties()["COMPOSER"];
         if (!list.isEmpty())
-            val = list[0];
+            val = list.toString(", ");
         else
             return;
         break;
     case M_CONDUCTOR:
         list = file.properties()["CONDUCTOR"];
         if (!list.isEmpty())
-            val = list[0];
+            val = list.toString(", ");
         else
             return;
         break;
     case M_ORCHESTRA:
         list = file.properties()["ORCHESTRA"];
         if (!list.isEmpty())
-            val = list[0];
+            val = list.toString(", ");
         else
             return;
         break;
@@ -416,12 +416,16 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
             if (frameList.isEmpty())
                 continue;
 
-            const TagLib::ID3v2::Frame* frame = frameList.front();
-            const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
-
-            const TagLib::String frameContents = textFrame->toString();
-            std::string value(frameContents.toCString(true));
-            value = sc->convert(value);
+            std::string value;
+            for (const auto& frame : frameList) {
+                const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
+                if (textFrame == nullptr)
+                    continue;
+                const TagLib::String frameContents = textFrame->toString();
+                if (!value.empty())
+                    value += ", ";
+                value += sc->convert(frameContents.toCString(true));
+            }
             log_debug("Adding auxdata: {} with value {}", desiredFrame.c_str(), value.c_str());
             item->setAuxData(desiredFrame, value);
             continue;
@@ -437,6 +441,8 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
 
             for (const auto& frame : frameList) {
                 const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
+                if (textFrame == nullptr)
+                    continue;
                 const TagLib::String frameContents = textFrame->toString();
                 std::string value(frameContents.toCString(true));
 
@@ -557,7 +563,7 @@ void TagLibHandler::extractFLAC(TagLib::IOStream* roStream, const std::shared_pt
             if (property.isEmpty())
                 continue;
 
-            auto val = property[0];
+            auto val = property.toString(", ");
             std::string value(val.toCString(true));
             value = sc->convert(value);
             log_debug("Adding auxdata: {} with value {}", desiredTag.c_str(), value.c_str());
--- a/src/metadata/taglib_handler.cc
+++ b/src/metadata/taglib_handler.cc
@@ -408,10 +408,7 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
     for (const auto& desiredFrame : aux_tags_list) {
 
         if (desiredFrame.empty()) {
-            continue;
-        }
-
-        if (frameListMap.contains(desiredFrame.c_str())) {
+        } else if (frameListMap.contains(desiredFrame.c_str())) {
             const auto frameList = frameListMap[desiredFrame.c_str()];
             if (frameList.isEmpty())
                 continue;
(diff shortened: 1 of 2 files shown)


src/metadata/taglib_handler.cc
```diff
@@ -416,12 +416,16 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
 
-            const TagLib::ID3v2::Frame* frame = frameList.front();
-            const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
-
-            const TagLib::String frameContents = textFrame->toString();
-            std::string value(frameContents.toCString(true));
-            value = sc->convert(value);
+            std::string value;
+            for (const auto& frame : frameList) {
+                const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
```
ytimenkov, line 421: You need to check that `dynamic_cast` succeed. Even if works now who knows about future versions...

src/metadata/taglib_handler.cc
```diff
@@ -416,12 +416,16 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
+                const TagLib::String frameContents = textFrame->toString();
+                std::string fvalue(frameContents.toCString(true));
+                fvalue = sc->convert(fvalue);
+                if (!value.empty())
+                    value += ", ";
+                value += fmt::format("{}", fvalue);
```
ytimenkov, line 427: No need to format a `std::string`.

src/metadata/taglib_handler.cc
```diff
@@ -416,12 +416,16 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
-
-            const TagLib::String frameContents = textFrame->toString();
-            std::string value(frameContents.toCString(true));
-            value = sc->convert(value);
+            std::string value;
+            for (const auto& frame : frameList) {
+                const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
+                const TagLib::String frameContents = textFrame->toString();
+                std::string fvalue(frameContents.toCString(true));
+                fvalue = sc->convert(fvalue);
```
ytimenkov, line 424: I would just do nested call. No need to come up with names for temporaries... But it's a matter of taste since no function makes use of rvalues.

src/util/tools.cc
```diff
@@ -326,14 +326,29 @@ static const char* HEX_CHARS2 = "0123456789ABCDEF";
...
+            cplen = 1;
+
         if ((c >= '0' && c <= '9') || (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || c == '_' || c == '-') {
             buf << static_cast<char>(c);
         } else {
             int hi = c >> 4;
             int lo = c & 15;
-            buf << '%' << HEX_CHARS2[hi] << HEX_CHARS2[lo];
+            if (cplen > 1)
+                buf << str.substr(i, cplen);
```
ytimenkov, line 347: Hopefully small buffer optimization kicks in for 4-byte string. Otherwise it feels as an overkill to allocate a memory for each codepoint ☺️

src/util/tools.cc
```diff
@@ -326,14 +326,29 @@ static const char* HEX_CHARS2 = "0123456789ABCDEF";
 std::string url_escape(const std::string& str)
 {
```
whyman, line 327: Call me lazy but having this stuff in our code seems like a waste of effort. Is it not in one of our libs? curl maybe?

KarlStraussberger: I'm sure it is. But this function is also used to store auxdata in database. That's how I came across it: One of my test files had umlauts in the composer TCOM.

The only way to solve this would be to migrate auxdata to its own table like metadata.

ytimenkov: I agree, but curl is optional.

About auxdata: does it need to be stored in the db in first place? It's not searchable and the only use case I remember is for import script.

On a side note I'm currently exploring the idea to store resources in a separate table like metadata. Could simplify code significantly especially handlers' part. Also had rid of current 4 different ways to have fanart...

KarlStraussberger:
@whyman UI tests seem to fail randomly at different places. Looks like timeout in selenium. Any way to fix that?

ytimenkov:
Was looking into CDS specification ("ContentDirectory:4 — Standardized DCP (SDCP) – June 30, 2015") and it says that quite many metadata elements like composer or artist are multi-valued which means "that the property may occur more than once in an object.".

So although concatenating into list is kind of works it may not be the proper solution.
I wonder what will be the impact on the codebase to support multi-values...

KarlStraussberger:
There is one thing we need to consider: How many clients out there do support the CD:4-Spec?

On the other side Metadata and Auxdata are stored in maps, which means we need not worry about mulitple entries. If we change this, we need to separate property handling according to single or multi-valued. This may also require changes to the database and to the import-scripting interface.

I'd recommend this as a new feature request for 2021 or later.

ytimenkov:
>  There is one thing we need to consider: How many clients out there do support the CD:4-Spec?

I have absolutely no idea. Neither how it changed between the versions.

I'm quite certain it is not the case for my setup either. Only if I decide to group by genres and most of movies are of multiple ones.

My only intention with this comment is to "raise awareness". Changes to gerbera internals for this are quite broad and breaking: db schema, js scripts, web UI...

PR #952, opened 2020-07-25
npupnp changes

neheb:
Same as https://github.com/gerbera/gerbera/pull/818 , just rebased and cleaned up (clang-format, less hacky cmake, etc...).

I'm not too familiar with CMake. No idea if the changes are OK.

find_package doesn't work with libnpupnp. It seems a .cmake file is required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ if (CONAN_EXPORTED)
 endif()
 
 set(CMAKE_VERBOSE_MAKEFILE off CACHE BOOL "Show verbose build commands")
+set(WITH_NPUPNP         0 CACHE BOOL "Use npupnp instead of pupnp")
 set(WITH_MAGIC          1 CACHE BOOL "Use libmagic to identify file mime types")
 set(WITH_MYSQL          0 CACHE BOOL "Store media information in MySQL DB")
 set(WITH_CURL           1 CACHE BOOL "CURL required for online services")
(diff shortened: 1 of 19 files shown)


neheb:
Hrm. conan doesn't like the first commit.

neheb:
Unfortunately that means that either conan or libnpupnp needs to be fixed.

pupnp installs in /usr/include/upnp
npupnp installs in /usr/include/npupnp/upnp

I'm guessing for compatibility reasons.

CMakeLists.txt
```diff
@@ -303,23 +303,31 @@ add_definitions(${LFS_DEFINITIONS})
 add_compile_options(${LFS_COMPILE_OPTIONS})
 target_link_libraries(libgerbera ${LFS_LIBRARIES})
 
-find_package (pupnp "1.12.1" REQUIRED)
+pkg_check_modules (NPUPNP libnpupnp)
```
ytimenkov, line 306: I prefer an explicit switch between pupnp / npupnp, for example in case one has both installed.
Also it would simnplify selection / conditional compilation logic.
If they were are 100% compatible then no changes should be needed, right?

src/iohandler/mem_io_handler.cc
```diff
@@ -35,7 +35,9 @@
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
+#ifndef USING_NPUPNP
```
ytimenkov, line 38: Instead of wrapping with ifdef it's better to just remove unused header.
Maybe there are other similar places...

neheb: All of these headers seem to be unnecessary...

Removed.

medoc92:
The install to /usr/include/npupnp/upnp was meant to avoid a change for code which used to #include <upnp/upnp.h>

There is some code around which did this, for example libupnpp, which  I can change easily, but also gmrender-resurrect.

So there is a choice of what to break, and I'd like to adopt "best practise" here. Further advice requested :)

neheb:
Think about what that means. Using upnp/upnp.h requires adding /usr/include for pupnp and /usr/include/npupnp for npupnp as the include path to work properly.

That include path for pupnp is not what the pkgconfig file says to do.

The proper solution is to install to /usr/include/npupnp and include upnp.h.

As an example: I have to add this to the OpenWrt package just to get it to compile: https://github.com/openwrt/packages/blob/master/multimedia/gerbera/Makefile#L73

There are two packages in OpenWrt that use libupnp currently. Both include upnp.h.

medoc92:
Ok, I made the change, it will be in the next release.

neheb:
Rebased.

neheb:
Changed the npupnp cmake check to be more similar to the curl one.

neheb:
@medoc92 would it make sense to add a .cmake file to npupnp? It would allow find_package to work.

medoc92:
I am not familiar with cmake, and I'm not sure that I understand what you mean. Isn't pkg-config the thing for finding packages ?

Please explain, I'm all for making things easier, but I am at a loss here.

neheb:
Never mind actually. It works fine currently.

According to https://cmake.org/cmake/help/latest/command/find_package.html , find_package uses a cmake file. It's used to find libupnp but I can't find a libupnp cmake file anywhere. Meaning this is magic as far as I'm concerned.

Current pkgconfig file works fine.

medoc92:
Ok. The FindLibUpnp.cmake file in gerbera does use pkgconfig. There is a bit too much hidden magic in cmake for my liking...

neheb:
Oh I'm looking at the wrong place...

whyman:
I think we should merge this :+1: :tada: 

@neheb Can you add documentation updates and add a minimum version check to the nupnp check path (what is the minimum version?)

whyman:
@ytimenkov Can you double check this wont impact the conan stuff? Build is passing so it _should_ be fine :smile:

neheb:
Minimum version is unreleased :).

The issue is that with current released versions, the pkgconfig cflags are wrong.

Looks like I need to learn Sphinx...

ytimenkov:
> @ytimenkov Can you double check this wont impact the conan stuff? Build is passing so it should be fine 😄

This should be fine as long as WITH_NPUPNP is optional, so one can play with this stuff locally.

It also should work if you "flip the switch" after have generated workspace with Conan by running cmake manually.

My suggestion however is to wrap find logic into separate find module named `cmake/Findnpupnp.cmake`. The best example I can find is `cmake/FindPugixml.cmake`. Also uses pkg-config.

The goal is that code in the main `CMakeLists.txt` should be like:
```
if(WITH_NPUPNP)
    find_package (npupnp REQUIRED)
    target_link_libraries (libgerbera npupnp::npupnp)
    add_definitions(-DUSING_NPUPNP)
else()
...
```
That is no explicit call to pkg-config, you probably don't want to parse version, `REQUIRED` is handled by `FindPackagaHandleStandardArgs` macro, etc. This way you decouple those details from main file making it easier to transition to find modules provided by npupnp (if switched to CMake for example) or Conan.

There is also possibility to use npupnp with Conan, but it's a bit of a process. Can describe if you want.

whyman:
> Minimum version is unreleased :).
> The issue is that with current released versions, the pkgconfig cflags are wrong.

@medoc92 Is it possible to get a release cut with this issue sorted?

medoc92:
I released libnpupnp 0.4.11 with the include files changes. Actually it's a change in the installation location ($includedir/npupnp instead of $includedir/npupnp/upnp), not to the pkgconfig Cflags, as per the above discussion with @neheb 

The Cflags in the pkgconfig file are still:

    Cflags: -I${includedir}/npupnp

whyman:
@neheb Can you add a check for that version, or having he right paths?

whyman:
> The goal is that code in the main CMakeLists.txt should be like...

@ytimenkov  Agree that is the goal, I was kind of assuming we can tidy it up once its merged :shrug: 

@neheb Another thought: It would be nice to add a variant to the CI builds to catch obvious regressions

neheb:
Pushed the version change.

Those .cmake files are voodoo to me. Second the suggestion to fix it after.

No idea how to add a variant to CI.

neheb:
Got rid of loops in upnp_headers.cc. Less code.

neheb:
@medoc92 do those loop removals look sane?

medoc92:
@neheb it's not obvious to me where these loop removals are :) Can you point me to the commit ?

neheb:
https://github.com/gerbera/gerbera/commit/2ebccbb993dca41674de295f2d513abd568f971a was the original. push_back for loop was replaced with std::copy (which does the same internally).

The map insert for loop was replaced with just placing it in the constructor. At least I think that's what I did...

medoc92:
Sorry, maybe I'm dense, but I can't find your change. Can you point me to one commit with the diff you are referring to ?

neheb:
https://github.com/gerbera/gerbera/commit/690ddae5df81aabf19866aa3eb657ecfde5042c0#diff-2a671292fa129ad6164a1803fcafa084R120

medoc92:
@neheb These look fine to me. Maybe there was an advantage of the loop against the copy constructor in case one of the types changed, but this is no big deal, because it's probably not going to happen, and it would cause an easily detected build error anyway.

neheb:
OK. This patch is currently in production. No problems to report.

neheb:
Added doc information.

whyman:
The CMake will want a tidy up I think, but lets get this get this into the world.

Thanks to @medoc92 and @neheb for getting this sorted!

PR #955, opened 2020-07-26
fix upnp header include

neheb:
This header was introduced with 1.14. The check as is will match on
1.12 versions.

### DIFF
--- a/src/util/upnp_headers.cc
+++ b/src/util/upnp_headers.cc
@@ -25,7 +25,7 @@
 
 #include "upnp_headers.h" // API
 
-#if (UPNP_VERSION > 11201)
+#if (UPNP_VERSION > 11299)
 #include <UpnpExtraHeaders.h>
 #else
 #include <ExtraHeaders.h>


src/util/upnp_headers.cc
```diff
@@ -25,7 +25,7 @@
 
 #include "upnp_headers.h" // API
 
-#if (UPNP_VERSION > 11201)
+#if (UPNP_VERSION_MINOR > 12)
```
KarlStraussberger, line 28: In order to be really precise you  also need to add `UPNP_VERSION_MAJOR > 1 ||`
Anyway, we're discussing things that do not exist!

neheb: Not quite. My installation of libupnp has a version of 12.2 (some old git revision).

version major won't get bumped. Based on how fast Gerbera moves, I assume this whole block will go away.

KarlStraussberger: Interesting, officially PUPnP goes from 1.12.1 to 1.14.0. Best suggestion is `UPNP_VERSION > 11299` then.

whyman: Can we add the major version too? Just avoid any chance of weirdness

neheb: Went with Karl's suggestion. This define will probably be removed at some point.

PR #957, opened 2020-07-28
Color folders with children instead of badges saying true

whyman:
image

Also autoscan badges now have full text and icon:
image


KarlStraussberger:
Looks nice, maybe there are similar things to change?

whyman:
I was just using the UI and it bugged me :+1: 

I need to fix the tests still.

PR #959, opened 2020-07-31
Add subtitle as resource and update entries with resources attached, improve ContentHandler

KarlStraussberger:
Fixes: #920 

Currently I do not see a better way to update all resources (media file was not changed!) than to remove the directory completely before rescanning it.

Identification is done by adding a special Resource which is used to tag the resource. Any change on such a file triggers rescanning of the directory


Eelisland:
Hi Karl,
Compile stop with error:
`-- Build files have been written to: /tmp/SBo/package-gerbera/gerbera/build
Scanning dependencies of target libgerbera
[  3%] Building CXX object CMakeFiles/libgerbera.dir/src/autoscan.cc.o
[  3%] Building CXX object CMakeFiles/libgerbera.dir/src/action_request.cc.o
[  3%] Building CXX object CMakeFiles/libgerbera.dir/src/autoscan_inotify.cc.o
[  4%] Building CXX object CMakeFiles/libgerbera.dir/src/cds_objects.cc.o
[  5%] Building CXX object CMakeFiles/libgerbera.dir/src/cds_resource.cc.o
[  6%] Building CXX object CMakeFiles/libgerbera.dir/src/config/config_generator.cc.o
[  7%] Building CXX object CMakeFiles/libgerbera.dir/src/config/config_manager.cc.o
[  8%] Building CXX object CMakeFiles/libgerbera.dir/src/content_manager.cc.o
/tmp/SBo/package-gerbera/gerbera/src/config/config_manager.cc: In member function ‘void ConfigManager::load(const std::filesystem::__cxx11::path&, const std::filesystem::__cxx11::path&)’:
/tmp/SBo/package-gerbera/gerbera/src/config/config_manager.cc:475:9: error: ‘DEFAULT_CLIENTS_EN_VALUE’ was not declared in this scope; did you mean ‘DEFAULT_ACCOUNTS_EN_VALUE’?
  475 |         DEFAULT_CLIENTS_EN_VALUE);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
      |         DEFAULT_ACCOUNTS_EN_VALUE
[  9%] Building C object CMakeFiles/libgerbera.dir/src/contrib/md5.c.o
[ 10%] Building CXX object CMakeFiles/libgerbera.dir/src/device_description_handler.cc.o
make[2]: *** [CMakeFiles/libgerbera.dir/build.make:160: CMakeFiles/libgerbera.dir/src/config/config_manager.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:124: CMakeFiles/libgerbera.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
`

KarlStraussberger:
> Hi Karl,
> Compile stop with error:
> `-- Build files have been written to: /tmp/SBo/package-gerbera/gerbera/build Scanning dependencies of target libgerbera [ 3%] Building CXX object CMakeFiles/libgerbera.dir/src/autoscan.cc.o [ 3%] Building CXX object CMakeFiles/libgerbera.dir/src/action_request.cc.o [ 3%] Building CXX object CMakeFiles/libgerbera.dir/src/autoscan_inotify.cc.o [ 4%] Building CXX object CMakeFiles/libgerbera.dir/src/cds_objects.cc.o [ 5%] Building CXX object CMakeFiles/libgerbera.dir/src/cds_resource.cc.o [ 6%] Building CXX object CMakeFiles/libgerbera.dir/src/config/config_generator.cc.o [ 7%] Building CXX object CMakeFiles/libgerbera.dir/src/config/config_manager.cc.o [ 8%] Building CXX object CMakeFiles/libgerbera.dir/src/content_manager.cc.o /tmp/SBo/package-gerbera/gerbera/src/config/config_manager.cc: In member function ‘void ConfigManager::load(const std::filesystem::__cxx11::path&, const std::filesystem::__cxx11::path&)’: /tmp/SBo/package-gerbera/gerbera/src/config/config_manager.cc:475:9: error: ‘DEFAULT_CLIENTS_EN_VALUE’ was not declared in this scope; did you mean ‘DEFAULT_ACCOUNTS_EN_VALUE’? 475 | DEFAULT_CLIENTS_EN_VALUE); | ^~~~~~~~~~~~~~~~~~~~~~~~ | DEFAULT_ACCOUNTS_EN_VALUE [ 9%] Building C object CMakeFiles/libgerbera.dir/src/contrib/md5.c.o [ 10%] Building CXX object CMakeFiles/libgerbera.dir/src/device_description_handler.cc.o make[2]: *** [CMakeFiles/libgerbera.dir/build.make:160: CMakeFiles/libgerbera.dir/src/config/config_manager.cc.o] Error 1 make[2]: *** Waiting for unfinished jobs.... make[1]: *** [CMakeFiles/Makefile2:124: CMakeFiles/libgerbera.dir/all] Error 2 make: *** [Makefile:149: all] Error 2`
Try to `make clean` before. Maybe you should pull my branch again into a new path.

src/content_manager.cc
```diff
@@ -496,6 +496,19 @@ int ContentManager::_addFile(const fs::path& path, fs::path rootPath, bool recur
...
         addRecursive(path, hidden, task);
     }
+
+    if (obj->hasResource(CH_RESOURCE)) {
+        std::string newPath = path.parent_path().c_str();
+        int objectID = storage->findObjectIDByPath(newPath, false);
+        if (objectID != INVALID_OBJECT_ID) {
+            // as there is no proper way to force refresh of unchanged files, delete whole dir and rescan it
+            removeObject(objectID);
+            auto adir = getAutoscanDirectory(objectID);
```
ytimenkov, line 506: If object is removed (I assume from the DB) what does getAutoscanDirectory return? Path is gone or?

Another thing is unclear: what prevents recursion? I.e. if folder.jpg is scanned and then triggers removal of directory...

At least this is not crystal clear from the code.

ytimenkov:
I still vote for a simple `stat` before adding an element (just move few lines from file_handler into a common function).

Here is a quite significant change to already quite complicated scan logic with no tests...

src/content_manager.cc
KarlStraussberger: getAutoscan is a leftover, will remove.

Adding a new directory uses another entry point into content manager, as far as I remember.

ytimenkov:
TBC. I like your approach to identify some files as "companion resources" and when they change update "main" resources and I believe this is the way to go, however I would do it slightly different way.

As mentioned removing directory is probably not a good idea. Instead some "rescan" should be done only updating items if they changed. Remove and readd causes a lot of disruption (notifications are sent, object ids change).

Also if those resource handlers and types are added now it will be harder to change later...

src/content_manager.cc
ytimenkov: So where will it readd files from the (just removed) directory?

KarlStraussberger:
The point why normal rescan fails, is that the media files have not changed. In order to get the satellites attached to fully rescan cannot be avoided - I don't want to mess around with individual resources.

Only the media-file itself "knows" if the satellite is fitting.

ytimenkov:
As I suspected there are few things in the PR which significantly and unnecessarily degrade performance. For example scanning one directory with jpeg files, Elapsed (wall clock) time (h:mm:ss or m:ss):
 branch|1000 files|5000 files
--------|--|--
master|0:06.60|0:37.82
pr|0:14.31|3:51.56

Not only it few times slower it also has non-linear complexity...

ytimenkov:
Actually I wanted to run it with a custom `<resource>`... Scan didn't finish in 30 minutes.

src/content_manager.cc
whyman: Could you call `_addFile()` and it should refresh it?

src/metadata/metacontent_handler.cc
```diff
@@ -0,0 +1,194 @@
+/*MT*
+
+    MediaTomb - http://www.mediatomb.cc/
+
+    metacontent_handler.cc - this file is part of MediaTomb.
```
whyman, line 5: If this is a new file then it needs a gerbera header.

src/content_manager.cc
whyman: No, as you pointed out the file doesnt change.

I guess we need to enhance the storage layer to be able to update resources.

ytimenkov: Probably a directory should be a unit of scan. This way it would be easier to keep track of such compound objects: multiple sound and subtitle tracks (for different languages) as well as case-insensitive file names for folder.jpg (if really important).

This makes sense only for directories with small number of files (say, 100, maybe configurable).

ytimenkov:
I also think that configuration in xml is quite difficult to get. Another recent issue suggest this too.

Instead I would rather have more interaction with js. This way user may tune behavior based on paths, for example. E.g. If only certain folder must be exceptional. Or as with mp3 tags, which separators to use.

src/metadata/fanart_handler.cc
```diff
@@ -35,39 +35,55 @@
...
     for (const auto& name : names) {
-        auto found = folder / name;
-        std::error_code ec;
-        bool exists = isRegularFile(found, ec); // no error throwing, please
-        log_debug("{}: {}", name, exists ? "found" : "missing");
-        if (!exists)
-            continue;
-        return found;
+        auto fileName = tolower_string(expandName(name, item));
+        for (auto& p : fs::directory_iterator(folder))
```
neheb: Can be const looks like.

KarlStraussberger:
> 
> 
> As I suspected there are few things in the PR which significantly and unnecessarily degrade performance. For example scanning one directory with jpeg files, Elapsed (wall clock) time (h:mm:ss or m:ss):
> branch 	1000 files 	5000 files
> master 	0:06.60 	0:37.82
> pr 	0:14.31 	3:51.56
> 
> Not only it few times slower it also has non-linear complexity...

I found one timing issue in the nesting of the loops: ` fs::directory_iterator(folder)) `should be in the outer and `names `in the inner loop. The number of significant files tested is at least 15 per folder instead of 2. The only folders we might exclude are pure image containers.

KarlStraussberger:
I also tried to add a file while gerbera is down. It seems that the rescan at startup does not see this!

KarlStraussberger:
DId some research in gerbera history and found the culprit PR #812. The containerId was assumed to be identical to the objectId of the AutoscanDirectory: It only is for the main dir. Thus, scanning of subdirs never worked since then.

KarlStraussberger:
There remains a gap when a file was changed or added while Gerbera was down. This cannot be closed without storing further data in the database and slowing down the startup process.

If we merge this, we have significantly more detection than before but not 100%, yet!

whyman:
I like this overall, but removing and readding the directory seems like a suboptimal way of doing this. If you have large flat directories then its a huge overhead.

KarlStraussberger:
That's why I removed all the trigger patterns to avoid it in the first place. Actually I am thinking of setting the configuation by directory as part of the configuration UI #960

ytimenkov:
Why this broken code was merged?

KarlStraussberger:
What's broken?

ytimenkov:
See my performance measurements. Default configuration increases complexity from O(N) to O(N^2). Adding custom configuration makes it exponential (or factorial). I think this is unacceptable performance impact, not worth the feature.

KarlStraussberger:
That's why there's no default configuration anymore. There have been some unnecessary loops which I removed. I managed to import my whole library (30000mp3, 7000mpeg, 65000jpeg) after these changes. I fixed an issue that subdirs have not been scanned at all. That improved speed considerably.

ytimenkov:
The problem is not with a big collection. The problem is when a single directory contains a large number of files. This (in my opinion) is much more common for photo albums (depending on organization, or for big events, for example).

Let me explain what terrifies me:

For case-insensitivity you sacrificed O(1) checking for file presence using FS and turned into **full directory swipe** for **every item in this directory**. Even if you tried to remove "default" you still load directory into a map, again blowing memory footprint (and just discarding it right after, which made it even worse than before).

Not only this is done for scanning, but also for **serving**! Sorry, but for me it doesn't make sense to load whole directory into memory to just serve a single file.

Another one, with removal of default `.srt` files handing you basically broke any existing user (like me): when things just worked now I have to go and fix the config file, and then wipe my DB to populate with new items (or I don't know about wiping, but forcing any unexpecting user to fix config is an unpleasant surprise already). 👎 

The last one is easiest to fix: just make `names` const and compiler will tell you immediately. This is a basic good practice to not have global variables. I don't even understand why you decided to go with a global variable in first place. If you append values from the config plain initializer list will do the job.

KarlStraussberger:
First of all, I started with the existing code and only change names to a static member in the classes. The same with the access to the path when serving.

Now as you detailed you concern I easily provided an update 
- with an option for case-sensitive handling which return the original fs-check
- which block filesystem access when names are not set
- which fixes the slow down you got with custom configuration (config was added each time instead of once), it stays there because I intend to set it on a per directory base. (PR #963)

I'm sorry, but there was no such thing as a default Subtilte-Handling. It worked by accident, because the subtilte was added to every video. Sometimes we have to break things to add new features.

PR #970, opened 2020-09-01
fixed mime type retrieval for symlinks with libmagic

cdbrendel:
I noticed (on my systems, at least) that when gerbera is built with `libmagic`, symlinks to files are not followed during the MIME type identification process, even when `CFG_IMPORT_FOLLOW_SYMLINKS` is set. This causes symlinks to media files to be identified as `inode/symlink` rather than the MIME of the target. 
 (By contrast, symlinks to directories containing non-symlink files are successfully identified as expected.) 

I've added the `MAGIC_SYMLINK` flag for the `magic_open()` call which causes `libmagic` to follow symlinks when identifying the MIME type of a file. I'm not sure if passing the symlink config setting to `getMIMETypeFromFile()` is the most elegant fix (first time contributor to gerbera here), but I thought I'd offer my solution.

### DIFF
--- a/src/content_manager.cc
+++ b/src/content_manager.cc
@@ -1095,7 +1095,7 @@ std::shared_ptr<CdsObject> ContentManager::createObjectFromFile(const fs::path& 
                 if (ignore_unknown_extensions)
                     return nullptr; // item should be ignored
 #ifdef HAVE_MAGIC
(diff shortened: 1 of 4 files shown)


whyman:
Hi @cdbrendel, this patch looks sensible to me! Thanks for the contribution!

PR #973, opened 2020-09-06
clang-tidy fixes

neheb:
Currently errors on the nodiscard change.


whyman:
These look fine to me apart from the nodiscard, which just looks like spam with no real advantage

neheb:
Removed.

PR #977, opened 2020-09-10
clang-tidy: use auto with several static functions

neheb:
Found with modernize-use-trailing-return-type


whyman:
I like this for Lambdas, but on plain functions like this you just increase the amount work to figure out the return type when looking at the code.

Issue #960, opened 2020-08-02
web interface to edit config.xml


EricdeWaal:
Would it be very difficult to make a web interface with the possible settings and options that modify config.xml when saved?
I am thinking of something like Webmin, which I use to manage some servers, and my homebrew NAS (debian buster) on which I run Gerbera.
It would make life a lot easier for newbies, and could be self documented. That way the current elaborate documentation would not be necessary anymore, and it can be updated together with a release, so that the 2 always match.
I think it would be a great step forward to making a good product a lot more user friendly!

Eric

whyman:
I think most of the settings could have a web UI to configure, transcoding comes to mind.

With some options there is a chicken and egg issue, for example network config, as we have a web based UI.

I worry about giving the daemon the option to edit its own config file, seems like a security issue waiting to happen... but I suppose we allow to add random files via web-ui already.

EricdeWaal:
Hi Ian,

Well, considering that through widely used systems like Webmin, Plesk, DirectAdmin and the likes you can remotely manage an entire server I would not worry about that too much.

And one could have 2 web ui's, one for the media management as we already have, and a separate UI for the system.

If both come default with a proper  authentication already in place, it should be no risk IMHO.

⁣Regards,

Eric

Op 4 aug. 2020 11:22, om 11:22, Ian Whyman <notifications@github.com> schreef:

KarlStraussberger:
I agree with @whyman that we should not expose all of the settings. But there's plenty of to edit by the user. Maybe we can even allow settings per directory as @ytimenkov proposed.

EricdeWaal:
Maybe I did not make myself very clear: I did not mean any gui generated by the deamon, just a separate web page that just edits the config file, to make Gerbera more user friendly.
So that a user does not need to know all the options, but makes selections from dropdown lists.

Op 4 aug. 2020 13:48, om 13:48, Karlchen <notifications@github.com> schreef:

KarlStraussberger:
That's exactly what I am thinking of. Plus an extra to configure some folders with different values.

KarlStraussberger:
> I worry about giving the daemon the option to edit its own config file, seems like a security issue waiting to happen... but I suppose we allow to add random files via web-ui already.

[...]

PR #982, opened 2020-09-15
Web configuration editor

KarlStraussberger:
First chunk. Displays most of current config file.

Fixes: #960


KarlStraussberger:
Web Interface is controlled by gerbera-config.json. Values are taken from current config.

Next step is to add editing features to web ui.
Also I plan to update ConfigManager::load and move default values and some checking to configMap (as vector<ConfigOption>)

KarlStraussberger:
Refactoring ConfigManager is finished for the moment.
ConfigGenerator might be done some time in the future.

Now it's time for the editing feature in UI.

KarlStraussberger:
Most of the options can already be changed (scalar, arrays and dictionaries). No adding for multiple values, yet.

Remaing topics have to wait until the weekend is over.

Next step might be updating the database and make changed persistent.

KarlStraussberger:
The roundtrip to the database is working now (untested for MySQL)

The remainign part is adding and removing values for lists (like mappings).

config/sqlite3.sql
```diff
@@ -55,6 +55,11 @@ CREATE TABLE "mt_metadata" (
   "property_value" text NOT NULL,
   CONSTRAINT "mt_metadata_idfk1" FOREIGN KEY ("item_id") REFERENCES "mt_cds_object" ("id") ON DELETE CASCADE ON UPDATE CASCADE
 );
+CREATE TABLE "mt_config_value" (
```
whyman, line 58: Would be `grb_` nowadays surely ;)

KarlStraussberger: I thought about it already, wanted to keep the naming scheme. But, it's not a bad idea to flag our additions correctly. I'll push an update.

KarlStraussberger: Just experienced an issue that the server hang when a database command caused an exception. Reason behind: The sqliteThread was still running and the exception could not break out.

KarlStraussberger:
It's almost done! (I do not plan to make transcoding items addable)
Next step is to cleanup and add the option "CFG_DIRECTORY_IMPORT" (similar to AutoscanDirectory but overriding lower nesting level settings)

whyman:
I personally dont think we should allow adding autoscans to the config via this method.

You can add autoscan dirs via the UI already, so I thik it would be confusing.

KarlStraussberger:
> 
> 
> I personally dont think we should allow adding autoscans to the config via this method.
> 
> You can add autoscan dirs via the UI already, so I thik it would be confusing.

Autoscan is working now, the only thing I did not implement is transcoding (you can edit existing but not add or remove anything)

KarlStraussberger:
Next 
- Make directoriy tweaking effective in content manager (and where necessary).
- Add dialog to edit tweaks in filesystem view

KarlStraussberger:
Final Steps:
- Cleanup
- Documentation
- Tests for UI

KarlStraussberger:
Last Task:
- Tests for UI

KarlStraussberger:
@whyman I'm satisfied with it now. I'd leave improved user experience (like dropdown lists, range checking) to new PRs

whyman:
Looks like there are conflicts, if you are happy then feel free to merge it

neheb:
```
 error: ‘CFG_SERVER_STORAGE_MYSQL_ENABLED’ was not declared in this scope
         mysql_en = setOption(root, CFG_SERVER_STORAGE_MYSQL_ENABLED)->getBoolOption();
```

KarlStraussberger:
> 
> 
> ```
>  error: ‘CFG_SERVER_STORAGE_MYSQL_ENABLED’ was not declared in this scope
>          mysql_en = setOption(root, CFG_SERVER_STORAGE_MYSQL_ENABLED)->getBoolOption();
> ```

Obviously, I did not compile with MYSQL disabled, it's config.h, line 84. Just swap with line 85.

PR #987, opened 2020-09-25
clang stuff

neheb:


src/storage/sql_storage.cc
```diff
@@ -481,17 +481,14 @@ std::vector<std::shared_ptr<CdsObject>> SQLStorage::browse(const std::unique_ptr
     std::shared_ptr<SQLResult> res;
     std::unique_ptr<SQLRow> row;
 
-    bool haveObjectType = false;
-
-    if (!haveObjectType) {
+    {
```
whyman, line 484: Shouldn't this block be removed instead?

src/transcoding/transcoding.cc
```diff
@@ -67,7 +67,6 @@ TranscodingProfile::TranscodingProfile(transcoding_type_t tr_type, std::string n
     buffer_size = 0;
     chunk_size = 0;
     initial_fill_size = 0;
-    tr_type = TR_None;
```
whyman, line 70: Looks fine?

src/storage/sql_storage.cc
neheb: I tried that. There's some variable redefinition stuff going on. Easiest to just keep it.

src/transcoding/transcoding.cc
neheb: ¯\_(ツ)_/¯

src/storage/sql_storage.cc
whyman: That makes it more worrying!

whyman:
Other than my two comments the rest looks fine

src/transcoding/transcoding.cc
neheb: The issue pointed out by clang-analyzer is that it's a dead store. It's never read.

src/storage/sql_storage.cc
neheb: Removed. Fixed the variable redefinition later on by clearing it before use.

neheb:
Fine now?

PR #993, opened 2020-10-01
Use getaddrinfo instead of gethostbyname

paolostivanin:
According to man page, `gethostbyname` is obsolete and `getaddrinfo` should be used instead.

### DIFF
--- a/src/util/tools.cc
+++ b/src/util/tools.cc
@@ -1366,11 +1366,15 @@ std::string sockAddrGetNameInfo(const struct sockaddr* sa)
 /// \brief
 int find_local_port(unsigned short range_min, unsigned short range_max)
 {
-    int fd;
     int retry_count = 0;
     int port;
-    struct sockaddr_in server_addr;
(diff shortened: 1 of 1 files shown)


whyman:
Thanks for the PR.

TBH I was not even aware this code was in the project.

Now Ive seen it I think I am more likely to strip out the "sopcast" feature and related code then update it...

src/util/tools.cc
```diff
@@ -1366,11 +1366,15 @@ std::string sockAddrGetNameInfo(const struct sockaddr* sa)
...
-    int fd;
     int retry_count = 0;
     int port;
-    struct sockaddr_in server_addr;
-    struct hostent* server;
+    struct addrinfo* result;
+    struct addrinfo* rp;
+
+    struct addrinfo hints;
+    memset(&hints, 0, sizeof(hints));
```
neheb, line 1375: You can remove memset and replace with = {} above.

PR #997, opened 2020-10-04
FFMPEG-metadata: Ensure date content can be parsed

### LINKED ISSUE #994: Gerbera in Gentoo, error: terminate called after throwing an instance of 'std::invalid_argument'   what():  stoi Aborted
Hello everyone!

I have installed the gerbera-1.6.1 program in my Gentoo, but this version and version 1.5 don't work either.
I manually launched the program with the command: "gerbera --logfile /var/log/gerbera.log --config /etc/gerbera/config.xml --port 49152 -i 192.168.1.10" and he worked a little bit, but after a short while he stopped with the message: 'terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
Aborted'

I thank you for your help!

KarlStraussberger:
Fixes #994


neheb:
Could use auto for the lambda.

PR #1004, opened 2020-10-14
Fix resource_zero_ok

KarlStraussberger:
Has to be true if resources contains at least one item

`error: Exception caught: tried to create object without at least one resource file://home/karl/Source/gerbera/src/storage/sql_storage.cc line:952 function:createObjectFromRow`


KarlStraussberger:
@whyman You should include this in 1.6.2 because it breaks loading resources for SQlite

neheb:
ACK

whyman:
Argh! Okay

whyman:
We really need to work on getting _way_ more tests in

Issue #1006, opened 2020-10-14
Unable to query existing sqlite3 db on upgrade from 1.6.1


limburgher:
Oct 14 10:47:41 bamboo gerbera[805427]: Query:SELECT "object_type" FROM "mt_cds_object" WHERE "id"=0SELECT "f"."id","f"."ref_id","f"."parent_id","f"."object_type","f"."upnp_class","f"."dc_title","f">
Oct 14 10:47:41 bamboo gerbera[805427]: error: unrecognized token: "0SELECT" file:/home/gwyn/fedora/git/gerbera/gerbera-1.6.3/src/storage/sqlite3/sqlite3_storage.cc line:441 function:waitForTask

KarlStraussberger:
Did you test 1.6.2 or 1.6.3?

limburgher:
1.6.3

PR #1007, opened 2020-10-14
fix clearing stream

KarlStraussberger:
stream.clear() just clears the error flags but not the content
fixes: #1006


KarlStraussberger:
@whyman I think we're up for another release :-(

whyman:
😐😔

limburgher:
That patch fixes it, thank you!

PR #1013, opened 2020-10-21
use structured binding declarations in for loops

neheb:
Small readability improvement.


whyman (approved):

whyman:
This is better, it would be even better if we could give them meaningful names

neheb:
I didn't read enough to figure out what they're used for.

PR #1021, opened 2020-10-23
treewide: add missing newlines

neheb:
Found with find -exec nvim -c "wq" '{}' \;


whyman:
Would be nice if we can find a way to get the CI to check this?

Can clang-tidy enforce trailing newlines?

neheb:
I think it already does that: https://reviews.llvm.org/rL314033

git clang-format does not modify the whole file.

PR #1023, opened 2020-10-26
Move call to MetadataHandler::getMetaFieldName into CdsObject::get/setMetadata

KarlStraussberger:
Move call to `MetadataHandler::getMetaFieldName` into `CdsObject::get/setMetadata`.

Avoids propagation of internal structure of `mt_keys`.
Additionally move names for scripting variables into separate array.


whyman:
So much clearer! Nice!

KarlStraussberger:
Actually have been inspired by the changes in #1022 
Also separates scripting from internal handling.

PR #1026, opened 2020-10-28
fix compilation under 32-bit

neheb:
size_t is an unsigned int under 32-bit, causing a redeclaration error.

### DIFF
--- a/src/web/pages.h
+++ b/src/web/pages.h
@@ -198,11 +198,8 @@ protected:
     std::vector<ConfigValue> dbEntries;
     std::map<std::string, pugi::xml_node*> allItems;
     void createItem(pugi::xml_node& item, const std::string& name, config_option_t id, config_option_t aid);
-    static void setValue(pugi::xml_node& item, bool value);
-    static void setValue(pugi::xml_node& item, const std::string& value);
-    static void setValue(pugi::xml_node& item, int value);
-    static void setValue(pugi::xml_node& item, unsigned int value);
-    static void setValue(pugi::xml_node& item, size_t value);
+    template <typename T>
+    static void setValue(pugi::xml_node& item, const T& value);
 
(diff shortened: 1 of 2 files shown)


KarlStraussberger:
I worked out a way to use templates which is more generic

neheb:
Updated. Please review. I had to remove 
template <> as that does not work with a declaration.

src/web/pages.h
```diff
@@ -198,11 +198,10 @@ class configLoad : public WebRequestHandler {
     std::vector<ConfigValue> dbEntries;
     std::map<std::string, pugi::xml_node*> allItems;
     void createItem(pugi::xml_node& item, const std::string& name, config_option_t id, config_option_t aid);
-    static void setValue(pugi::xml_node& item, bool value);
+    template <typename T>
+    static void setValue(pugi::xml_node& item, const T value);
```
KarlStraussberger, line 202: You may even make it a rereference 
```
    template <typename T>
    static void setValue(pugi::xml_node& item, const T& value);
```

src/web/pages.h
```diff
@@ -198,11 +198,10 @@ class configLoad : public WebRequestHandler {
...
     static void setValue(pugi::xml_node& item, const std::string& value);
-    static void setValue(pugi::xml_node& item, int value);
-    static void setValue(pugi::xml_node& item, unsigned int value);
-    static void setValue(pugi::xml_node& item, size_t value);
+    static void setValue(pugi::xml_node& item, const fs::path value);
```
KarlStraussberger, line 204: You don't have to declare this here. It works as a template override with `template<>` before the definition as a posted

KarlStraussberger (changes requested):

src/web/pages.h
neheb: Done.

src/web/pages.h
neheb: Removed.

neheb:
I think this is what you were looking for.

PR #1030, opened 2020-10-30
clang-tidy

neheb:


src/util/tools.h
```diff
@@ -338,8 +338,7 @@ std::string sockAddrGetNameInfo(const struct sockaddr* sa);
 
 #ifdef SOPCAST
 /// \brief Finds a free port between range_min and range_max on localhost.
-int find_local_port(unsigned short range_min,
-    unsigned short range_max);
+int find_local_port(uint16_t range_min, uint16_t range_max);
```
KarlStraussberger, line 341: I'd like to avoid using exact sizes in code, so I recommend a
`typedef uint16_t network_port_t;`
e.g. in `tool.h` and using that instead. Makes our code a lot more readable.

KarlStraussberger (changes requested):

src/util/tools.h
neheb: Done.

KarlStraussberger: I thaught about all of those, because they are ports!

neheb:
Decided to use struct member type directly.

src/database/mysql/mysql_database.cc
```diff
@@ -149,7 +149,7 @@ void MySQLDatabase::init()
     std::string dbHost = config->getOption(CFG_SERVER_STORAGE_MYSQL_HOST);
     std::string dbName = config->getOption(CFG_SERVER_STORAGE_MYSQL_DATABASE);
     std::string dbUser = config->getOption(CFG_SERVER_STORAGE_MYSQL_USERNAME);
-    unsigned short dbPort = (unsigned short)config->getIntOption(CFG_SERVER_STORAGE_MYSQL_PORT);
+    auto dbPort = uint16_t(config->getIntOption(CFG_SERVER_STORAGE_MYSQL_PORT));
```
KarlStraussberger, line 152: change this as well

src/server.cc
```diff
@@ -115,7 +115,7 @@ void Server::run()
     if (!ip.empty() && iface.empty())
         throw_std_runtime_error("Could not find ip: " + ip);
 
-    unsigned short port = (unsigned short)(config->getIntOption(CFG_SERVER_PORT));
+    auto port = uint16_t(config->getIntOption(CFG_SERVER_PORT));
```
KarlStraussberger, line 118: change this as well, please

KarlStraussberger (changes requested):

KarlStraussberger:
That's great.

KarlStraussberger:
Please see #1031 as well

neheb:
Should be fine now.

PR #1032, opened 2020-10-31
size optimizations + clang-tidy

neheb:


src/upnp_xml.cc
```diff
@@ -326,14 +326,7 @@ std::unique_ptr<pugi::xml_document> UpnpXMLBuilder::renderDeviceDescription()
     {
         auto serviceList = device.append_child("serviceList");
 
-        struct ServiceInfo {
```
KarlStraussberger, line 334: Personally, I prefer structured and typed code to anonymous tuples. It may be fine here, but I consider moving all those initialization to some header and make it a visible interface. In this case having the struct might be better.
@whyman : How do you feel about this?

whyman: Agreed, lets keep the typed code as much as possible, it makes working with it much easier

KarlStraussberger: @neheb : Please move struct and property (as static member) to upnp_xml.h to make these things visible.

neheb: Meh removed.

PR #1038, opened 2020-11-11
Fix lastfm scrobbler build

lineinthesand:
The lastfm scrobbler enabled via cmake option `-DWITH_LASTFM=1` would not build due to mismatches in the function signatures. The scrobbler passed strings but the functions expect enum `metadata_fields_t`

I have yet to test whether the scrobbler actually works, but at least the build succeeds [(openSUSE 15.2 package here)](https://download.opensuse.org/repositories/home:/lineinthesand/openSUSE_Leap_15.2/).

### DIFF
--- a/src/onlineservice/lastfm_scrobbler.cc
+++ b/src/onlineservice/lastfm_scrobbler.cc
@@ -80,12 +80,12 @@ void LastFm::startedPlaying(std::shared_ptr<CdsItem> item)
     currentTrackId = item->getID();
 
     log_debug("Artist:\t{}",
-        item->getMetadata(MetadataHandler::getMetaFieldName(M_ARTIST)).c_str());
+        item->getMetadata(M_ARTIST).c_str());
     log_debug("Title:\t{}",
-        item->getMetadata(MetadataHandler::getMetaFieldName(M_TITLE)).c_str());
+        item->getMetadata(M_TITLE).c_str());
 
-    std::string artist = item->getMetadata(MetadataHandler::getMetaFieldName(M_ARTIST));
-    std::string title = item->getMetadata(MetadataHandler::getMetaFieldName(M_TITLE));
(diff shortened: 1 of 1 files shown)


whyman:
Awesome, thank you!

lineinthesand:
Wow, that was fast!
I tested scrobbling in the meantime – it works! I played a track via my Yamaha NP-S303 (I had to patch libupnp to remove that silly port restriction >= 49152 because the player only seems to recognize dlna servers on port 8200; libupnp did not yet respond to my pull request, though )o:) and it was indeed scrobbled. The "Now Playing" feature doesn't seem to be implemented, but that's OK with me.

neheb:
@medoc92 can you comment on this restriction? libnpupnp has it too.

medoc92:
I see no reason for this restriction, the code should just test for a 0 value, because this is what means "use the default 49152" in the API init call.

I'll change it in npupnp. 

However it's the Yamaha in the wrong here, there is no assigned port for UPnP (apart from the SSDP UDP 1900 port). As far as I can see, 8200 is just the HTTP port used by MiniDLNA by default. 

A UPnP device must  use what the remote device told it to use during discovery.

lineinthesand:
I made two mutually exclusive patches for libupnp, one that just test for 0 and one that makes it configurable whether it uses the old behavior of restricting to 49152 or to just test for 0.
I just learned about libnupnp today, but I couldn't find the source code.

Regarding Yamaha: I'm not 100% sure about it only using port 8200, but I was using minidlna previously and it immediately worked. When I tried mediatomb/gerbera it didn't work with 49152 (or 49200, which I also tried) but it worked immediately with 8200. I'm not sure whether I did something wrong, but I found no way to configure it on the hardware and it doesn't seem to work automatically.

neheb:
https://framagit.org/medoc92/npupnp

PR #1039, opened 2020-11-12
some changes

neheb:


neheb:
Hmm for the ffmpeg find_if change, I wonder if it's better to use all_of instead.

KarlStraussberger (changes requested):
There is no need to reformat 

-  src/database/mysql/mysql_create_sql.h
and
- src/database/sqlite3/sqlite3_create_sql.h 

the files are generated

neheb:
Removed.

whyman:
Isnt `e` glibc specific? Is it supported by e.g. musl

neheb:
Yes it is.

neheb:
https://git.musl-libc.org/cgit/musl/tree/src/stdio/fopen.c#n23

KarlStraussberger (approved):

KarlStraussberger:
Worst thing possible is that there is an environment which does not compile.

PR #1044, opened 2020-11-14
Build on FreeBSD (via sr.ht)

whyman:


neheb:
```
ld: error: undefined symbol: libiconv_open
```

Is it possible to replace iconv with normal C++?

whyman:
This makes no sense, it should have "native" iconv so I we dont need to link anything extra, or at least thats my understanding....

neheb:
https://github.com/openwrt/packages/blob/master/multimedia/gerbera/Makefile#L48

I ask mainly because OpenWrt under uClibc-ng does not really have iconv support. It's stubbed out. It's also annoying dealing with iconv in libc vs. external iconv.

If Gerbera deals with UTF-8 and UTF-16, iconv could be replaced with charconv.

diizzyy:
https://svnweb.freebsd.org/ports/head/Mk/Uses/iconv.mk?revision=450634&view=markup this is what the ports framework does about when iconv support is requested

whyman:
@diizzyy Thanks for the link, Ive tried a bunch of different cmake args that were mentioned but still get the same linking issue. What am I missing?

KarlStraussberger:
I found the following message in the logs.

`Manually-specified variables were not used by the project: ICONV_LIBRARY`

It seems you need to add it to your link command.

whyman:
Turns out the magic is `-DLIBICONV_PLUG`

PR #1052, opened 2020-11-28
constexpr conversions

neheb:

### DIFF
--- a/src/web/config_load.cc
+++ b/src/web/config_load.cc
@@ -296,12 +296,11 @@ void web::configLoad::process()
         pr++;
     }
 
-    std::vector<config_option_t> autoscanList = {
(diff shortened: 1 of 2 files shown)


src/web/config_load.cc
```diff
@@ -296,10 +296,10 @@ void web::configLoad::process()
         pr++;
     }
 
-    std::vector<config_option_t> autoscanList = {
+    constexpr config_option_t autoscanList[] = {
```
whyman, line 299: Could we use std::array here too?

neheb: Part of the problem is that some of the members are ifdef'd out. std::array requires a size.

neheb: Converted to std::array

whyman:
Thanks!

PR #1056, opened 2020-12-09
Stop generating our own IDs

whyman:
Use DB autoincrement as ID values instead, like sensible people

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -104,7 +104,7 @@ enum {
                                              << '=' << TQD("rf", "id") << " LEFT JOIN " << TQ(AUTOSCAN_TABLE) << ' '                                \
                                              << TQ("as") << " ON " << TQD("as", "obj_id") << '=' << TQD('f', "id") << ' '
 
-enum SearchCol {
+enum class SearchCol {
     id = 0,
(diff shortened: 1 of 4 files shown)


src/database/sql_database.cc
```diff
@@ -1219,9 +1218,10 @@ std::unique_ptr<Database::ChangedContainers> SQLDatabase::removeObjects(const st
     if (count <= 0)
         return nullptr;
```
whyman, line 1220: This was a fun bug discovered when I switched to `enum class`

KarlStraussberger:
I'm not quite sure, but `id` seems no auto increment in sqlite.

whyman:
> I'm not quite sure, but `id` seems no auto increment in sqlite.

Yes I thought I would need to do a table modification, but apparently sqlite has a `rowId` automatically for tables and

> If a table contains a column of type INTEGER PRIMARY KEY, then that column becomes an alias for the ROWID

and

> The usual algorithm is to give the newly created row a ROWID that is one larger than the largest ROWID in the table prior to the insert. 

See https://sqlite.org/autoinc.html

I've tested and it seemed to work :+1: , interesting thing that I didn't know though!

PR #1058, opened 2020-12-11
several static to anonymous namespace changes

neheb:
The latter is the recommended way to do this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,4 @@
 BasedOnStyle: Webkit
 DeriveLineEnding: false
+FixNamespaceComments: true
 UseCRLF: false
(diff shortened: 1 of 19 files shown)


whyman:
I like the improved clang-format stuff.

Static was "un-deprecated" in C++11, and I think its clearer as it was, personally. http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1012

Open to others thoughts, though. @KarlStraussberger?

KarlStraussberger:
I also prefer static to the namespaces. On the other hand, I've got a mental note to refactor some of the statics to proper methods, attributes etc. Eg. `IS_CDS_ITEM` might become a virtual method in `CdsObject` with an override in `CdsItem`.

I'm just waiting for paid work to cease :-)

whyman:
Yeah, there is a lot of things to be improved, but not sure this is going to make things _more clear_

PR #1073, opened 2020-12-22
Fix handling of LastModifiedTime of AutoscanDirectory

### LINKED ISSUE #992: Adding files does not update last_modified_time in autoscan
Created to follow up root cause of #985

KarlStraussberger:
I started out to fix #992 . Along the way I found out that the messages I got while scanning some of my test files originated for LATIN1 charset in some of the old image files. In order to test I found it helpful to know the exact amount of files in my database.

So I ended up with a rather big patch:
- Keep LastModifiedTime for each directory and store correct LMT when shutting down
- Add configuration of charset for metadata in LibraryOptions and DirectoryTweak
- Add database counters to config view (Status)
- Initialize SQLite DB without restore flag being set
- Cleanup xsd

I was hoping for full update capability but we still have to remove items and trigger a rescan.


whyman:
Nice work! Especially like the counts in the UI - I was thinking about this recently.

I guess it would be difficult to split this out into multiple commits? I'm just wondering if we have to come back to this if it will be hard to work out what happened.

Plus I'm having hard time digesting what parts are related :laughing: 

That said, I'm not going to stop you merging it as is.

src/content_manager.cc
```diff
@@ -1,1768 +1,1811 @@
-/*MT*
```
whyman, line 1: Why the permission change? We just end up with a giant diff... does it have changes too?

KarlStraussberger: Happened by accident. Sometimes my Samba share gets confused :-)

KarlStraussberger: Thanks for looking!

PR #1076, opened 2020-12-26
clang-tidy: add const to several globals

neheb:
Found with cppcoreguidelines-avoid-non-const-global-variables


KarlStraussberger:
Thanks

PR #1086, opened 2021-01-04
Happy New Year

whyman:


src/config/config.h
```diff
@@ -4,7 +4,7 @@
     
     config.h - this file is part of Gerbera.
     
-    Copyright (C) 2020-2020 Gerbera Contributors
+    Copyright (C) 2021-2021 Gerbera Contributors
```
KarlStraussberger, line 7: Upps: 2021-2021

src/config/client_config.cc
```diff
@@ -4,7 +4,7 @@
 
     client_config.cc - this file is part of Gerbera.
 
-    Copyright (C) 2020 Gerbera Contributors
+    Copyright (C) 2021 Gerbera Contributors
```
KarlStraussberger, line 7: Wouldn't it be better to have 2020-2021?

KarlStraussberger (commented):
Wouldn't it be better to have 2020-2021 instead of 2021 only?

src/config/client_config.cc
whyman: Yes, this was just me lazily find and replacing!

KarlStraussberger: I managed to break bitmap images in this way (replace in all files)

PR #1087, opened 2021-01-04
cleanup unused code

pamapa:
as seen in `server.cc` open is always called with empty range, so drop this. In the `transcoding path` range seems to be used via parameters, lets keep this...


whyman:
Thanks!

PR #1089, opened 2021-01-05
Simplify CdsObject Type detection

KarlStraussberger:
Introduce simply isType() methods.

Reduce usage of IS_CDS_ function to basic operations (load to server)


whyman:
The end result is cleaner, My concern is that we have to tell the CDSObject class about every one of its sub-classes, which feels wrong.

KarlStraussberger:
You're right. It's not picture perfect. I prefer the readability part.

Only if we need special handling a new method is required. I'd love to carve out specific code to content handler hooks before it starts to get nasty.

KarlStraussberger:
I just tried a way to get rid of calls to the new methods alltogether. At least the example looks better than the code before. It's a pity, C++ hass no dynamic typing like in C#, so `dynamic_pointer_cast` has to be used.

src/content_manager.cc
```diff
@@ -914,74 +915,107 @@ void ContentManager::updateObject(int objectID, const std::map<std::string, std:
...
-            auto resource = cloned_item->getResource(0);
-            resource->addAttribute(R_PROTOCOLINFO, renderProtocolInfo(cloned_item->getMimeType(), protocol));
-        } else if (!mimetype.empty()) {
-            cloned_item->setMimeType(mimetype);
-            auto resource = cloned_item->getResource(0);
-            std::vector<std::string> parts = splitString(resource->getAttribute(R_PROTOCOLINFO), ':');
-            protocol = parts[0];
-            resource->addAttribute(R_PROTOCOLINFO, renderProtocolInfo(mimetype, protocol));
-        }
+    log_error("updateCdsObject: CdsObject {} not updated", title);
```
whyman, line 918: Perhaps this should throw an exception?

KarlStraussberger: it didn't before and I'm not sure whether the web page handler proberly deals with exceptions. The error message already is an improvement, if at any time in the future new classes are not handled.

PR #1091, opened 2021-01-05
improve upnp defines

pamapa:


src/upnp_common.h
```diff
@@ -0,0 +1,149 @@
+/*MT*
+    
+    MediaTomb - http://www.mediatomb.cc/
+    
+    upnp_common.h - this file is part of MediaTomb.
```
whyman, line 5: This header should be a Gerbera one as its a new file

whyman:
I like this idea, its clearer, but would go further and split out the DLNA stuff into its own header too.

Also, Lets rename `UPNP_DEFAULT` to just `UPNP`

src/upnp_common.h
pamapa: done

PR #1093, opened 2021-01-05
improve duration and bitrate calculations

pamapa:
- avoid using deprecated taglib audioProps->length()
- unify milliseconds to upnp duration string conversion
- remove cap to 999 hours, upnp spec does not enforce this, other services are also not doing this (this was only done in one of our paths)
- rename HMSToSeconds to HMSFToMilliseconds, so it matches millisecondsToHMSF


whyman:
Thanks!

PR #1094, opened 2021-01-05
consolidate mime type base functions in new mime class

pamapa:
- behind libmagic is a small database, avoid reopening it all the time
- consolidate mime type function

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,6 +177,8 @@ add_library(libgerbera STATIC
         src/util/generic_task.h
         src/util/jpeg_resolution.cc
         src/util/logger.h
+        src/util/mime.cc
+        src/util/mime.h
         src/util/mt_inotify.cc
         src/util/mt_inotify.h
         src/util/process.cc
--- a/src/content_manager.cc
+++ b/src/content_manager.cc
@@ -47,6 +47,7 @@
 #include "layout/fallback_layout.h"
 #include "metadata/metadata_handler.h"
 #include "update_manager.h"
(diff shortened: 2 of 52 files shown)


whyman:
I like the renames for sure, much nicer and the handling of the magic cookie is a good idea.

Im not so sure on passing the ContentManager as a shared pointer to every handler, mostly because the handlers dont "own" the contentManager. I am aware we abuse the shared pointers a lot because of history, but lets try to be clearer with it now.

Would it be sensible to create a new `Mime` class that can manage the libmagic lifecycle and provide these methods? Then we can pass that into anyone who needs mime-stuff?

pamapa:
good idea i will adapt the code to do so.

pamapa:
- the metadata handler only get the Mime object
- all classes which inherit from RequestHandler now only get content object (they needed it be for too), the others (config, database, mime) are then taken via content. This highly simplifies a lot of constructors. The alternate would be to pass along the Mime object through all those constructors too, which sucks.

pamapa:
ready to be merged

KarlStraussberger:
That really looks great. Object orientation moves in!

Maybe we can promote `Server` to the mentioned context class.

whyman:
Really nice work.

I would be keen for the new context class, and to keep the number of things in it as small as possible to reduce coupling.

Also wondering about using [observer-ptr](https://github.com/martinmoene/observer-ptr-lite) to make the ownership explicit.

pamapa:
ok i will provide a patch for the new context class later, i think this can be merged independently....

PR #1096, opened 2021-01-06
fix compilation with npupnp

neheb:
The improve upnp defines commit left out the npupnp codepath.


pamapa:
@neheb thanks

PR #1097, opened 2021-01-06
use serveContent for transcoding too

pamapa:


src/request_handler.cc
```diff
@@ -65,3 +66,27 @@ void RequestHandler::splitUrl(const char* url, char separator, std::string& path
         path = url_s.substr(0, i1);
     }
 }
+
+std::map<std::string, std::string> RequestHandler::getParameters(const char* filename, const char* baseLink)
```
whyman, line 70: parseParameters?

I always think of getters as simple accessors.

pamapa: good idea

pamapa:
ready to merge or i am missing something?

Issue #1098, opened 2021-01-06
Segfault on git head


whyman:
When trying to use the web ui.

I imagine @pamapa will sort it with the context changes, but issue raised just in case

```Thread 8 "gerbera" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe7fff640 (LWP 163502)]
0x0000555555719d32 in std::__shared_ptr<web::SessionManager, (__gnu_cxx::_Lock_policy)2>::__shared_ptr (this=0x7fffb0002a78) at /usr/include/c++/10.2.0/bits/shared_ptr_base.h:1181
1181          __shared_ptr(const __shared_ptr&) noexcept = default;
(gdb) bt full
#0  0x0000555555719d32 in std::__shared_ptr<web::SessionManager, (__gnu_cxx::_Lock_policy)2>::__shared_ptr (this=0x7fffb0002a78) at /usr/include/c++/10.2.0/bits/shared_ptr_base.h:1181
No locals.
#1  0x0000555555719d7d in std::shared_ptr<web::SessionManager>::shared_ptr (this=0x7fffb0002a78) at /usr/include/c++/10.2.0/bits/shared_ptr.h:149
No locals.
#2  0x0000555555780cab in ContentManager::getSessionManager (this=0x0) at ../src/content_manager.h:329
No locals.
#3  0x000055555577e713 in web::WebRequestHandler::WebRequestHandler (this=0x7fffb0002a30, content=std::shared_ptr<ContentManager> (empty) = {...}) at ../src/web/web_request_handler.cc:50
No locals.
#4  0x0000555555822a8b in web::auth::auth (this=0x7fffb0002a30, content=std::shared_ptr<ContentManager> (empty) = {...}) at ../src/web/auth.cc:69
No locals.
#5  0x00005555557782f9 in std::make_unique<web::auth, std::shared_ptr<ContentManager> const&> () at /usr/include/c++/10.2.0/bits/unique_ptr.h:962
No locals.
#6  0x0000555555777174 in web::createWebRequestHandler (content=std::shared_ptr<ContentManager> (use count 2, weak count 1) = {...}, page="auth") at ../src/web/pages.cc:47
        __func__ = "createWebRequestHandler"
[...]

PR #1101, opened 2021-01-07
Fix refactoring leftover. Breaks browsing files.

KarlStraussberger:
Similar to #1098


whyman:
We need more tests, this is silly stuff

PR #1104, opened 2021-01-09
Set track number if needed and support some more meta data on FFmpeg handler.

mlaerd:
I found that DSF files are not sorted by track number on UPNP_CLASS_CONTAINER_MUSIC_ALBUM.
So I propose to update FFmpeg handler to set it if needed. 
I ever see 2 formats of track number on commercial DSF files. One is "[track number]", e.g, "1", and the other is "[track number]/[total tracks]", e.g. "1/10". I confirmed this patch can set track number correctly for both of the 2 formats.

Furthermore, I want to import some more meta data types on FFmpeg handler refer to [FFmpeg Metadata#mp3](https://wiki.multimedia.cx/index.php/FFmpeg_Metadata).


whyman:
Looks good!, Thanks!

PR #1105, opened 2021-01-09
introduce context

pamapa:


KarlStraussberger:
Looks great.

I recommend moving creation (from `Server`) and ownership of `content` to `Context` also. Then, you have to pass `context` only and can do a `context->getContent()` instead of e.g. `content->getContext()->getConfig()`.

pamapa:
My intention was to not create cyclic dependencies:
- ContentManager needs context, but should also be in the context....

Server is not worth to move into context. std::shared_ptr<Server> is current only needed by update_manager.

Lets do this step by step. Please merge as is.

In a next step i plan to move more stuff into ContentManager, as it solely used inside there:
- UpdateManager
- TaskProcessor
- scripting Runtime
- LastFm

PR #1111, opened 2021-01-10
fix ui crash

pamapa:


pamapa:
looks like i accidentally dropped your fix, this merge request fixes this again

Issue #1014, opened 2020-10-22
samsung subtitles problem + hack to address it


druban:
Hi!

Samsung subtitles appear to be broken. There are couple of issues here:
CaptionInfo.sec header name contains ':' (needs to be removed)
URL provided in the header above is incorrect. Currently it refers to actual Unix path of a subtitle file which needs to be replaced w/ internal gerbera path.

I have no idea how to properly address URL bit, so I just added a temporary hack:

```
diff --git a/src/util/upnp_quirks.cc b/src/util/upnp_quirks.cc
index b25efcf2..1a9b4a2e 100644
--- a/src/util/upnp_quirks.cc
+++ b/src/util/upnp_quirks.cc
@@ -65,6 +65,8 @@ void Quirks::addCaptionInfo(const std::shared_ptr<CdsItem>& item, std::unique_pt
     if (validext.length() == 0)
         return;
 
-    std::string url = "http://" + Server::getIP() + ":" + Server::getPort() + pathNoExt + validext;
-    headers->addHeader("CaptionInfo.sec:", url);
+    // std::string url = "http://" + Server::getIP() + ":" + Server::getPort() + pathNoExt + validext;
+    std::string url = "http://" + Server::getIP() + ":" + Server::getPort() +
+        "/content/media/object_id/" + std::to_string(item->getID()) + "/res_id/0/ext/file" + validext;
+    headers->addHeader("CaptionInfo.sec", url);
 }
```

While it works, it's definitely not the right way to compose URL. Somebody who understands code better than I do should fix it properly

Cheers!

whyman:
Can you tell us the "before" and "after" values of this change?

druban:
Not sure I understand the question. Before this change subtitles aren’t displayed. After this change they work fine ...

whyman:
I  was talking about the URL values, for some reason the first time I read the report I didnt really get what you were saying. Its quite clear in retrospect.

pamapa:
The fix looks reasonable to me, i will provide a merge request out of your changes...

PR #1115, opened 2021-01-10
fix: #1014 samsung subtitles problem, patch done by druban

pamapa:


whyman:
LGTM as its broken, but we need to find a way to get rid of those direct `Server` calls

pamapa:
+1

PR #1118, opened 2021-01-11
pass "*" to UpnpInit2 instead of nullptr

neheb:
This binds to all interfaces instead of just one.

ping @medoc92 I haven't actually tested this. I remember an issue with docker where libnpupnp did not work properly. It only bound to an IPv6 interface or something.

Nevertheless I do believe this should work properly based on my reading of the npupnp code.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -104,7 +104,11 @@ void Server::run()
     auto port = in_port_t(config->getIntOption(CFG_SERVER_PORT));
 
     log_info("Initialising libupnp with interface: '{}', port: {}", iface.c_str(), port);
+#if defined(USING_NPUPNP)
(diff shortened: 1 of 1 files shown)


neheb:
that build error looks bogus.

pamapa:
Thanks

medoc92:
I believe that using multiple (all) interfaces currently works fine for a device. 

This is new functionality (compared to pupnp), it's quite a bit more complicated than it appears, and it has seen a lot less testing than the parts which where just carried over, so feedback is particularly welcome.

medoc92:
Also this seems like rather a big change, with possibilities for user astonishment, and it does not actually bring new function (the same result could be obtained by specifying * as interface name).

PR #1119, opened 2021-01-11
fix UserAgent lookup, different between UpnP services and file request

pamapa:
Refactor User-Agent lookup.

Some client do not report the same User-Agent for UPnP services and file request, therefore we use what is cached previously via UpnpActionRequest to keep the code simple.

The main idea here is that before FileRequest (image/video file access) is called, an UPnP service (e.g. doBrowse) is called. From the first call we get the client info and cache it, afterwards we can reuse that info.

VLC example User-Agent info:
- ActionRequest: `Linux/5.9.0-5-amd64, UPnP/1.0, Portable SDK for UPnP devices/1.8.4`
- FileRequest: `VLC/3.0.11.1 LibVLC/3.0.11.1`

Same happens with Samsung TVs:
- ActionRequest: `DLNADOC/1.50 SEC_HHP_[TV] Samsung Q70 Series/1.0 UPnP/1.0`
- FileRequest: `samsung-agent/1.1`

I guess the reason about this is, that clients use different libraries for each of those. For UPnP services they use something like libupnp. For file request they use hand made code or any other library.

### DIFF
--- a/src/util/upnp_clients.cc
+++ b/src/util/upnp_clients.cc
@@ -52,7 +52,7 @@ std::vector<struct ClientInfo> Clients::clientInfo = std::vector<struct ClientIn
         "BubbleUPnP" },
 
     // This is AllShare running on a PC. We don't want to respond with Samsung capabilities, or Windows (and AllShare) might get grumpy.
-    // User-Agent: DLNADOC/1.50 SEC_HHP_[PC]LPC001/1.0  MS-DeviceCaps/1024
+    // User-Agent(actionReq): DLNADOC/1.50 SEC_HHP_[PC]LPC001/1.0  MS-DeviceCaps/1024
     {
         "AllShare",
         ClientType::SamsungAllShare,
@@ -60,7 +60,7 @@ std::vector<struct ClientInfo> Clients::clientInfo = std::vector<struct ClientIn
         ClientMatchType::UserAgent,
         "SEC_HHP_[PC]" },
 
-    // User-Agent: DLNADOC/1.50 SEC_HHP_[TV] Samsung Q7 Series (49)/1.0
+    // User-Agent(actionReq): DLNADOC/1.50 SEC_HHP_[TV] Samsung Q7 Series (49)/1.0
     {
         "Samsung Series [Q] TVs",
         ClientType::SamsungSeriesQ,
@@ -68,7 +68,7 @@ std::vector<struct ClientInfo> Clients::clientInfo = std::vector<struct ClientIn
         ClientMatchType::UserAgent,
         "SEC_HHP_[TV] Samsung Q" },
 
-    // User-Agent: DLNADOC/1.50 SEC_HHP_BD-D5100/1.0
+    // User-Agent(actionReq): DLNADOC/1.50 SEC_HHP_BD-D5100/1.0
     {
         "Samsung Blu-ray Player BD-D5100",
         ClientType::SamsungBDP,
@@ -76,9 +76,9 @@ std::vector<struct ClientInfo> Clients::clientInfo = std::vector<struct ClientIn
         ClientMatchType::UserAgent,
         "SEC_HHP_BD" },
 
-    // User-Agent: DLNADOC/1.50 SEC_HHP_[TV]UE40D7000/1.0
-    // User-Agent: DLNADOC/1.50 SEC_HHP_ Family TV/1.0
-    // User-Agent: DLNADOC/1.50 SEC_HHP_[TV] UE65JU7000/1.0 UPnP/1.0
+    // User-Agent(actionReq): DLNADOC/1.50 SEC_HHP_[TV]UE40D7000/1.0
+    // User-Agent(actionReq): DLNADOC/1.50 SEC_HHP_ Family TV/1.0
+    // User-Agent(actionReq): DLNADOC/1.50 SEC_HHP_[TV] UE65JU7000/1.0 UPnP/1.0
     {
         "Samsung other TVs",
         ClientType::SamsungSeriesCDE,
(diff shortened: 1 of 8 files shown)


KarlStraussberger:
I'm not quite sure whether this works with multiple clients on same machine (like Windows or FireTv with Kodi and VLC used in parallel)

KarlStraussberger (commented):
I'm not quite sure whether this works with multiple clients on same machine (like Windows or FireTv with Kodi and VLC used in parallel)

pamapa:
You are right it wont work in parallel only in sequence. Any ideas to improve this dilemma?

KarlStraussberger:
I do on my Windows PC. Movies with Kodi, Music with Foobar2000.

One solution might be to allow multiple UserAgents per Client and merge cache onto one item (by ip). It may still become nasty if two different clients send similar UserAgent in FileRequest. Isn't there something like a deviceID which we get on discovery for the client?

pamapa:
Discovery wont help as it comes async in, we only have addr+user-agent to look for in the requests.

The main question is for what are we doing this. The currently only propose for all this is to apply quirks for special clients like Samsung TV and others.

What we can do is to pass userAgent always inside the Quirks class. This should work for Samsung TVs, as the user-agent of the `request file` case wont be part in the clients info table. Then cache will win. And you case should work too. As we each time update the cache with latest info.

If you agree on this, i will adapt the merge request. Let me know.

pamapa:
Also see Samsung User-Agent info in updated description above.

src/util/upnp_clients.cc
```diff
@@ -225,28 +161,40 @@ void Clients::getInfo(const struct sockaddr_storage* addr, const std::string& us
...
                 }
             }
         }
+
+        return false;
+    });
+
+    if (it != clientInfo.end()) {
+        *ppInfo = &(*it);
+#if false
```
whyman, line 193: Bin this or use TOMBDEBUG

pamapa:
@KarlStraussberger Could you please collect the User-Agent info of Kodi and Foobar2000 of both cases (ActionRequest + FileRequest).

src/util/upnp_clients.cc
pamapa: ok i will adapt the code

src/util/upnp_clients.cc
```diff
@@ -255,18 +203,75 @@ bool Clients::getInfoByAddr(const struct sockaddr_storage* addr, const ClientInf
...
-    auto it = std::find_if(clientInfo.begin(), clientInfo.end(), [=](const auto& c) //
-        { return c.matchType == type && match.find(c.match) != std::string::npos; });
+    if (!match.empty()) {
+        auto it = std::find_if(clientInfo.begin(), clientInfo.end(), [=](const auto& c) //
+            { return c.matchType == type && match.find(c.match) != std::string::npos; });
+
+        if (it != clientInfo.end()) {
+            *ppInfo = &(*it);
+#if TOMBDEBUG
+            log_debug("found client by type (match='{}')", match.c_str());
```
whyman, line 213: log_debug get compiled out if debug is off anyway, so dont need the #if

pamapa: my idea first hand was to not have log_debug, as is not log_trace i made #if 0 around it. what shall we do now?

pamapa: ok the more debug info the better, i just removed the #if

KarlStraussberger:
> 
> 
> @KarlStraussberger Could you please collect the User-Agent info of Kodi and Foobar2000 of both cases (ActionRequest + FileRequest).

Kodi: `UPnP/1.0 DLNADOC/1.50 Kodi `+ `Kodi/18.9 (Windows NT 10.0.19041; Win64; x64) App_Bitness/64 Version/18.9-(18.9.0)-Git:20201023-0655c2c718`

Foobar2000 : `UPnP/1.0 DLNADOC/1.50 Platinum/1.0.4.2-bb / foobar2000` + `foobar2000/1.6.2`

pamapa:
@KarlStraussberger Thanks for the info. In both cases we could add a working match iff needed. In that case if i do not touch `src/file_request_handler.cc in this merge request` it will work.

pamapa:
Ok code is now adapted.

Maybe its time to start a documentation with User-Agent info of various clients we can get of.
@whyman 
Where should we document this?

whyman:
The [Supported Device](https://docs.gerbera.io/en/stable/supported-devices.html) doc already has some info on the flags and its most relevant there, I think.

pamapa:
@whyman i would like to add all info we can get, even iff we do not directly support the device. The info would be for understanding the code. Maybe a comment containing the list at the beginning of upnp_clients.cc would be sufficient.

whyman:
Ah, I see. If it were me I would probably create unit tests with the values from real world devices.

If you are thinking writing it down, a comment would be closest in terms of context.

KarlStraussberger:
As far as I understand the change now, we configured clients as before. The cache is only using the source IP, so the second client overwrites the first.

I'm sorry but again my Windows machine causes trouble: `Microsoft-Windows/10.0 UPnP/1.0 Microsoft-DLNA DLNADOC/1.50` is sent by Explorer.

**New idea**: Add a multi-client flag to client config and use this in updateCache to check for userAgent also. If you add match to the cacheEntry, you can use this to compare instead of the full userAgent string.

pamapa:
> As far as I understand the change now, we configured clients as before. The cache is only using the source IP, so the second client overwrites the first.

Your understanding is correct.

> I'm sorry but again my Windows machine causes trouble: `Microsoft-Windows/10.0 UPnP/1.0 Microsoft-DLNA DLNADOC/1.50` is sent by Explorer.
This should match the `ClientType::StandardUPnP`. What happens when you add Kodi and Foobar2000 explicit as `ClientType::StandardUPnP`:

```
    {
        "Standard UPnP (Kodi)",
        ClientType::StandardUPnP,
        QUIRK_FLAG_NONE,
        ClientMatchType::UserAgent,
        "Kodi" },
    {
        "Standard UPnP (Foobar2000)",
        ClientType::StandardUPnP,
        QUIRK_FLAG_NONE,
        ClientMatchType::UserAgent,
        "foobar2000" },

before "Standard UPnP"...
```

> **New idea**: Add a multi-client flag to client config and use this in updateCache to check for userAgent also. If you add match to the cacheEntry, you can use this to compare instead of the full userAgent string.
I do not understand this solution. userAgent is different for e.g. Kodi depending on the request type...

pamapa:
i will add unit tests...

whyman:
I guess some kind of "best effort" fuzzy match could help:

1: From same IP
2. User agent is stripped of `UPnP/1.0`, `DLNADOC/1.50` (Protocol details)
3. Split by space, dash, or forward splash
4. More than one word match with an existing entry is considered a hit

Rough worked example
```
1. New IP
2. UPnP/1.0 DLNADOC/1.50 Platinum/1.0.4.2-bb / foobar2000
3. Platinum/1.0.4.2-bb / foobar2000
4. [Platinum, 1.0.4.2-bb, foobar2000]
```

Then
```
1. Matches existing IP
1. foobar2000/1.6.2
2. foobar2000/1.6.2
3. [foobar2000, 1.6.2]
```

foobar is in both - profit

This still doesnt work the VLC case though, so not perfect.

KarlStraussberger:
> > **New idea**: Add a multi-client flag to client config and use this in updateCache to check for userAgent also. If you add match to the cacheEntry, you can use this to compare instead of the full userAgent string.
> 
> I do not understand this solution. userAgent is different for e.g. Kodi depending on the request type...

My comment was related to caching where you only use the IP. I was thinking about a flag in `ClientInfo`/`ClientConfig` in conjunction with IP to enable storage of multiple clients per IP (matching by UserAgent).

pamapa:
ready to review now with fixes + unit test code.

I guess the unit test code is the best to further talk about what is not working -> please provide a test...

pamapa:
Also consider my main concern:
Samsung TVs might just not return a sane match-able userAgent info in the file request case. E.g. out of `samsung-agent/1.1` we simply do not know which version of SamsungTV. When i look at other DLNA code, i can see that older SamsungTVs needed different quirks than newer...

My code changes work with this situation!

PR #1121, opened 2021-01-12
make Clients class non static

pamapa:
now that we have context it is quiet easy....

Saw this when writing unit tests.


whyman:
Nice.

PR #1122, opened 2021-01-13
clang-tidy fixes

neheb:


neheb:
Last commit was largely done manually.

src/util/url.cc
```diff
@@ -154,7 +154,7 @@ std::unique_ptr<URL::Stat> URL::getInfo(const std::string& URL, CURL* curl_handl
     }
 
     std::string mt = ct ? ct : MIMETYPE_DEFAULT;
-    log_debug("Extracted content length: %lld", (long long)cl);
+    log_debug("Extracted content length: {}", uint64_t(cl));
```
whyman, line 157: Surely the cast is not needed, it can format a double.

But why is double anyway?

neheb: No idea.

whyman: https://curl.se/libcurl/c/CURLINFO_CONTENT_LENGTH_DOWNLOAD.html

> Pass a pointer to a double to receive the content-length of the download. This is the value read from the Content-Length: field. Since 7.19.4, this returns -1 if the size isn't known.
> 
> CURLINFO_CONTENT_LENGTH_DOWNLOAD_T is a newer replacement that returns a more sensible variable type. 

So I guess we should switch to use the more sensible version, and just pass it without a cast.

whyman:
LGTM, do you fancy tidying up the curl stuff too?

neheb:

neheb:
Got another commit. All that stuff looks like it can be replaced with std::chrono.

PR #1126, opened 2021-01-14
replace gettimeofday with timespec_get

neheb:
gettimeofday is deprecated under POSIX and is not year 2038 safe.

This is C++17. timespec_get should be supported.


whyman:
What is the implication of select vs pselect?

PR #1136, opened 2021-01-16
some cleanups

neheb:
I'm sure manual review of these is needed.

### DIFF
--- a/src/config/config_manager.cc
+++ b/src/config/config_manager.cc
@@ -1333,16 +1333,12 @@ void ConfigManager::setOrigValue(const std::string& item, const std::string& val
 
 void ConfigManager::setOrigValue(const std::string& item, bool value)
 {
-    if (origValues.find(item) == origValues.end()) {
-        origValues[item] = value ? "true" : "false";
-    }
+    origValues.try_emplace(item, value ? "true" : "false");
 }
 
 void ConfigManager::setOrigValue(const std::string& item, int value)
 {
-    if (origValues.find(item) == origValues.end()) {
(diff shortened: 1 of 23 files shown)


whyman:
Does it run?

neheb:
It does yes.

master seems to be broken in a different way though. I can't play anything on any device.

whyman:
Well that sounds quite concerning!

neheb:
There were lots of changes recently so probably to be expected.

neheb:
looks like it's actually this PR. Wonder which commit it is.

neheb:
Oh I see what's happening. It's not master that's broken. It's gerbera with libupnp master. When I switch back to npupnp master, everything works.

whyman:
Can you explain more what the issue is? Things being broken with libupnp is not good as that's what 99 percent of users will do!

neheb:
Yeah but 99% of users will not be using a master build of libupnp and building using its shiny new CMake build :).

The issue is that none of the files play. In the web UI, when pressing a file you get a 0 byte download. On foobar2000, it just says can't find file.

neheb:
Rebased.

neheb:
Anyone?

I've tested this to work fine with foobar2000 mobile.

whyman:
Looks good to me except for the `FILE* f __attribute__((__cleanup__(clean_file)));` which seems a bit C to me.

neheb:
Well, it's a GCC attribute that works with both.

neheb:
Removed GCC cleanup attribute commit.

PR #1140, opened 2021-01-17
Load image for container chain when creating containers

### LINKED ISSUE #365: Upward propagation of media images
note: this is about a upnp client, not the gerbera UI.

I have a set of audio files (mp3 with embedded images).
Depending on where I navigate, I get images or not:
- Audio/Albums: the images propagate upwards and get displayed: fine!
- All - full name / All Audio: these display individual files and the embedded image is displayed: fine (as expected, there is no propagation)
- Artists/ Composers/ Genres/ Year: there are no images. **is there a way for me to control this?** (other upnp servers, like twonky, would propagate the image of the first contained element).

I have a set of movies (mkv plus folder.jpg, each in a folder):
Depending on where I navigate, I get images or not:
- Video/ All Video: all individual files are listed, and the folder.jpg in the containing folder is displayed: fine!
- PC Directory/[path to the list of the movie directories]: the images propagate upwards and get displayed: fine!
- Video/ Directories: there are no images, even though is lists the same directory list as the previous one. **is there a way for me to control this?**

### LINKED ISSUE #1099: Password Authentication in Docker Container
When authentication for the web interface is turned on, the configured username/password is not accepted.

### LINKED ISSUE #1139: Web UI user accounts not working
or at least it seems so, for some reason they are not loaded out the config, on master.

KarlStraussberger:
fixes #365 
fixes #1139 
fixes #1099


src/metadata/libexif_handler.cc
```diff
@@ -353,8 +353,12 @@ void LibExifHandler::process_ifd(ExifContent* content, const std::shared_ptr<Cds
     }
 }
 
-void LibExifHandler::fillMetadata(std::shared_ptr<CdsItem> item)
+void LibExifHandler::fillMetadata(std::shared_ptr<CdsObject> obj)
 {
```
whyman, line 357: If these handlers require an item then isnt it nicer to have the typing check? (as it was?)

KarlStraussberger: It's because of CdsContainer being used in some cases. I didn't want to use separate calls. Any suggestions?

whyman: I don't have a better suggestion 😬

PR #1143, opened 2021-01-18
Display album art or thumbnail in web ui

KarlStraussberger:
Getting even more colour into the web UI.

The change in ffmpeg_handler assigns the resolution of the artwork image to the correct resource!


whyman:
Nice, and great to see the docs images updated to match.

I guess we should make some kind of placeholder image for the docs in the future, and for items without images.

KarlStraussberger:
> 
> 
> Nice, and great to see the docs images updated to match.
> 
> I guess we should make some kind of placeholder image for the docs in the future, and for items without images.

It may also become necessary to rework the tests to make the screenshot contents look a bit more "real live" instead of "Test1" etc.

Issue #1066, opened 2020-12-18
No metadata import of specific 24bit/48kHz Flac files


ghlupe2:
I've got quite some 24bit/48kHz Flac files that get imported without any metadata. Debug log says "TagLibHandler: could not open flac file".
Foobar2000 reads the metadata okay and after re-generating the file with foobar2000 flac converter it gets imported well ([see log file](https://github.com/gerbera/gerbera/files/5716324/log.txt)).

This is the original file's header:

`fLaC���"€€�”�¸p�:6€%TtÏyËÒÆâÇ4¼Ø����À ���reference libFLAC 1.1.3 20070917`

This is the re-generated file's header:

`fLaC���"���9�AØ¸p�:6€%TtÏyËÒÆâÇ4¼Ø�À ���reference libFLAC 1.3.2 20170101
`

whyman:
This is one to report with the taglib maintainers.

ghlupe2:
Thanks. I did so and they say it is fixed in the current taglib head. Is there a chance to run Gerbera with it? Would I need to re-compile Gerbera after installing the latest taglib build?

whyman:
Assuming they didnt break their ABI, you should be able to simply rebuild the library without a Gerbera rebuild.

whyman:
Also, this is the taglib issue referenced for those following along at home: https://github.com/taglib/taglib/issues/985

PR #1144, opened 2021-01-18
Add bitsPerSample resource from taglib if available

### LINKED ISSUE #978: Add bits_per_sample (sample width)  to obj.res
Is there a way to make this resource information available for scripting? I would like to organise my audio files by format, like 
**FLAC | 24 bit | 96 KHz | stereo**. Taglib presents this as bitsPerSample if I'm not mistaken.

KarlStraussberger:
fixes #978 : It's available as `R_BITS_PER_SAMPLE`

Also add properties of additional resources (e.g. fanart) to script context `obj.res`


src/metadata/taglib_handler.cc
```diff
@@ -520,6 +520,13 @@ void TagLibHandler::extractASF(TagLib::IOStream* roStream, const std::shared_ptr
     }
     populateGenericTags(item, asf);
 
+    auto audioProps = asf.audioProperties();
```
whyman, line 523: There is already handling for audioProperties in `populateGenericTags`, should be able to add it there.

KarlStraussberger: `TagLib::File` does not have the required audioProperties. The base method returns a simpler type with fewer methods.

ghlupe2:
That does not seem to work. On the web UI I could not find a single flac file where "Edit Item, Show Details, bitsPerSample" is not empty. The reason might be that some file types like mp3 do NOT contain this information by design. I guess that taglib provides this information only on file type basis, but not overall.

KarlStraussberger:
That's why the code looks duplicated. If there is no value, there are two options
- the file does not contain it - check it with tagreader
- taglib does not properly read it - what's your taglib version?

ghlupe2:
Thanks for looking into this, I don't think that tagreader can read bitsPerSample, it is not covered by audioproperties.cpp, but exiftool does for all flac files I tested. My taglib version is the current one from github (1.12).

ghlupe2:
I assume that in your environment this feature works well? 
I double checked against install_manifest.txt that all taglib 1.11.1 related files have been removed before 1.12 install. And taglib 1.11.1 related issue #1066 is fixed now, so I'm pretty sure that I am on a current Gerbera (v1.8.2-166-gc8e2bbe0) and Taglib version.

KarlStraussberger:
Proof of the working feature at my server:
grafik

Did you reimport your library? Metadata are not updated without rescanning, i.e. stop gerbera, delete your sqlite-db file and start gerbera again.

ghlupe2:
Great! This is an example of a file I just imported:
Screenshot 2021-07-19 11 49 17
Are more libraries or software to be considered in this process? My list of resources looks quite different (colorDepth, dc:language, ...)

KarlStraussberger:
Everything should be done by taglib.

From the empty entries I can see that your running the first version using grb_cds_resource. Please update to the current head (like always, some flaws made it into the main branch). Maybe that solves the issue.

ghlupe2:
I am on an almost current version but still see strange effects regarding this issue. Flac files with integrated artwork show the bits per sample information in Gebera Web UI, all the others don't. 

Re-tagging such files adding/removing artwork changes the behavior vice versa. 

An Apple lossless file without integrated artwork works well.

foobar2000, Picard, and exiftool display bits per sample for any of the above files in question. 

Can anybody confirm this behavior?

KarlStraussberger:
Now, I understood your issue, found the source and created a new issue to track its fix

ghlupe2:
Your latest fix does solve the issue on my side as well, thanks!

PR #1145, opened 2021-01-19
Reworked ConfigGenerator to benefit from definitions in ConfigManager.

KarlStraussberger:
Most of the XML tags are gone now, i.e. changes in config layout are reflected in ConfigGenerator.

### DIFF
--- a/src/config/config_generator.cc
+++ b/src/config/config_generator.cc
@@ -28,56 +28,178 @@
 #include <string>
 #include <utility>
 
+#include "config/config_setup.h"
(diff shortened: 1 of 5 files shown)


whyman:
Not sure I fully grok all these changes, so LGTM!

KarlStraussberger:
I didn't want to annoy you :-)

It should/could be quite easy: ConfigSetup knows the xpath of each option. Except those options that were already hacked in ConfigManager. 

Now, with all options in place we have a link to straighten things out. Especially, transcoding options, which took half the time!

Issue #1142, opened 2021-01-17
Add hexdump check to CMake


whyman:
Its a build time dep now

neheb:
what is hexhump?

neheb:
Can xxd be used instead?

whyman:
Haha! I hadnt noticed that typo :laughing: 

Im sure xxd could be used as an alternative, patches are welcome :)

neheb:
xxd -i seems to have similar output. What do you think?

whyman:
Could wortk with a a bit of post processing, I dont see what it gains though

neheb:
In OpenWrt, we have xxd available but not hexdump.

whyman:
Closed by #1183 instead

PR #1153, opened 2021-01-24
Switch from hexdump to xxd

whyman:
Hexdump is apparently not available on OpenWrt

Closes: #1142

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -15,6 +15,7 @@ packages:
   - libexif
   - libmatroska
   - libiconv
+  - vim # for xxd
 sources:
   - https://github.com/gerbera/gerbera
 tasks:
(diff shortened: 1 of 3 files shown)


whyman:
Hah, of course it doesnt work on busybox :/

neheb:
either I'm on crack or OpenWrt's busybox includes hexdump.

neheb:
Yeah forget about it. hexdump is part of util-linux which is included in every linux installation.

I think the packages in OpenWrt that use xxd need to be fixed.

whyman:
Okay, I will update the patch to check for hexdump as originally planned.

Is there any way we can include OpenWrt in our CI? Docker builds or something?

neheb:
Good question. There is https://github.com/openwrt/docker but that's for the entire tree.

ping @aparcar

aparcar:
You can use a [GitHub action](https://github.com/openwrt/gh-action-sdk) to build the package from a custom feed or use the `openwrtorg/sdk` container directly.

Issue #1109, opened 2021-01-10
Improve DockerHub README


whyman:
Currently the Docker Hub page simply uses the Github README which is not that useful if just looking to run Gerbera via docker.

Proposed solution:

- New `README.Docker.md` file with Docker specific info
- Implement updates via the [DockerHub Description](https://github.com/peter-evans/dockerhub-description) action.

PR #1154, opened 2021-01-24
Update README, DockerHub Readme

whyman:
Dont try to create a config file if it exists
Dont put git dirty status into build version
Add npupnp flag to README

Closes: #1109
Closes: #1193

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -35,8 +35,8 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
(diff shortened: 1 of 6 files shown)


neheb:
Is a docker-compose style installation supported? eg: https://hub.docker.com/r/linuxserver/daapd

I think that would be useful as many people install via portainer and such.

There's also the networking issue. I have no idea if gerbera is supposed to use host networking or the default bridge one. daapd for example uses host because of mdns. Last time I tried creating the container in bridge mode it didn't work.  I attributed it to libupnp weirdness.

I see arm containers which is quite nice :).

PR #1159, opened 2021-01-26
Allow replacement of internal url by something different.

### LINKED ISSUE #1160: Display album art behind reverse proxy
The new feature to [display album art in the web UI](https://github.com/gerbera/gerbera/commit/d072f69375e7b930405b110e0c82ed4480c6da10) is cool.
If it would also be possible from "outside" localhost -> that would be great!

When using gerbera behind a reverse proxy "localhost" doesn't work in that scenario.
If you - by any chance - could find the time and solution, I would be very happy :)

Thanks a lot for your work - it is much appreciated!

KarlStraussberger:
May be helpful when running behind reverse proxies

fixes #1160


whyman:
Nice!

Issue #1149, opened 2021-01-23
Change in Artist handling


neheb:
Some recent commit to master changed artist handling such that it only looks at albumartist and not artist. Here's an example:

![Screenshot_20210123-135304](https://user-images.githubusercontent.com/2918873/105615072-9319b800-5d82-11eb-9b24-e1ec16f17f52.png)

the artist field is populated but not albumartist.

KarlStraussberger:
I don't have an issue.

Can you double-check by looking at the details on the web ui. Do you get any warning in the log? Which version of taglib are you using, it is responsible for reading the tags?

neheb:
```
libtag1v5-vanilla/stable,now 1.11.1+dfsg.1-0.3+deb10u1 armhf [installed,automatic]
libtag1v5/stable,now 1.11.1+dfsg.1-0.3+deb10u1 armhf [installed,automatic]
```

Nothing interesting in the logs. I assume I would need to run gerbera with -D to find something.

neheb:
I have WITH_EXIV2, WITH_FFMPEGTHUMBNAILER, and WITH_AVCODEC if that makes a difference.

neheb:
```

Metadata
--
dc:date | 2000-01-01
dc:description | Astral Projection, In the Mix, Let There Be Light, 2000, Trance
dc:title | Let There Be Light
upnp:album | In the Mix
upnp:artist | Astral Projection
upnp:date | 2000
upnp:genre | Trance
upnp:originalTrackNumber | 3
```

shows up in the webui. Weird.

KarlStraussberger:
Please post output of

`gerbera --compile-info`

`ldd gerbera`

neheb:
```
Gerbera UPnP Server version 1.7.0_git - http://gerbera.io/

===============================================================================
Gerbera is free software, covered by the GNU General Public License version 2

Copyright 2016-2021 Gerbera Contributors.
Gerbera is based on MediaTomb: Copyright 2005-2010 Gena Batsyan, Sergey Bostandzhyan, Leonhard Wimmer.
===============================================================================
Compile info
-------------
WITH_MAGIC=ON
WITH_MYSQL=OFF
WITH_CURL=1
WITH_INOTIFY=1
WITH_JS=1
WITH_TAGLIB=1
WITH_AVCODEC=ON
WITH_FFMPEGTHUMBNAILER=ON
WITH_EXIF=1
WITH_EXIV2=ON
[...]

PR #1162, opened 2021-01-27
fix broken for loop breaking artist display

neheb:
0242666a65befc51832dd057f2d2b9e22469882c broke this by incorrectly
converting several statements to a loop. What happens with that commit
is that the second vector member does not even get looked at.

Change so that it's a simple loop. Even though two members are the same,
it seems to work properly as before.

Also cleaned up the surrounding code.

Fixes: https://github.com/gerbera/gerbera/issues/1149

### DIFF
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -126,22 +126,19 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t 
         if (upnp_class == UPNP_CLASS_MUSIC_ALBUM) {
             auto meta = obj->getMetadata();
 
-            const auto albumProperties = std::array<std::pair<const char*, std::vector<metadata_fields_t>>, 5> {
+            constexpr auto albumProperties = std::array<std::pair<const char*, metadata_fields_t>, 6> {
                 {
-                    { "dc:creator", { M_ALBUMARTIST, M_ARTIST } },
-                    { "upnp:composer", { M_COMPOSER } },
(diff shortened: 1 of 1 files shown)


src/upnp_xml.cc
```diff
@@ -126,22 +126,19 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
...
+                    { "upnp:date", M_UPNP_DATE },
                 }
             };
-            for (const auto& [tag, fields] : albumProperties) {
+            for (const auto& [tag, field] : albumProperties) {
                 std::string value = "";
-                for (const auto& field : fields) {
-                    value = getValueOrDefault(meta, MetadataHandler::getMetaFieldName(field));
-                    if (!value.empty())
-                        break;
```
neheb, line 143: maybe this should have been continue...

KarlStraussberger:
The point is, that artist is removed from the container metadata. So if your files do not provide album artist you don't get it anymore! 
You can double check in web ui or with tagreader.

The loop was fine. Your change may produce two `creator` entries. The reason is in `addContainerChain`

neheb:
That's a regression IMO. I expect an artist even when album artist is missing.

KarlStraussberger:
It looks to me that, foobar mobile takes the album artist from the parent container if there is no album artist in the file itself.

see #1163

neheb:
Closing in favor of https://github.com/gerbera/gerbera/pull/1163

PR #1163, opened 2021-01-27
Copy artist to album artist for container

### LINKED ISSUE #1149: Change in Artist handling

KarlStraussberger:
Ensures that album artist is set on container

fixes #1149


src/upnp_xml.cc
```diff
@@ -126,22 +126,18 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
         if (upnp_class == UPNP_CLASS_MUSIC_ALBUM) {
             auto meta = obj->getMetadata();
 
-            const auto albumProperties = std::array<std::pair<const char*, std::vector<metadata_fields_t>>, 5> {
+            const auto albumProperties = std::array<std::pair<const char*, metadata_fields_t>, 5> {
                 {
-                    { "dc:creator", { M_ALBUMARTIST, M_ARTIST } },
-                    { "upnp:composer", { M_COMPOSER } },
```
KarlStraussberger, line 132: Artist is no longer in container metadata

src/upnp_xml.cc
```diff
@@ -126,22 +126,18 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
         if (upnp_class == UPNP_CLASS_MUSIC_ALBUM) {
             auto meta = obj->getMetadata();
 
-            const auto albumProperties = std::array<std::pair<const char*, std::vector<metadata_fields_t>>, 5> {
+            const auto albumProperties = std::array<std::pair<const char*, metadata_fields_t>, 5> {
```
neheb, line 129: this can be constexpr.

src/content/content_manager.cc
```diff
@@ -1060,6 +1060,13 @@ int ContentManager::addContainerChain(const std::string& chain, const std::strin
...
 
     log_debug("received chain: {} -> {} ({}) [{}]", chain.c_str(), newChain.c_str(), lastClass.c_str(), dictEncodeSimple(lastMetadata).c_str());
+    // copy artist to album artist if empty
+    const auto aaItm = lastMetadata.find(MetadataHandler::getMetaFieldName(M_ALBUMARTIST));
+    const auto taItm = lastMetadata.find(MetadataHandler::getMetaFieldName(M_ARTIST));
+    if (aaItm == lastMetadata.end() && taItm != lastMetadata.end()) {
+        lastMetadata[MetadataHandler::getMetaFieldName(M_ALBUMARTIST)] = taItm->second;
+    }
+
     const std::array<metadata_fields_t, 4> unwanted { M_DESCRIPTION, M_TITLE, M_TRACKNUMBER, M_ARTIST }; // not wanted for container!
```
neheb, line 1070: This can be constexpr.

whyman:
From rough recollection CDS standard says something like this for `musicAlbum` fields:

| Field       | Namespace | Req |
|-------------|-----------|-----|
| artist      | upnp      | No  |
| genre       | upnp      | No  |
| producer    | upnp      | No  |
| albumArtURI | upnp      | No  |
| toc         | upnp      | No  |

neheb:
I just checked this. So is albumartist...

KarlStraussberger:
> 
> 
> From rough recollection CDS standard says something like this for `musicAlbum` fields:
> Field 	Namespace 	Req
> artist 	upnp 	No
> genre 	upnp 	No
> producer 	upnp 	No
> albumArtURI 	upnp 	No
> toc 	upnp 	No
> 
> I assume we will have extras internally for e.g. scripting purposes though.

At the moment we send none of these for album (except `albumArtURI`). Added some more.

whyman:
> I just checked this. So is albumartist...

I just checked the CDSv4 pdf and couldnt see it, only artist?

1. http://upnp.org/specs/av//UPnP-av-ContentDirectory-v4-Service.pdf (page 258)

KarlStraussberger:
It's more than standard now so the clients can pick want they understand. We need to add editing (scripting support) for album properties instead of picking from some random track.

neheb:
@whyman http://upnp.org/specs/av//UPnP-av-ContentDirectory-v3-Service.pdf page 215

whyman:
> instead of picking from some random track.

You would expect that any track in album would have the appropriate album data, there is no "file" for them after all.

neheb:
Hmm this PR is causing Gerbera to crash when foobar2000 enters Gerbera.

KarlStraussberger:
Can you share a stacktrace

neheb:
not crashing anymore. weird.

Issue #1165, opened 2021-01-28
Resources not working


twdkeule:
Hi,

I have added a resources section to the configuration to add fanart and subtitles:
```
  <import>
    [...]
    <resources>
      <subtitle>
        <add-file name="%filename%.srt"/>
        <add-file name="%filename%.en.srt"/>
        <add-file name="%filename%.nl.srt"/>
      </subtitle>
      <fanart>
        <add-file name="%filename%-thumb.jpg"/>
        <add-file name="poster.jpg"/>
      </fanart>
      <container>
        <add-file name="poster.jpg"/>
      </container>
      <resource>
        <add-file name="%filename%.srt"/>
        <add-file name="%filename%.en.srt"/>
        <add-file name="%filename%.nl.srt"/>
        <add-file name="%filename%-thumb.jpg"/>
        <add-file name="poster.jpg"/>
      </resource>
    </resources>
  </import>
```

However, when testing with VLC, subtitles don't seem to be loaded.
Fanart works for the movie items themselves, but not for containers.

Loggin:

Fanart found:
```
Jan 28 09:32:58 acer-laptop gerbera[101111]: 2021-01-28 09:32:58  debug: [fanart_handler.cc:56] getFanArtPath(): Folder name: /.../film/Aladdin (2019)
Jan 28 09:32:58 acer-laptop gerbera[101111]: 2021-01-28 09:32:58  debug: [fanart_handler.cc:63] getFanArtPath(): folder.jpg: missing
Jan 28 09:32:58 acer-laptop gerbera[101111]: 2021-01-28 09:32:58  debug: [fanart_handler.cc:63] getFanArtPath(): poster.jpg: found
Jan 28 09:32:58 acer-laptop gerbera[101111]: 2021-01-28 09:32:58  debug: [fanart_handler.cc:87] serveContent(): FanArt: Opening name: /.../film/Aladdin (2019)/poster.jpg
```

Subtitles not found:
```
Jan 28 09:33:03 acer-laptop gerbera[101111]: 2021-01-28 09:33:03  debug: [server.cc:485] createRequestHandler(): Filename: /content/media/object_id/10187/res_id/0/ext/file.srt
Jan 28 09:33:03 acer-laptop gerbera[101111]: 2021-01-28 09:33:03  debug: [file_request_handler.cc:69] getInfo(): start
[...]

PR #1166, opened 2021-01-28
Use subtitle resource for SAMSUNG clients if available

KarlStraussberger:
Remove unused method

Found this while researching on #1165 

There is still a bit of hardcoding for srt files which should be bypassed for explicit resources already.

### DIFF
--- a/test/core/test_upnp_xml.cc
+++ b/test/core/test_upnp_xml.cc
@@ -125,6 +125,71 @@ TEST_F(UpnpXmlTest, RenderObjectItem)
     EXPECT_STREQ(didl_lite_xml.c_str(), expectedXml.str().c_str());
 }
 
+TEST_F(UpnpXmlTest, RenderObjectItemWithResources)
+{
+    // arrange
+    pugi::xml_document didl_lite;
+    auto root = didl_lite.append_child("DIDL-Lite");
+    auto obj = std::make_shared<CdsItem>();
+    obj->setID(42);
+    obj->setParentID(2);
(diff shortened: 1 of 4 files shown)


test/core/test_upnp_xml.cc
```diff
@@ -125,19 +125,6 @@ TEST_F(UpnpXmlTest, RenderObjectItem)
     EXPECT_STREQ(didl_lite_xml.c_str(), expectedXml.str().c_str());
 }
 
-TEST_F(UpnpXmlTest, CreatesSecCaptionInfoElement)
```
whyman, line 128: It would be nice to add some tests too

whyman:
If we enabled the subs patterns:
``` 
"<filename>.srt",
"<filename>.ssa",
"<filename>.smi",
"<filename>.sub"
```
By default, then we could get rid of that munging entirely?

KarlStraussberger:
> 
> 
> If we enabled the subs patterns:
> 
> ```
> "<filename>.srt",
> "<filename>.ssa",
> "<filename>.smi",
> "<filename>.sub"
> ```
> 
> By default, then we could get rid of that munging entirely?

I'd leave that change for some time after 1.7 (feel free to create an issue). The code is note completely clear to me and adding all sub patterns is a performance killer.

test/core/test_upnp_xml.cc
KarlStraussberger: Added a combined test for albumart and subtitle

PR #1167, opened 2021-01-29
clang-analyzer fixes

neheb:


src/metadata/matroska_handler.cc
```diff
@@ -152,6 +152,7 @@ void MatroskaHandler::parseMKV(const std::shared_ptr<CdsItem>& item, MemIOHandle
 {
     file_io_callback ebml_file(item->getLocation().c_str());
     EbmlStream ebml_stream(ebml_file);
+    *p_io_handler = nullptr;
```
KarlStraussberger, line 155: That has to be moved to `serveContent`. The method is called with `nullptr` from `fillMetadata`.

KarlStraussberger (changes requested):

neheb:
Looks like that did it. I also see a different error there:

```
src/metadata/matroska_handler.cc:267:25: warning: Potential memory leak [cplusplus.NewDeleteLeaks]
        attachedFile = &GetNextChild<KaxAttached>(*attachments, *attachedFile);
```

I have no idea about that one.

KarlStraussberger:
> Looks like that did it. I also see a different error there:
> 
> ```
> src/metadata/matroska_handler.cc:267:25: warning: Potential memory leak [cplusplus.NewDeleteLeaks]
>         attachedFile = &GetNextChild<KaxAttached>(*attachments, *attachedFile);
> ```
> 
> I have no idea about that one.

I prefer potential memory leaks to segfaults, thus I would not touch that code without looking into the API docs.

PR #1169, opened 2021-01-29
Revert docker build matrix

whyman:
By running all the builds in a single job we dont have to worry about building a manifest list ourselves

In the current situation whichever job finishes last overwrites the tag to point at its image, instead of building a multi-arch list.

Practically this means if you run `gerbera/gerbera:master` on amd64 currently, it tries to run the armv8 image :-1:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -12,9 +12,6 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


whyman:
43 mins sucks, we can come up with caching to help that in the future.

PR #1170, opened 2021-01-29
Docker: Update to Alpine 3.13 base

whyman:
The lovely alpine people have now added libupnp and duktape packages
(and gerbera itself!) with 3.13, so we no longer have to build them.

Add git and bash to builder as needed for version and db header gen.


whyman:
Gets a the docker build time down to 33 minutes from 54 by avoiding the lib builds

Issue #33, opened 2017-01-17
Autoscan directories are not scanned at start-up


gburca:
The `config.xml` file allows the user to specify directories to be auto-scanned (`import - scripting - autoscan`). When the application starts up with no pre-existing database, a new database is created, but the directories listed in the config file are not scanned to populate the database.

v00d00:
Being lazy here, can you post some example XML?

gburca:
I looked at it closer, and it turns out I placed the `<autoscan>` in the wrong section. After fixing that I hit another snag. Will open a new issue for it though.

For reference, here's where the autoscan entry goes:

```
<config ...>
  ...
  <import ...>
    <autoscan use-inotify="auto">
      <directory location="/some/data/path/MediaTomb"
        mode="inotify" level="full" recursive="yes" hidden-files="no"/>
      <directory location="/some/timed/path"
        mode="timed" level="full" recursive="yes" hidden-files="no" interval="86400"/>
    </autoscan>
  </import>
</config>
```

PR #1173, opened 2021-01-30
Fix Docker logged version info

whyman:
checkout@v2 fetches only a single commit by default, which makes us unable to generate a version from tag info


whyman:
` #33 1.102 -- Gerbera Version: v1.6.4-167-g34f59c00` :tada:

PR #1178, opened 2021-01-31
convert const std::array to constexpr

neheb:
Added extra stuff to make it possible.


neheb:
Hmm guess not.

PR #1181, opened 2021-01-31
Improved building with conan

r4start:
Changes: 
- cmake uses correct module files
 - names of Arch Linux packages were corrected
 - use mariadb connector from conan instead of system one


whyman:
Thanks for the patch, looks like conan build is tripping up on libuuid in CI.

I am not convinced that anyone had actually tried conan on arch before, so good to iron the package names out :laughing: 

As for the MySQL changes, it would be good to make it not conditional on conan, but search for mariadb-connector-c and if that fails search for mysql, as its useful outside conan too.

As for moving the modules, can we move them to `modules` instead of `common` if they have to be in a subfolder?

r4start:
Fixed build issues. Strangely js tests have failed. I'll look into them during this week.
Other suggestions have been fixed.

KarlStraussberger:
> Fixed build issues. Strangely js tests have failed. I'll look into them during this week.

Js test sometimes needs a second kick. Likely a timing in headless tests.

CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ if (CONAN_EXPORTED)
     conan_check_compiler()
     conan_set_libcxx()
     conan_set_vs_runtime()
+    list(APPEND CMAKE_MODULE_PATH "${CMAKE_BINARY_DIR}/")
+else()
+    list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
```
whyman, line 29: Isnt this handled by the "/modules" above?

Issue #1156, opened 2021-01-25
zpipe: cannot execute binary file: Exec format error


joerg-krause:
Since commit c443248787dcfb417191637a16daffde747faae7 cross-compiling gerbera fails because of the tool `zpipe`:

```
>>> gerbera custom Building
PATH="/home/joerg/buildroot/output/host/bin:/home/buildroot/output/host/sbin:/usr/local/bin:/usr/local/sbin"  /usr/bin/make -j9  -C /home/joerg/buildroot/output/build/gerbera-custom/
Scanning dependencies of target zpipe
[  1%] Building C object CMakeFiles/zpipe.dir/src/database/create/zpipe.c.o
[  2%] Linking C executable zpipe
[  2%] Built target zpipe
[  3%] Generating sqlite3_create_sql.h
build_headers.sh: line 35: /home/joerg/buildroot/output/build/gerbera-custom/zpipe: cannot execute binary file: Exec format error
make[4]: *** [CMakeFiles/libgerbera.dir/build.make:85: sqlite3_create_sql.h] Error 126
make[3]: *** [CMakeFiles/Makefile2:152: CMakeFiles/libgerbera.dir/all] Error 2
make[2]: *** [Makefile:171: all] Error 2
make[1]: *** [package/pkg-generic.mk:269: /home/joerg/buildroot/output/build/gerbera-custom/.stamp_built] Error 2
make: *** [common.mk:70: _all] Error 2
```

The issue is that the tool `zpipe` is cross-compiled for the target architecture, but CMake tries to run it on the host.

whyman:
Oh interesting! I didnt think of this scenario.

I wonder how we can address this, can CMake know about the host compiler? I guess not?

joerg-krause:
My workaround is to use a Python command instead of zpipe. I dropped zpipe from the CMakeLists.txt and replaced line 35 with the following Python command:

``` sh
$PYTHON3_PATH -c "import zlib,sys; sys.stdout.buffer.write(zlib.compress(sys.stdin.buffer.read()))" < ${FN} > "${FN}.Z"
```

If you can think of that as a solution, then I'll prepare a PR for replacing zpipe with a Python command.

whyman:
Your python based solution is nice, but would add python a build time dependency everywhere.

What I am thinking currently is "why do we need this SQL in the executable?"

I dont think we do

We already install the plain SQL as text files in our install target, so we should probably just load those files and stop padding out the executable with extra stuff for no real gain.

[...]

PR #1183, opened 2021-01-31
Dont copy SQL DB creation scripts into executable

whyman:
This change stops generation of zlib compressed blobs in headers
containing the initialisation SQL for the 2 DB types.

Instead we simply load it off of the filesystem.

Introduces two new config options:
/storage/sqlite|mysql/init-sql-path which allows overriding the default path.

Default path is `datadir/mysql.sql` and `datadir/sqlite3.sql` respectively.

Fixes cross compilation issues caused by trying to run a executable we had
just built with a non-native compiler at build time.

Fixes: #1156


src/database/mysql/mysql_database.cc
```diff
@@ -203,30 +194,18 @@ void MySQLDatabase::init()
...
-        if (sql_end == nullptr) {
-            throw_std_runtime_error("';' not found in mysql create sql");
-        }
-        do {
-            ret = mysql_real_query(&db, sql_start, sql_end - sql_start);
+        auto sqlFilePath = config->getOption(CFG_SERVER_STORAGE_MYSQL_INIT_SQL_PATH);
+        log_debug("Loading initialisation SQL from: {}", sqlFilePath.c_str());
+        auto sql = readTextFile(sqlFilePath);
+
+        for (const auto& statement : splitString(sql, ':')) {
```
KarlStraussberger, line 201: I suppose you want ; here

KarlStraussberger (changes requested):

src/database/mysql/mysql_database.cc
whyman: Nice spot! :facepalm:

KarlStraussberger:
We're running a bit of a security risk without hash check, but I don't see a flexible way except another config option for the hash which we default to the current value.

src/database/mysql/mysql_database.cc
KarlStraussberger: That's what mutual review is good for.

whyman:
Yeah I had a think about the security aspect, but should be mostly mitigated as this SQL is only run if the DB is empty/cannot be accessed anyway.

PR #1184, opened 2021-01-31
treewide: use fmt::to_string instead of std version

neheb:
The former is faster. More info: https://github.com/fmtlib/fmt/issues/326

I am not sure whether or not this is worth it. Thoughts?

### DIFF
--- a/src/cds_objects.cc
+++ b/src/cds_objects.cc
@@ -126,7 +126,7 @@ std::shared_ptr<CdsObject> CdsObject::createObject(unsigned int objectType)
     } else if (IS_CDS_ITEM(objectType)) {
         obj = std::make_shared<CdsItem>();
     } else {
-        throw_std_runtime_error("invalid object type: " + std::to_string(objectType));
+        throw_std_runtime_error("invalid object type: " + fmt::to_string(objectType));
     }
     return obj;
 }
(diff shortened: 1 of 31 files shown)


neheb:
https://www.zverovich.net/2020/06/13/fast-int-to-string-revisited.html indicates it's not much of a win for gcc. it's somewhat of a win with clang at 7%.

whyman:
Ran the numbers how?

If we can get solid proof this is better then I see no harm merging.

neheb:
From that page. fmt::to_string = 83,700,563 , std::string = 65,638,239

whyman:
I think we would see more improvement by changing string concat arround:

`throw_std_runtime_error("invalid object type: " + fmt::to_string(objectType));`

vs


`throw_std_runtime_error("invalid object type: " + std::to_string(objectType));`

is no real gain, where as

`throw_std_runtime_error(fmt::format("invalid object type: {}", objectType));`

Is cleaner and removes the string concats.

whyman:
And equally stuff like:
```
(std::string(XML_XMLNS) + std::to_string(CONFIG_XML_VERSION) + " " + XML_XMLNS + std::to_string(CONFIG_XML_VERSION) + ".xsd").c_str();
```
Should be converted to use fmt too

PR #1190, opened 2021-02-01
Docker: Non root support, scan /content by default

whyman:
Adds entrypoint and su-exec to drop down to a gerbera user if run as
root.
/var/run/gerbera is world writable to make arbitary users work
correctly.
Updated cofig to autoscan /content instead of /root

Fix to stop cmake complaining loudly about mariadb-connector-c not
existing.


whyman:
@eggbean, this should help you once its been merged/built.

`docker run -v /path/to/your/files:/content:ro -u $(id -u) gerbera/gerbera:master`

I will be working on Dockerhub readme updates soon, so if you have feeback on what you would like to see there, let me know.

PR #1192, opened 2021-02-02
fmt::format conversions

neheb:
using std::string's append is inefficient and costs performance. format
is also cleaner.

### DIFF
--- a/src/cds_objects.cc
+++ b/src/cds_objects.cc
@@ -126,7 +126,7 @@ std::shared_ptr<CdsObject> CdsObject::createObject(unsigned int objectType)
     } else if (IS_CDS_ITEM(objectType)) {
         obj = std::make_shared<CdsItem>();
     } else {
-        throw_std_runtime_error("invalid object type: " + fmt::to_string(objectType));
+        throw_std_runtime_error("invalid object type: {}", objectType);
     }
     return obj;
 }
@@ -176,7 +176,7 @@ void CdsItem::validate()
 
     std::error_code ec;
     if (!isRegularFile(location, ec))
-        throw_std_runtime_error("Item validation failed: file " + location.string() + " not found");
+        throw_std_runtime_error("Item validation failed: file {} not found", location.c_str());
 }
 
 //---------
@@ -243,5 +243,5 @@ std::string CdsObject::mapObjectType(unsigned int type)
         return STRING_OBJECT_TYPE_ITEM;
     if (IS_CDS_ITEM_EXTERNAL_URL(type))
         return STRING_OBJECT_TYPE_EXTERNAL_URL;
-    throw_std_runtime_error("illegal objectType: " + fmt::to_string(type));
+    throw_std_runtime_error("illegal objectType: {}", type);
 }
--- a/src/config/config_setup.cc
+++ b/src/config/config_setup.cc
@@ -51,7 +51,7 @@ pugi::xml_node ConfigSetup::getXmlElement(const pugi::xml_node& root) const
         xpathNode = root.select_node(xpath);
     }
     if (required && xpathNode.node() == nullptr) {
-        throw std::runtime_error(fmt::format("Error in config file: {}/{} tag not found", root.path(), xpath));
+        throw_std_runtime_error("Error in config file: {}/{} tag not found", root.path(), xpath);
     }
     return xpathNode.node();
 }
@@ -75,7 +75,7 @@ std::string ConfigSetup::getXmlContent(const pugi::xml_node& root, bool trim) co
     if (xpathNode.node() != nullptr) {
         std::string optValue = trim ? trimString(xpathNode.node().text().as_string()) : xpathNode.node().text().as_string();
         if (!checkValue(optValue)) {
-            throw std::runtime_error(fmt::format("Invalid {}{} value '{}'", root.path(), xpath, optValue));
+            throw_std_runtime_error("Invalid {}{} value '{}'", root.path(), xpath, optValue.c_str());
         }
         return optValue;
     }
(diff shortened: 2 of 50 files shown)


neheb:
```
C++ exception with description "invalid format string" thrown in the test body.
```

oh boy...

neheb:
Turns out my little micro optimization doesn't work:

```
config->append_attribute("xsi:schemaLocation") = fmt::format("{1}{2} {1}{2}.xsd", XML_XMLNS, CONFIG_XML_VERSION).c_str();
```

src/cds_objects.cc
```diff
@@ -126,7 +126,7 @@ std::shared_ptr<CdsObject> CdsObject::createObject(unsigned int objectType)
     } else if (IS_CDS_ITEM(objectType)) {
         obj = std::make_shared<CdsItem>();
     } else {
-        throw_std_runtime_error("invalid object type: " + fmt::to_string(objectType));
+        throw_std_runtime_error(fmt::format("invalid object type: {}", objectType));
```
KarlStraussberger, line 129: I recommend updating `throw_std_runtime_error(msg)` in `exceptions.h` to include `fmt::format`. Make code a lot clearer and remove another bunch of string concats!

KarlStraussberger (changes requested):

KarlStraussberger:
> 
> 
> Turns out my little micro optimization doesn't work:
> 
> ```
> config->append_attribute("xsi:schemaLocation") = fmt::format("{1}{2} {1}{2}.xsd", XML_XMLNS, CONFIG_XML_VERSION).c_str();
> ```

I'm not sure, but I suppose counting starts at 0. So it has to be `{0}{1}`.

src/cds_objects.cc
neheb: That's already the case with this PR.

neheb:
> 
> 
> > Turns out my little micro optimization doesn't work:
> > ```
> > config->append_attribute("xsi:schemaLocation") = fmt::format("{1}{2} {1}{2}.xsd", XML_XMLNS, CONFIG_XML_VERSION).c_str();
> > ```
> 
> I'm not sure, but I suppose counting starts at 0. So it has to be `{0}{1}`.

Weird. In any case, the current version works.

src/cds_objects.cc
KarlStraussberger: I was thinking about getting rid of the nesting.

KarlStraussberger: Maybe you can add a second macro  and we reduce it to nested macros.

```
#define throw_runtime_error_with_message(...) throw_std_runtime_error(fmt::format(__VA_ARGS__))
```

whyman:
Nice work @neheb

We can also use `#include <fmt/ostream.h>` to use filesystems built in ostream support with fmt to avoid calling `path.c_str()` everywhere

src/config/config_setup.cc
```diff
@@ -187,21 +186,21 @@ bool ConfigPathSetup::checkAgentPath(std::string& optValue)
...
-            log_error("error in configuration, transcoding profile: could not find transcoding command \"" + optValue + "\"");
+            log_error("Error in configuration, transcoding profile: could not find transcoding command \"{}\"", optValue.c_str());
             return false;
         }
         tmp_path = optValue;
     } else {
         tmp_path = findInPath(optValue);
         if (tmp_path.empty()) {
-            log_error("error in configuration, transcoding profile: could not find transcoding command \"" + optValue + "\" in $PATH");
+            log_error("error in configuration, transcoding profile: could not find transcoding command \"{}\" in $PATH", optValue.c_str());
```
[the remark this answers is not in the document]
neheb: :). There are a lot more of these.

neheb:
> 
> 
> Nice work @neheb
> 
> We can also use `#include <fmt/ostream.h>` to use filesystems built in ostream support with fmt to avoid calling `path.c_str()` everywhere

Hmm? Isn't the purpose of that to keep size down?

src/cds_objects.cc
whyman: fmt happily handles the "no-args" case, so just updating the existing macro should work

neheb:
Alright. Seems this PR also gets the size down:

```
4481928
4346536
```

neheb:
Oh I see. The macro is wrong...

src/exceptions.h
```diff
@@ -43,8 +43,7 @@
 #define PRETTY_FUNCTION __func__
 #endif
 
-#define throw_std_runtime_error(msg) throw std::runtime_error(std::string(msg) \
-    + " file:" + std::string(__FILE__) + " line:" + fmt::to_string(__LINE__) + " function:" + std::string(PRETTY_FUNCTION))
+#define throw_std_runtime_error(...) throw std::runtime_error(fmt::format("{} file: {} line: {} functions: {}", fmt::format(__VA_ARGS__), __FILE__, __LINE__, PRETTY_FUNCTION))
```
KarlStraussberger, line 47: Well, it's `functions` instead of `function`. What about turning things around: "ERROR in file {}, line {}, function {}: {}"

KarlStraussberger (approved):

KarlStraussberger:
@neheb  Really nice! No more formatting issues when texting exections!

src/exceptions.h
whyman: Perhaps a format similar to the debug messages, nice and compact:

`"[{}:{}] {} Error: {}", __FILE__, __LINE__, PRETTY_FUNCTION, fmt::format(__VA_ARGS__))`

neheb: Switched to this format string.

PR #1203, opened 2021-02-05
use clock_gettime instead of timespec_get

neheb:
The latter is not available in Haiku for some reason.


neheb:
This is quite the hack. I wonder if all this code can be converted to use std::chrono.

whyman:
I think a conversion to std::chrono would be nicer too.

extrowerk:
It builds fine on Haiku.

neheb:
macos failure is a false positive I think.

Anyway, using clock_gettime is not really portable. It happens to work on Linux and BSDs, but will fail on Windows and old macOS.

extrowerk:
Still builds fine

whyman (commented):
Shouldnt `getTimespecNow` get removed entirely thanks to chrono?

neheb:
Patches welcome :). I wasn't able to figure it out fully.

PR #1206, opened 2021-02-05
Haiku build fixes

extrowerk:
Some small build fixes for Haiku

### DIFF
--- a/src/iohandler/process_io_handler.cc
+++ b/src/iohandler/process_io_handler.cc
@@ -34,15 +34,12 @@
 #include <csignal>
 
 #include <fcntl.h>
+#include <sys/select.h>
 #include <sys/stat.h>
 #include <unistd.h>
 
 #include "content/content_manager.h"
 #include "util/process.h"
-
-#ifndef FD_SET
-#include <sys/select.h>
-#endif
 
 // after MAX_TIMEOUTS we will tell libupnp to check the socket,
 // this will make sure that we do not block the read and allow libupnp to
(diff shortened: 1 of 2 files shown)


src/iohandler/process_io_handler.cc
```diff
@@ -40,6 +40,10 @@
 #include "content/content_manager.h"
 #include "util/process.h"
 
+#ifndef FD_SET
+#include <sys/select.h>
```
neheb, line 44: Move it up before sys/stat and remove the ifndef

src/util/tools.cc
```diff
@@ -48,6 +48,10 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#ifdef __HAIKU__
+#define _DEFAULT_SOURCE
```
neheb, line 52: This is interesting. Why is DEFAULT_SOURCE needed?

neheb:
I am somewhat curious why sys/select.h is being included on every other platform. No libc++ header includes it, meaning it's some libc header implicitly including it.

neheb:
I see. Lazy libc++ coding. sys/time.h (which includes sys/select.h) gets unconditionally included in the filesystem header while it's only conditionally used.

src/util/tools.cc
extrowerk: It guards the BSD additions.
https://github.com/haiku/haiku/blob/master/headers/compatibility/bsd/ifaddrs.h#L10

neheb: Huh. Interesting.

extrowerk: Maybe just features.h should be included, idk.
https://github.com/haiku/haiku/blob/master/headers/compatibility/bsd/features.h

whyman:
Thanks!

Would be good to get a doc for the install page to add Haiku :+1:

PR #1207, opened 2021-02-06
run through include-what-you-use

neheb:
Add clang-format flag to align comments

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -31,10 +31,14 @@
 
 #include "action_request.h" // API
 
-#include <sstream>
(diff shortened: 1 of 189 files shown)


whyman:
I like the icwyu, not a fan of the comment alignment.

We would have to add a CI check to enforce it or it would get ignored

neheb:
how so? It's a clang-tidy check.

neheb:
Rebased.

whyman:
Looks good now it doesnt mess with the other comments :+1:

whyman:
@neheb whats the plan here?

PR #1217, opened 2021-02-08
Support multi volume media

### LINKED ISSUE #832: Support order by track number with multiple CDs
I have two CDs of the same album, and Gerbera sorts the titles after track number without taking the CD number into account:

1. Track 1 (CD1)
1. Track 1 (CD2)
1. Track 2 (CD1)
1. Track 2 (CD2)
1. ..

It would be nice if Gerbera would use the CD number field to sort the tracks in the correct order:

1. Track 1 (CD1)
1. Track 2 (CD1)
1. Track 1 (CD2)
1. Track 2 (CD2)

(I'm seeing that problem on gerbera 1.3.0, but I looked at the master code and did not see any changes regarding that behavior)

----

A quick fix could be to store the track number as `$trackNumber + $cdNumber * 100`. That would not need another field in the database table.

----

Original track number support was made in 3e42e3c9388cc3db2820d15f507a13c4d59be729

KarlStraussberger:
Load disc number from file meta data to sort by it.

May be extended to store season info from video files e.g.

fixes #832 

It's one of these late additions that is likely to break a release, so I'm not disappointed if it's merged after 1.7


whyman:
Amazing change @KarlStraussberger, good one for 1.7 release :tada:

PR #1218, opened 2021-02-08
Scripting Support for Container Tree

KarlStraussberger:
Allowing to transport details like container class from import script.

Changing script architecture so functions can easily be overwritten without copying complete script files.

Inject configuration to scripts


KarlStraussberger:
Instead of just setting `lastClass` in import script, this can be extended like
```
    obj.parent = { objectType: OBJECT_TYPE_CONTAINER, upnpclass: UPNP_CLASS_CONTAINER_MUSIC_ALBUM };
    obj.parent.parent = { objectType: OBJECT_TYPE_CONTAINER, upnpclass: UPNP_CLASS_CONTAINER_MUSIC_ARTIST };
```

In this example we get a modified hierarchy: "_artist_" contains "_album_".

Future changes may allow to set the first item im the container defining the image and other metadata.

The change keeps the former way to supply `lastClass`

whyman:
Perhaps a cleaner approach would be a need JS/Storage method/overload of `createContainerChain` that takes an array of objects, that API can be extended with additional fields for each level of the chain.

```js
createContainerChain([
  {"name": "spam", "class": "something", "metadata": "or something"},
  {"name": "spam", "class": "something", "metadata": "wombat"}
])
```
Which I think is what you are trying to do?

KarlStraussberger:
My intension also was to keep exiting code running without adding extra load.

On the other hand a new function offers a lot more scripting options.

KarlStraussberger:
Status: Almost done
Need to add some documentation, tests and update the js code.

src/metadata/metacontent_handler.cc
```diff
@@ -187,11 +195,17 @@ void ContainerArtHandler::fillMetadata(std::shared_ptr<CdsObject> item)
...
+    log_info("Running ContainerArt handler on {}", !path.empty() ? path.c_str() : item->getLocation().c_str());
 
     if (!path.empty()) {
         auto resource = std::make_shared<CdsResource>(CH_CONTAINERART);
-        resource->addAttribute(R_PROTOCOLINFO, renderProtocolInfo("jpg"));
+        std::string type = path.extension().string().substr(1);
+#ifdef HAVE_MAGIC
+        std::string mimeType = mime->fileToMimeType(path);
+#else
+        std::string mimeType = fmt::format("image/{}", type);
```
whyman, line 206: I guess we should should run this through the mapping configuration too?

Possibly worth pulling it out into a separate method for use here and in the main import flow?

KarlStraussberger: It got a taste of spring cleaning 😄

KarlStraussberger:
6 Days after _almost done_, it's really done!

PR #1219, opened 2021-02-09
Fixed formatting matroska date values

TVrijdag:
This PR fixes the format for dates of matroska handler. Currently the format is stored as a value.


whyman:
Thanks looks good!

I wonder if the check for an empty date should also be moved up the check the value from MKV instead of the formatted value?

I wouldn't be surprised if we have lots of videos with date "--" or similar!

KarlStraussberger:
A quick grep found no other occurences of `%Y` without proper formatting.

KarlStraussberger:
Thanks

PR #1220, opened 2021-02-09
Fixed isMetaResource for VIDEO_SUB

TVrijdag:
Hi, I'm new here :-)

I noticed that external subtitles weren't working. After some digging and debugging I discovered that `isMetaResource(VIDEO_SUB)` [would yield false](https://github.com/gerbera/gerbera/blob/cc49b2b2323155a47b19726544ab6db4e766c7b9/src/upnp_xml.cc#L710). This PR adds the handler type for external subtitle resources to the disjunction within `isMetaResource`

### DIFF
--- a/src/cds_resource.h
+++ b/src/cds_resource.h
@@ -110,7 +110,7 @@ public:
     std::string getOption(const std::string& name) const;
     bool isMetaResource(const char* rct) const
     {
-        return ((handlerType == CH_ID3 || handlerType == CH_MP4 || handlerType == CH_FLAC || handlerType == CH_FANART || handlerType == CH_CONTAINERART) && getParameter(RESOURCE_CONTENT_TYPE) == rct) || (handlerType == CH_EXTURL && getOption(RESOURCE_CONTENT_TYPE) == rct);
+        return ((handlerType == CH_ID3 || handlerType == CH_MP4 || handlerType == CH_FLAC || handlerType == CH_FANART || handlerType == CH_CONTAINERART || handlerType == CH_SUBTITLE) && getParameter(RESOURCE_CONTENT_TYPE) == rct) || (handlerType == CH_EXTURL && getOption(RESOURCE_CONTENT_TYPE) == rct);
     }
 
     bool equals(const std::shared_ptr<CdsResource>& other) const;


whyman:
Welcome!

Thanks for taking the time to put together this PR.

It looks like this change has broken a test, so might want to check that

KarlStraussberger:
Thanks for supporting us.

I just checked the test.

Looks to me that CH_SUBTITLE was left out to create the CaptionInfoEx element later on. That's why the test fails.

KarlStraussberger:
It may depend on you client whether it handles CaptionInfo. You can try with Samsung flag on for the client.

TVrijdag:
Thank you for your feedback! 

So if I understood it correctly it means that `CaptionInfoEx` is a proprietary element, right?

I've just tested with the Samsung client flag. The server added the `CaptionInfo.sec` response header, but the external resource didn't work for the TV.

I noticed that the added resource has an invalid mimetype within protocolInfo ([source](https://github.com/gerbera/gerbera/blob/cf001f38b911b6eecd33541aaa465037014a5b63/src/metadata/metacontent_handler.cc#L245)). So I did some more testing. I manually changed this within the db. I also added entries to mappings such that the mimetype could be mapped to a contenttype. This fixed the issue for VLC, but it unfortunately did not for the TV (LG webos). 

The final part of the protocolInfo is also modified before it is added to the xml. Once I disabled that for the specific content type ([here](https://github.com/gerbera/gerbera/blob/cc49b2b2323155a47b19726544ab6db4e766c7b9/src/upnp_xml.cc#L743)) it also worked for my TV! The final protocolInfo which did work is `http-get:*:text/srt:*`. However, I'm not sure whether an asterisk would be okay here, or whether it actually needs specific value pairs.

KarlStraussberger:
Sounds very promising. Subtitles did never really work on many clients. If you are able to scrape your changes into this PR, it would mean a vast improvement in this area.

KarlStraussberger:
@TVrijdag I incorporated some changes in #1218. Maybe you can tell us whether this is similar to the things you did to make it run on your TV.

TVrijdag:
@KarlStraussberger I see that the mentioned PR already addresses the mimeType issue I encountered within fillMetadata. :)

I'll checkout your PR for some testing too. I think I'll be able to test it later today. 

I expect it would still need some additional changes around line 757 for it to work in the TV.
https://github.com/gerbera/gerbera/blob/9a3ebc87974f02b9ba4a3117567263667ab6d43d/src/upnp_xml.cc#L757

As for this PR, I haven't had the time to incorporate changes to the PR yet

TVrijdag:
I've just tested with master and everything is working now without additional changes!

KarlStraussberger:
Thanks for reporting and testing 🥇

PR #1235, opened 2021-02-17
Update SPDLog and TAGLIB scripts to latest versions

### LINKED ISSUE #1232: Update library versions
Taglib 1.12 is [out](https://github.com/taglib/taglib/commits/master)

spdlog 1.8.2 is [available](https://github.com/gabime/spdlog/tags)

Do we want to update before 1.7?

KarlStraussberger:
closes #1232


whyman:
I think we can consider dropping the taglib one as 1.11 is pretty much everywhere now, and its not used by docker or CI etc.

For the spdlog we need to consider making sure a new-enough fmt is also available.

Issue #1012, opened 2020-10-21
Web configurator for import filter


EricdeWaal:
Excited to see my previous idea #982 has been implemented! Can hardly wait for the next tag to see it in operation.
Time for a new idea: leverage one of the great strenghts of Gerbera, it's unique import filter. 
If you have music from over 325 artists (like me), a simple artist/album/song scheme will not be very handy if you control your Pioneer N30 network player by the Hi-Fi Cast app (or similar) on a mobile, let alone the tiny screen on the player itself.
I set mine up to do initial_letter/artist/album/song. Excellent but not so easy to do for a novice.

So the idea is to make the import script configuarable through the web interface, with blocks like artist, album, song, year, etc that can be selected or not, reordered and so on to build the import script.
IMHO that would be enough to make Gerbera a lot less nerdy, much more mainstream!

KarlStraussberger:
The idea is really great, but there is no way to import configuration into javascript. There is already am import script that does exactly what you want. I, myself, implemented it already for mediatomb.

EricdeWaal:
I know there is an abc option and I'm using it myself already from the mediatomb days. And agree it may not be easy to edit the javascript import files from javascript in a browser!
Would it be an option to have some pre-configured options (like abc on or off) that the user can select from the web ui after installation?

PR #1238, opened 2021-02-18
Add more config properties to and map genre

KarlStraussberger:
This allows the use of `config['/import/autoscan/directory'][0].location` in scripts


whyman:
Whats the use-case for this? Letting the script know which autoscan a file is from?

KarlStraussberger:
> 
> 
> Whats the use-case for this? Letting the script know which autoscan a file is from?

Exactly. Having all the configuration available in scripts allows us to influence the behaviour of the script. I've already got a couple of ideas, like replacing the rootpath, mapping the genres (not in config yet). In the end the js-layout can be adjusted without modifying the script - a little bit of #1012 .

whyman:
I like the idea.

I am wondering if it would be easier for the script user to make an object available with the details of only the **current** autoscan, if any?

I cant think of case where you would want to access other autoscan details than the one that is relevant for the file being processed? So currently you have to loop and match based on path for no reason?

Feel free to stay this is nonsense :D

KarlStraussberger:
I've been considering this, but the whole config is passed on startup, while the current autoscan has to be passed each time the script is called.

The solution can be to pass the index to the current autoscan config entry as global with each call (which is quite fast) and buy the extra memory for the other unused entries. Also, this enables scripting of scenarios we cannot make up at the moment.

KarlStraussberger:
Added mapping for genres and switch to use structured layout. Config may look like

```
<virtual-layout type="js" audio-layout="Strucured">
  <genre-map>
    <genre from="Rock|Alternative|Punk" to="Rock"/>
    <genre from="Dance|Rap|Hip|Progressive Electronic|Urban Crossover" to="Dance"/>
    <genre from="Audio Book|Stories" to="Audio Book"/>
    <genre from="Metal|Trance|Hard Rock|Quirk" to="Metal"/>
    <genre from="Comedy" to="Comedy"/>
    <genre from="Classical" to="Classical"/>
    <genre from="Blues|R.B" to="Blues"/>
    <genre from="Soundtrack|Film|Movie" to="Soundtrack"/>
    <genre from="Pop|A Capella|Folk|Sports|Data|Disco|Diva|Electronica|World|Glam|New Age|Romantic|Vocals|European|Classic Prog|Urban AC" to="Pop"/>
  </genre-map>
</virtual-layout>
```

KarlStraussberger:
The `script-options` can be used to set any option required in a script.

KarlStraussberger:
It finally fixes #1012 and allows us to add more control on scripts.

config/config.xsd
```diff
@@ -345,11 +347,45 @@
                     </xs:restriction>
                 </xs:simpleType>
             </xs:attribute>
+            <xs:attribute name="audio-layout" type="string"/>
```
whyman, line 350: What are these for?

whyman (commented):
LGTM mostly

config/config.xsd
KarlStraussberger: They are used in `import.js` to switch to e.g. structured layout. So there is only one import script which selects the layout function instead of replacing the import file. Obviously I forgot to add documentation for it. Currently only `audio-layout` is referenced, but I added the others to make things symmetric.

KarlStraussberger: See scripts/js/import.js

whyman:
Is this ready to go?

A thought I had in the shower this morning, hows does this code interact with autoscans added dynamically?

KarlStraussberger:
> 
> 
> Is this ready to go?
> 
> A thought I had in the shower this morning, hows does this code interact with autoscans added dynamically?

Good point. That's why I changed to dictionary instead of array. Also it gets the list from content manager. Should be fine.

KarlStraussberger:
It's ready - for the moment (as always)

PR #1245, opened 2021-02-23
Reload scripts after adding new autoscan via web ui

KarlStraussberger:


KarlStraussberger:
been a bit too slow for #1238 ...

PR #1246, opened 2021-02-24
use more CTAD

neheb:
Happens to avoid extra {}.

### DIFF
--- a/src/content/scripting/script.cc
+++ b/src/content/scripting/script.cc
@@ -54,17 +54,17 @@
 #include "content/onlineservice/atrailers_content_handler.h"
 #endif
 
-static constexpr std::array<duk_function_list_entry, 9> js_global_functions = { {
-    { "print", js_print, DUK_VARARGS },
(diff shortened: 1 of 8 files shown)


neheb:
Also did conversions to string_view since it seems to work better in constexpr context. Well, maybe future PR.

neheb:
Compiles now.

KarlStraussberger:
I don't see the benefit of this change. Speed? Size? The code looks more complicated as before.

neheb:
No speed or size difference. It's all done in constexpr context, so it generates the same assembly. Main benefit is to get rid of the explicit size parameter. CTAD does not work for just one parameter.

There was an experiment I did once where I imported a constexpr version of map (without insertion of course) and used that instead of these arrays. That one resulted in a slightly smaller binary. I have no idea why.

whyman:
Im not sure this is any improvement, so in this scenario I think its always best to stick with what is proven.

PR #1252, opened 2021-02-25
Enable configuration of system directories (hidden in web ui)

### LINKED ISSUE #1227: Cannot get to media directories from gerbera.html
Fedora 33

There is no way to get to an external hard drive to add files via /run/media/[user]/[resource]. Nor can I get to subdirectories from home.

KarlStraussberger:
fixes #1227


KarlStraussberger:
I don't like things being hard-coded

KarlStraussberger:
Have been looking at all the array stuff in upnp_xml, but that's enough material for another PR (will e..g. allow us to add further UPnP tags to containers at runtime).

PR #1255, opened 2021-02-26
Detect mimetype from extension

### LINKED ISSUE #1253: Extension-Mimetype-Preference broken?
When running a full scan I get lots of `Unable to handle mime type application/octet-stream`.

The message is new, but i want to ensure that there is no other reason, because there also was a change in mime code.

KarlStraussberger:
fixes #1253


whyman:
Oof thats a beefy regression. We definitely need more tests.

PR #1261, opened 2021-02-27
fix ConfigGeneratorTest

KarlStraussberger:


KarlStraussberger:
That should do the trick.

Been a lot less than expected.

PR #1273, opened 2021-03-02
include <optional> header

kraj:
Fixes build with gcc11
/src/util/tools.h:165:6: error: 'optional' in na
mespace 'std' does not name a template type
|   165 | std::optional<std::vector<std::byte>> readBinaryFile(const fs::path& path);
|       |      ^~~~~~~~

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/util/tools.h
+++ b/src/util/tools.h
@@ -34,6 +34,7 @@
 #include <filesystem>
 #include <map>
 #include <memory>
+#include <optional>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


neheb:
It's failing because headers must be alphabetical.

neheb (approved):

whyman:
Thaks @kraj!

PR #1274, opened 2021-03-02
Add UPnP tags to configuration

KarlStraussberger:
Configuration of dictionary defaults

Add metadata for containers to config

Allow configuration of additional metadata for tracks


whyman:
Code looks good :+1: 

Why would you want to change this configuration though?

whyman (approved):

KarlStraussberger:
Maybe you remember the discussion we recently had. Which property for album should be set - based on the standard and what clients digest.

Now we have the means to allow everything: by setting container metadata in script and enabling this in upup-properties, any detail can be send to the client without changing our code - only js and config.

PR #1277, opened 2021-03-03
Taglib: Extract from mp3 with ID3v1 tags, improve messaging

### LINKED ISSUE #1271: TagLibHandler messages are very misleading
This bit of code in src/metadata/taglib_handler.cc line 415 in v1.7.1 is indicative of the issue:
```
    if (!mp3.isValid() || !mp3.hasID3v2Tag()) {
        log_info("TagLibHandler {}: could not open mp3 file",
            item->getLocation().c_str());
        return;
    }
```
Old mp3s which are valid but not containing modern tags will clog the logs with an incorrect message that they could not be opened, which sounds like a permission or other file access problem.  It would be much friendlier, if "info" messages are to be given, to produce proper info - "Invalid mp3 file" and "contains no ID3v2Tag" for example

This same misleading idiom can be found for most file types

whyman:
Fixes #1271


whorfin:
🖤

PR #1278, opened 2021-03-03
DB: Transactions for inserts/updates

whyman:
- Convert exec to take a string
- Set a minium sqlite version to 3.7.0
- Use SQLite Write Ahead Log mode (https://sqlite.org/wal.html)
- More detailed log timestamps in debug mode


KarlStraussberger:
@whyman ready to merge?

whyman:
Erm, probably? Probably needs a bit of testing to make sure, works fine on my machine. Would be good to hear from @whorfin if it makes any difference.

PR #1279, opened 2021-03-04
Ensure sqlite is in correct mode for recovery and initialization

### LINKED ISSUE #1272: Unfortunate, almost unusable performance regression in v1.6.1 to 1.7.0
I have a relatively large collection of audio files, approx 250K of them
The database for v1.6.1 was a little over 1G
Gerbera 1.7.0 successfully "migrated" the database, but has been consuming 100% of one core and occasionally emitting messages about "Taglib: can't open mp3 file" for untagged mp3 files.  These messages haven't yet repeated, and seem to indicate some very unperformant passing over every single file, smelling like an exponential or other bad-big-O algorithm.
For three days and counting it's been doing this; the database file is very very slowly growing, like a few kilobytes every ten seconds
I'm about to give up, but figured I'd see if it ever finishes.
Is this expected?  This isn't even a "Rescan from scratch" situation, it is a post-migration.

I can not yet tell if there is an additional regression or the massive CPU usage and a threading problem causing query latency, but in addition, at least at the moment, this behavior completely breaks `foobar2000` on iOS.  When top-level `Albums` or `PC Directory` queries are made, `8player` on iOS seems to do them in chunks.  `foobar2000` seems to request "all the things".  Gerbera, in turn, seems to never return "all the things", either because it's stuck because some re-indexing/updating process is still happening and it never finishes the query, or because it's too damn slow - i've given up watching `foobar2000` be stuck on "Please wait..."

KarlStraussberger:
fixes #1272: This only happened when you first created the database or after recovery.

### DIFF
--- a/src/database/sqlite3/sqlite_database.cc
+++ b/src/database/sqlite3/sqlite_database.cc
@@ -129,6 +129,14 @@ Sqlite3Database::Sqlite3Database(std::shared_ptr<Config> config, std::shared_ptr
     hasBackupTimer = false;
 }
 
+void Sqlite3Database::prepare()
+{
+    _exec("PRAGMA locking_mode = EXCLUSIVE");
+    _exec("PRAGMA foreign_keys = ON");
+    _exec("PRAGMA journal_mode = WAL;");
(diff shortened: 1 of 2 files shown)


whyman:
Looks sensible

PR #1282, opened 2021-03-06
Add cache for containers to improve import speed

KarlStraussberger:
contributes in my case another 15% to #1272


whyman:
Looks sensible, so lets merge it. 

We should work on test coverage at some point because cache invalidation is tricky!

PR #1284, opened 2021-03-07
Misc fixes

KarlStraussberger:
Fix compliation for Ubuntu:18.04 and debian:stable. Somehow it worked before the merge 😟 

Also fix message for ui when virtual url already contains protocol.

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -281,7 +281,9 @@ void ContentManager::timerNotify(std::shared_ptr<Timer::Parameter> parameter)
 
     if (parameter->whoami() == Timer::Parameter::IDAutoscan) {
         std::shared_ptr<AutoscanDirectory> adir = autoscan_timed->get(parameter->getID());
-        if (adir == nullptr)
+
+        // do not rescan while other scans are still active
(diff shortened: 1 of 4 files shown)


src/content/content_manager.cc
```diff
@@ -281,7 +281,9 @@ void ContentManager::timerNotify(std::shared_ptr<Timer::Parameter> parameter)
 
     if (parameter->whoami() == Timer::Parameter::IDAutoscan) {
         std::shared_ptr<AutoscanDirectory> adir = autoscan_timed->get(parameter->getID());
-        if (adir == nullptr)
+
+        // do not rescan while other scans are still active
+        if (adir == nullptr || adir->getActiveScanCount() > 0 || adir->getTaskCount() > 0)
```
KarlStraussberger, line 286: May this be the root cause of #1272 ?

src/server.cc
```diff
@@ -219,8 +219,11 @@ void Server::run()
     if (url.empty()) {
         url = renderWebUri(ip, port);
     }
+    if (url.substr(0, 4) != "http") {
```
whyman, line 222: find()?

PR #1285, opened 2021-03-07
Mime fix

KarlStraussberger:
nested call leads to checking magic each time

### DIFF
--- a/src/config/config_setup.cc
+++ b/src/config/config_setup.cc
@@ -606,7 +606,7 @@ std::vector<std::string> ConfigArraySetup::getXmlContent(const pugi::xml_node& o
         }
     }
     if (result.empty()) {
-        log_debug("{} assinging {} default values", xpath, defaultEntries.size());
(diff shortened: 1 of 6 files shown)


KarlStraussberger:
@whorfin tends to reduce scanning time on my side to 50% (after 10 minutes already 1/6 of the database size)

whorfin:
@KarlStraussberger I still see timestamps indicating 5-7 sec per audio file, although general startup and getting to where audio file scanning is happening does seem considerably snappier
I've patched by hand, so once there's overall integrated changes and/or version bump i can try again

KarlStraussberger:
> 
> 
> @whorfin tends to reduce scanning time on my side to 50% (after 10 minutes already 1/6 of the database size)

Well it's an improvement but images have fewer data but take some time. It only another 5%.

whyman:
Heh, I had this change in a branch locally too :laughing:

Issue #1264, opened 2021-02-28
Ubuntu 20.04 package on Bintray seems incorrectly configured


lckarssen:
Let me start off by mentioning that this is the first time I try to install Gerbera, so I may easily be missing something. 

I tried to install Gerbera (1.7.0) from [the Ubuntu repository at Bintray](https://bintray.com/gerbera/gerbera), as mentioned in the documentation. After adding the repo I installed the package and its dependencies and I expected the service to have started (as is usual on Debian). This turned out not to be the case. 

I then read on in the documentation and came across the section about the config file. A `dpkg-query -L gerbera` showed me that no default `config.xml` was supplied be the package and that the `/etc/gerbera` required byt the daemon wasn't created either. I created the directory and the `config.xml` file, `chown`-ed both, changed the `<home>` entry to `/etc/gerbera` and expected all to be fine. Yet, a `systemctl start gerbera` followed by a `systemctl status gerbera` showed me that the service still didn't run. 

Next, I decided to run `gerbera` manually to see if there were any error messages: 
```
$ sudo -u gerbera gerbera -c /etc/gerbera/config.xml
2021-02-28 22:33:27   info: Gerbera UPnP Server 1.7.0 - http://gerbera.io/
2021-02-28 22:33:27   info: Copyright 2016-2021 Gerbera Contributors.
2021-02-28 22:33:27   info: Gerbera is free software, covered by the GNU General Public License version 2
2021-02-28 22:33:27   info: Loading configuration from: /etc/gerbera/config.xml
2021-02-28 22:33:27   info: Checking configuration...
2021-02-28 22:33:27  error: File '/usr/local/share/gerbera/sqlite3.sql' does not exist
```
As you can see it is looking for a file in `/usr/local/share/gerbera`, whereas the `dpkg-query` command I mentioned above only lists files in `/usr/share/gerbera` (so without the `local/`). This is also what one expects on Debian. I then checked the generated `config.xml` file and replaced all `/usr/local/` paths with `/usr/`, but still no luck. I guess the path to the `sqlite3.sql` file is hardcoded somewhere. 

Finally, I checked the systemd service description: 
```
$ cat /lib/systemd/system/gerbera.service 
[Unit]
Description=Gerbera Media Server with MySQL
After=mysql.target network.target

[Service]
Type=simple
User=gerbera
Group=gerbera
ExecStart=/usr/local/bin/gerbera -c /etc/gerbera/config.xml
Restart=on-failure
RestartSec=5

[...]

PR #1286, opened 2021-03-07
Do make a homedir for the gerbera user in Debian postinst

lckarssen:
This commit removes the '--no-create-homedir' argument from the
adduser command in the postinst script used in the Debian package.
Without an existing homedir the Gerbera service won't start.

This is part of the fix for #1264.


whyman:
Thanks @lckarssen! :tada:

Issue #848, opened 2020-05-06
Unrecognized action X_SetBookmark (aka make resume work)


kyak:
Hi!

I'm using latest gerbera git with Samsung UHD 4K 2019 model.

When stopping video, I see the following in the log file:

```
info: unrecognized action X_SetBookmark
error: ActionRequest::update(): response is nullptr, code -115 for X_SetBookmark
```

I assume that TV is trying to set a "bookmark" in order to resume playing from that moment later on.
However, resuming doesn't work.
What needs to be done in order to support resuming?

Resuming works perfectly when using minidlna with the same TV.

whyman:
Hi, thanks for reporting.

Looks like another Samsung extension, yay!

There is a post over on the Serviio forum where someone has worked out how it works by sniffing the Samsung server response: https://forum.serviio.org/viewtopic.php?t=272#p6504

whyman:
Now implemented thanks to @dyseg in #1288

aleho:
How are multiple flags specified in the config?

Or-ed like this? Or comma separated? Couldn't find the place where its parsed at first glance.
Maybe I'm too stupid or docs need an update too?

```XML
<clients enabled="yes">
    <client ip="192.168.1.69" flags="SAMSUNG|SAMSUNG_BOOKMARK_SEC"></client>
</clients>
```

kyak:
Nice, will try that, thanks!

dyseg:
@aleho You guessed it right.
https://docs.gerbera.io/en/latest/config-clients.html
>Containing the flags you want to set. Must be given in the following format SAMSUNG|0x100, where the text either contains one of the known flags or an integer number if the flags has no name. For valid flags see: `Supported Devices`.

They are parsed here: https://github.com/gerbera/gerbera/blob/bb924ba43d5ba52fc6781539b1b01708a50d8d69/src/config/config_setup.cc#L1422-L1424

whyman:
Note you will need to run a git build or 1.8.0 when it is released to use this feature :+1:

PR #1288, opened 2021-03-07
Handle Samsung X_SetBookmark action

dyseg:
I originally created these modifications only for myself (with my limited C++ skills). But since it is working, why not share it?
I stored the "bookmark position" in the existing mt_cds_object table in a new column. Maybe it is not ideal, considering not all TVs will support this, but the column will be there anyway. Also, I was only able to test with sqlite.
The implementation details are coming from the [Serviio blog post from 2011](https://github.com/gerbera/gerbera/issues/848#issuecomment-624485095).
All tests were done on UE50TU7102, which sends back the bookmark position as msec, although in he Serviio blog post the example shows that it is sent back as sec. Also it would be better to store the bookmark position on per client basis, maybe in a new SQL table.


whyman:
This is great, thanks @dyseg!

I think it would be ideal to wire this into our client detection "tweaks" mechanism (at least for sending the saved position), but otherwise it looks nice.

dyseg:
I've moved the X_SetBookMark action handling to the Quirks class. Please share your thoughts @whyman. Also the global config option to switch on/off the bookmark handing was removed, I think it's is not necessary any more, since we have to switch it on for every client explicitly anyway.

I would still like to store the bookmarks corresponding to single clients, I'll try to implement it as a next step.

dyseg:
I've added a new quirk option, since older Samsung TVs send back the bookmark position as second(like in the original Serviio forum post). This was verified with UE43J5500. Also from now on, in the database the bookmark value is always stored as 'second', and multiplied if necessary.

I think at this point You should merge this PR, and later on if I have time I'll finish the per client based bookmark storing. By the way, maybe it's not even necessary, I found it useful to stop on one TV and continue on the other :).

Also, please mark that this PR will resolve #848

KarlStraussberger:
There is a conflict in `src/upnp_xml.cc` which needs to be resolved before the PR can be merged

src/util/upnp_quirks.h
```diff
@@ -31,10 +31,15 @@
 #include <pugixml.hpp>
 namespace fs = std::filesystem;
 
+class ActionRequest;
```
whyman, line 34: These should go with the other forward declarations below

src/upnp_cds.h
```diff
@@ -95,10 +97,11 @@ class ContentDirectoryService {
     /// \brief Samsung Extension X_SetBookmark
     /// \param request Incoming ActionRequest.
     ///
-    static void doSamsungBookmark([[maybe_unused]] const std::unique_ptr<ActionRequest>& request);
+    void doSamsungBookmark(const std::unique_ptr<ActionRequest>& request);
 
     std::shared_ptr<Config> config;
     std::shared_ptr<Database> database;
+    std::shared_ptr<ContentManager> content;
```
whyman, line 104: I guess this is from the first version, looks like its not actually needed now?

src/upnp_xml.cc
```diff
@@ -105,6 +105,9 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
     if (obj->isItem()) {
         auto item = std::static_pointer_cast<CdsItem>(obj);
 
+        if (quirks != NULL)
```
whyman, line 108: Probably should be `nullptr`?

whyman (changes requested):
Looks nice!

src/util/upnp_quirks.h
dyseg: Resolved in https://github.com/gerbera/gerbera/pull/1288/commits/cb16aa7074d22b899544f986df6b1d0821b2e66f

src/upnp_cds.h
dyseg: Yes, you are right, this is not necessary any more. Reverted in: https://github.com/gerbera/gerbera/pull/1288/commits/fc6dc8807a616bbf64c663fc8a625cc62c0760d4

src/upnp_xml.cc
dyseg: Resolved in https://github.com/gerbera/gerbera/pull/1288/commits/e23aeff8f0cdb93e062b27881e206442902dbba2

whyman:
LGTM, happy to merge, but there are conflicts

dyseg:
> LGTM, happy to merge, but there are conflicts

Sorry but I don't see where. Locally auto merge is possible into gerbera:master, and Github also states the same. Am I missing something?

whyman:
We generally do rebase merges.

We can squash merge without conflicts if you prefer?

dyseg:
Squash merge is fine for me. Maybe it's even better to have all the changes in a single commit.

PR #1289, opened 2021-03-08
Warn if blocking-tcp is not disabled in libupnp

whyman:
If control points do not correctly unsubscribe this can result in
libupnp threads hanging.

https://github.com/pupnp/pupnp/discussions/245#discussioncomment-448988
https://github.com/pupnp/pupnp/commit/32cffb5bb55a650b1eb962c6fe2e58e6bf4fe2c5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,6 +353,13 @@ else()
     Without this option Gerbera will be unable to restart with the same port number.]=])
     endif()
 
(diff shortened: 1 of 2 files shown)


whyman:
I've also raised a PR upstream to enable the option by default: https://github.com/pupnp/pupnp/pull/246

KarlStraussberger:
This still leaves us with the problem that large folders are not returned properly to foobar et. al.

whyman:
Yeah, this is not related to that issue :grin:

PR #1292, opened 2021-03-09
Only count entries if container was just created

KarlStraussberger:
Should be 1 entry then :smile:

Reduced import time by another 10%. So we come closer fixing #1272


neheb:
I like all the fmt::format changes.

PR #1294, opened 2021-03-10
WIP rescanDirectory test

whyman:
It is not trivial to test this as everything is so tightly coupled...


whyman:
Need to do some fairly big refactoring before this can make sense :disappointed:

PR #1312, opened 2021-03-14
Change C file handling to fs:directory_entry

KarlStraussberger:


src/content/content_manager.cc
```diff
@@ -619,13 +624,12 @@ void ContentManager::_rescanDirectory(std::shared_ptr<AutoscanDirectory>& adir,
...
         return;
-        //        continue;
-        // throw_std_runtime_error("Container has no location information");
     }
 
-    DIR* dir = opendir(location.c_str());
-    if (!dir) {
-        log_warning("Could not open {}: {}", location.c_str(), std::strerror(errno));
+    std::error_code ec;
+    auto dirEnt = fs::directory_entry(location);
```
whyman, line 630: The constructor thows on error, I guess passing the ec and checking it would help.

src/content/content_manager.cc
```diff
@@ -659,9 +663,8 @@ void ContentManager::_rescanDirectory(std::shared_ptr<AutoscanDirectory>& adir,
     time_t last_modified_new_max = last_modified_current_max;
     adir->setCurrentLMT(location, 0);
 
-    struct dirent* dent;
-    while ((dent = readdir(dir)) != nullptr) {
-        char* name = dent->d_name;
+    for (const auto& dirEnt : fs::directory_iterator(location, fs::directory_options::skip_permission_denied, ec)) {
```
whyman, line 666: dirEnt is redefined here, I wonder if we can give it a more clear name? or change the other one?

whyman (approved):
Nice! I had a the same thoughts :)

src/content/content_manager.cc
KarlStraussberger: Checking of entry is done already, so I don't check the constructor return separately

PR #1313, opened 2021-03-15
Add Directories container to Audio in builtin layout (#1310).

### LINKED ISSUE #1310: Add Directories container to Audio in builtin layout
I would like to see a Directories container in Audio when using builtin layout, just like in Videos and Photos. What do you think?
Would you accept such a patch?
Yes, i know i can just use js layout and do this via javascript, but this ist damn slow compared to builtin layout.

phi-whiterabbit:
This fixes #1310, adding a Directories container under Audio.


whyman:
Thanks!

PR #1315, opened 2021-03-16
Log message from directory_entry constructor

KarlStraussberger:
All errors are handled but the messages are misleading


src/content/content_manager.cc
```diff
@@ -406,7 +406,7 @@ void ContentManager::addVirtualItem(const std::shared_ptr<CdsObject>& obj, bool
 
     std::error_code ec;
     auto dirEnt = fs::directory_entry(path, ec);
-    if (!dirEnt.is_regular_file(ec))
+    if (ec.value() || !dirEnt.is_regular_file(ec))
```
whyman, line 409: You can just check `ec` as it has a boolean overload for non-zero :+1:

KarlStraussberger: Thanks!

PR #1318, opened 2021-03-16
Fix crash in filetoMimeType and remove ancient Solaris hack

whorfin:
13 years ago in the original mediatomb, a hack was done to allow printf() of a NULL pointer to print "NULL" rather than crash, on Solaris
That was done by mapping *0=0 behind the scenes, which makes a null dereference no longer an immediate crash, which it otherwise is on Solaroid machines, and quite deliberately
No, just, no - a nullptr dereference is a bug
Removed
And then I rather quickly found a crash which was hiding behind a totally messed up stack otherwise
Mime::fileToMimeType was blindly dereferencing a nullptr to see if it pointed to an empty string

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -126,19 +126,6 @@ static void installSignalHandler()
 
 int main(int argc, char** argv, char** envp)
 {
-#ifdef SOLARIS
-    std::string ld_preload;
-    char* preload = getenv("LD_PRELOAD");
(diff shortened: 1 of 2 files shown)


whyman:
> That was done by mapping *0=0 behind the scenes, which makes a null dereference no longer an immediate crash, which it otherwise is on Solaroid machines, and quite deliberately

PR #1326, opened 2021-03-17
Improve error messages when directory_entry causes an error

KarlStraussberger:


src/content/content_manager.cc
```diff
@@ -670,6 +670,11 @@ void ContentManager::_rescanDirectory(std::shared_ptr<AutoscanDirectory>& adir,
     adir->setCurrentLMT(location, 0);
 
     for (const auto& dirEnt : fs::directory_iterator(location, fs::directory_options::skip_permission_denied, ec)) {
+        fs::path newPath = dirEnt.path();
```
whyman, line 673: Save an assignment in the failure case by moving this after the check :grin:

whyman (approved):

src/content/content_manager.cc
KarlStraussberger: It's needed in the message and therefore only the property is wasted in the error case - which we still assume is not the standard 😺

PR #1328, opened 2021-03-18
Fix dirent iteration in _rescanDirectory and addRecursive

whorfin:
There were a few problems here...
1. skip_permission_denied causes the iterator to quit as soon as it hits a permissions denied error; it is arguably a poorly named option.  We don't want this behavior, so we just let the existing code handle permissions denied but continue the dirent iteration.
```
 if std::filesystem::directory_options::skip_permission_denied is set in options and construction encounters a permissions denied error, constructs the end iterator and does not report an error.
```
2. the ec from the iterator is checked inside the loop, and then the loop skipped if set; the iterator ec is only set if the iterator construction failed, in which case the for loop would never happen.  ec has nothing to do with iteration once in process
```
Constructs a directory iterator that refers to the first directory entry of a directory identified by p. If p refers to an non-existing file or not a directory, returns the end iterator and sets ec.
```

That's how I read the standard at any rate; these changes cause the code to no longer malfunction on my directory tree.

The incredibly unhelpful "Invalid argument" error is still returned instead of Permission denied, which seems to be some allowed semantic horribleness - calling an extant directory with denied permissions a non-existent entity is an odd behavior from an old-timer perspective, but there ya go.

Also note the caching behavior which presumably motivated these changes is only a "suggestion" not a guaranteed behavior according to the standard.

### DIFF
--- a/src/content/autoscan_inotify.cc
+++ b/src/content/autoscan_inotify.cc
@@ -466,31 +466,27 @@ void AutoscanInotify::monitorUnmonitorRecursive(const fs::directory_entry& start
         log_warning("Could not open {}: {}", startPath.path().c_str(), ec.message());
         return;
     }
-
-    for (const auto& dirEnt : fs::directory_iterator(startPath, fs::directory_options::skip_permission_denied, ec)) {
+    auto dIter = fs::directory_iterator(startPath, ec);
+    if (ec) {
+        log_error("monitorUnmonitorRecursive: Failed to iterate {}, {}", startPath.path().c_str(), ec.message());
+        return;
+    }
+    for (const auto& dirEnt : dIter) {
         if (shutdownFlag)
             break;
-        auto name = dirEnt.path().c_str();
-        if (name[0] == '.') {
-            if (name[1] == 0)
-                continue;
-            if (name[1] == '.' && name[2] == 0)
-                continue;
-        }
-        if (!followSymlinks && dirEnt.is_symlink()) {
+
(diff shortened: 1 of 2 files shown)


whorfin:
I have no clue what is up with the branch conflict thrown above; if you look at the changes in the PR it should be very clear

neheb:
Rebase against current master.

```
curl https://patch-diff.githubusercontent.com/raw/gerbera/gerbera/pull/1328.patch | git am
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1834    0  1834    0     0   4801      0 --:--:-- --:--:-- --:--:--  4813
Applying: Fix dirent iteration
error: patch failed: src/content/content_manager.cc:669
error: src/content/content_manager.cc: patch does not apply
Patch failed at 0001 Fix dirent iteration
hint: Use 'git am --show-current-patch=diff' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
```

KarlStraussberger:
Thank you for the PR. You happen to be some three hours late. I cleaned it up already and it now does exactly what you intend it to do. Originally, I was in hope of avoiding exceptions and error messages, which did not work out.

whorfin:
Hah, classic race condition...
🧐

PR #1331, opened 2021-03-19
Encapsulate thread handling in own class

KarlStraussberger:
Removes several copies of similar code

I'd also like to move the lock and mutex handling to that class so we can control multithreading by configuration.


phi-whiterabbit:
If the idea is to refactor threading, shouldn't it then be using stuff from https://en.cppreference.com/w/cpp/thread instead of the pthread things which are much more C than C++ like?

KarlStraussberger:
Thanks for the hint. I was looking at it already, but had to carve out pthread in the first place.

neheb:
using std::thread instead of pthread would also get rid of the CMake pthread stuff.

KarlStraussberger:
I stumbled upon this while looking for `pthread_`. It was just the same code with different messages.

KarlStraussberger:
@whyman Do you think it is sensible to even try to remove pthread (the mysql code looks bound to it)

whyman:
`std::thread` is implemented with `pthreads`, so it will need to stay in terms of CMake.

That said we can use the `std::thread` API instead of pthread directly as its probably nicer.

neheb:
> `std::thread` is implemented with `pthreads`, so it will need to stay in terms of CMake.
> 
> That said we can use the `std::thread` API instead of pthread directly as its probably nicer.

AFAIK libc++/std++ links to pthread statically so that's a moot point.

whyman:
> > `std::thread` is implemented with `pthreads`, so it will need to stay in terms of CMake.
> > That said we can use the `std::thread` API instead of pthread directly as its probably nicer.
> 
> AFAIK libc++/std++ links to pthread statically so that's a moot point.

You need to pass at least `-pthread` or whatever other stuff for other OS, this is why `Threads::Threads` is in CMake.

https://stackoverflow.com/a/62093937

neheb:
Weird. I would think it's an implementation deal with some libc++

KarlStraussberger:
From the above said, I don't see a point in introducing std::thread.

whyman:
LGTM, much tidier, and better logging, but I admit my C/C++ concurrency primitives knowledge is not as great as it should be.

I have a couple of questions:
1. If you return a `lock_guard`, what is the behaviour there? Pretty sure they cant be copied.
1. Can we move from passing config object to passing the value of `CFG_THREAD_SCOPE_SYSTEM` directly? In the future it will be much easier to unittest
2. When I was reading up on this stuff after for the review it seems that Linux only supports `PTHREAD_SCOPE_SYSTEM` anyway  :shrug:

KarlStraussberger:
> LGTM, much tidier, and better logging, but I admit my C/C++ concurrency primitives knowledge is not as great as it should be.
> 
That's been my objective and I addressed the stuff from the same practical point of view: if it works, it works

>     1. If you return a `lock_guard`, what is the behaviour there? Pretty sure they cant be copied.
It can't be copied, it can't be assigned so I assume the code is inlined - It works 🙈 

>     2. Can we move from passing config object to passing the value of `CFG_THREAD_SCOPE_SYSTEM` directly? In the future it will be much easier to unittest
I had some more options in mind, like priorities and other pthread_attribute properties for particular threads (they have names now). That's why I went for the whole `config`.

>     3. When I was reading up on this stuff after for the review it seems that Linux only supports `PTHREAD_SCOPE_SYSTEM` anyway  🤷
I was looking at: https://man7.org/linux/man-pages/man3/pthread_attr_setscope.3.html and decided to override the default `PTHREAD_SCOPE_PROCESS`. With that I've seen CPU-usage over 100%!

phi-whiterabbit:
> > LGTM, much tidier, and better logging, but I admit my C/C++ concurrency primitives knowledge is not as great as it should be.
> 
> That's been my objective and I addressed the stuff from the same practical point of view: if it works, it works
> 
> > ```
> > 1. If you return a `lock_guard`, what is the behaviour there? Pretty sure they cant be copied.
> > ```
> 
> It can't be copied, it can't be assigned so I assume the code is inlined - It works 🙈
> 

Well, it has no copy constructor, no move constructor and cannot be copy- or move assigned.
The reason why this works at all is that copy elision (which happens here) is mandatory since C++17:
https://en.cppreference.com/w/cpp/language/copy_elision

KarlStraussberger:
> 
> 
> > > LGTM, much tidier, and better logging, but I admit my C/C++ concurrency primitives knowledge is not as great as it should be.
> > 
> > 
> > That's been my objective and I addressed the stuff from the same practical point of view: if it works, it works
> > > ```
> > > 1. If you return a `lock_guard`, what is the behaviour there? Pretty sure they cant be copied.
> > > ```
> > 
> > 
> > It can't be copied, it can't be assigned so I assume the code is inlined - It works 🙈
> 
> Well, it has no copy constructor, no move constructor and cannot be copy- or move assigned.
> The reason why this works at all is that copy elision (which happens here) is mandatory since C++17:
> https://en.cppreference.com/w/cpp/language/copy_elision

We require C++17, so we're save here and the code can be kept nice and clear 👏 Thank you for teaching an old C++ guy.

whyman:
I think to reduce the WTF-per-minute we should probably stick to keeping the AutoLocks in the blocks themselves?

KarlStraussberger:
The point is the `mutex` which I prefer being hidden inside the class. 
The easy way out can be to add exactly the link to cppreference as comment.

PR #1336, opened 2021-03-21
fix MySQL initial setup

dyseg:
Error during the very first startup when the DB tables should be created:
>Mysql: error while creating db: mysql_error (1065): "Query was empty"

The problem is that after the mysql.sql file content is split by `';'`, during the last loop the `statement` variable will only contain a newline character, which causes the mentioned MySQL error.

https://github.com/gerbera/gerbera/blob/d6e57d1211a813941ef23978f7a6a65c0b8a7cfb/src/database/mysql/mysql_database.cc#L212


KarlStraussberger:
Thank you.

PR #1339, opened 2021-03-24
fs::path conversions

neheb:
there are several places in the code that implicitly convert an
fs::path to an std::string. This breaks on platforms like Windows
where the paths are wstrings. Remove some of these implicit
conversions and make others explicit.

### DIFF
--- a/src/cds_objects.h
+++ b/src/cds_objects.h
@@ -440,7 +440,7 @@ public:
     void setURL(const std::string& URL) { this->location = URL; }
 
     /// \brief Get the URL of the item.
-    std::string getURL() const { return location; }
+    std::string getURL() const { return location.string(); }
     /// \brief Copies all object properties to another object.
     /// \param obj target object (clone)
     //void copyTo(std::shared_ptr<CdsObject> obj) override;
--- a/src/config/config_manager.cc
+++ b/src/config/config_manager.cc
@@ -1251,9 +1251,10 @@ void ConfigManager::load(const fs::path& userHome)
 
     temp = setOption(root, CFG_SERVER_APPEND_PRESENTATION_URL_TO)->getOption();
     if (((temp == "ip") || (temp == "port")) && getOption(CFG_SERVER_PRESENTATION_URL).empty()) {
-        throw std::runtime_error("Error in config file: \"append-to\" attribute "
-                                 "value in <presentationURL> tag is set to \""
-            + temp + "\" but no URL is specified");
+        throw_std_runtime_error("Error in config file: \"append-to\" attribute "
+                                "value in <presentationURL> tag is set to \"{}\""
+                                "but no URL is specified",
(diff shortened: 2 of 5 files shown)


src/cds_objects.h
```diff
@@ -440,7 +440,7 @@ class CdsItemExternalURL : public CdsItem {
     void setURL(const std::string& URL) { this->location = URL; }
 
     /// \brief Get the URL of the item.
-    std::string getURL() const { return location; }
+    fs::path getURL() const { return location; }
```
whyman, line 443: URL is not a path

src/config/config_manager.cc
```diff
@@ -1007,14 +1007,14 @@ void ConfigManager::addOption(config_option_t option, std::shared_ptr<ConfigOpti
 
 void ConfigManager::load(const fs::path& userHome)
 {
-    std::string temp;
+    fs::path temp;
```
KarlStraussberger, line 1010: `temp` is reused in several places, some as string some as path. You need to introduce a separate variable `tempPath`.

src/config/config_manager.cc
```diff
@@ -1007,14 +1007,14 @@ void ConfigManager::addOption(config_option_t option, std::shared_ptr<ConfigOpti
...
     pugi::xml_node tmpEl;
 
     std::map<std::string, std::string> args;
     auto self = getSelf();
     std::shared_ptr<ConfigSetup> co;
 
-    log_info("Loading configuration from: {}", filename.c_str());
+    log_info("Loading configuration from: {}", filename.string().c_str());
     pugi::xml_parse_result result = xmlDoc->load_file(filename.c_str());
```
KarlStraussberger, line 1018: `fs::path` has a `c_str()` method https://en.cppreference.com/w/cpp/filesystem/path/native. No need to do two steps

src/config/config_manager.cc
```diff
@@ -1253,7 +1253,7 @@ void ConfigManager::load(const fs::path& userHome)
     if (((temp == "ip") || (temp == "port")) && getOption(CFG_SERVER_PRESENTATION_URL).empty()) {
         throw std::runtime_error("Error in config file: \"append-to\" attribute "
                                  "value in <presentationURL> tag is set to \""
-            + temp + "\" but no URL is specified");
+            + temp.string() + "\" but no URL is specified");
     }
```
KarlStraussberger, line 1257: You'd better change that to a `fmt::format` string

KarlStraussberger (changes requested):

src/config/config_manager.cc
neheb: Not quite. c_str() can be a wchar_t.

src/config/config_manager.cc
neheb: agreed.

PR #1344, opened 2021-03-26
Remove direct resource file handling

KarlStraussberger:
Since 1.7.0 images, subtiltes etc. are added as resources.
The old way to search for files is hidden in IFDEF before removing it completely


whyman:
I would just get rid of it, we can always revert the commit if needed.

PR #1349, opened 2021-03-27
Add cmdline options -u, -d and -P

flya320:
-u,user: drop to user USER after startup, also changing all groups to those of USER
-d,daemon: daemonize after startup
-P,pidfile: write a pidfile


flya320:
As far as I can see, the build checks are failing due to the different output of gerbera -h due to the added options.

flya320:
regarding the lint-clang-format check I would like to keep my formatting, since I find it more readable

whyman:
Hey, thanks for the PR :+1: 

The failing test you can fix by updating the content of `fixtures/mock-help.out` to match the new output.
The clang-format I agree makes it marginally less readable, but it also makes it consistent with the rest of the project.

flya320:
As I have mo Mac available perhaps someone could check if all uids, all gids are being changed to the requested user and additional groups are being set. I think they should if I understand the mac os manpages correctly.

whyman:
Looks good!

The pidfile write can be simplfied using the fmtlib library:
```c++
#include <fmt/os.h>

int main() {
  auto out = fmt::output_file("guide.txt");
  out.print("Don't {}", "Panic");
}
```

src/main.cc
```diff
@@ -214,6 +231,145 @@ int main(int argc, char** argv, char** envp)
...
+            log_info("Daemonized.");
+        }
+
+        // are we requested to write a pidfile ?
+        std::optional<std::string> pidfile;
+        if (opts.count("pidfile") > 0) {
+            pidfile = opts["pidfile"].as<std::string>();
+
+            // exit if the pidfile already exists
+            struct stat* s = NULL;
```
KarlStraussberger, line 331: You'd better make it an object `struct stat statbuf;` and call `stat()` with `&statbuf`

src/main.cc
```diff
@@ -214,6 +231,145 @@ int main(int argc, char** argv, char** envp)
...
+            if (0 > snprintf(pidstr, 18, "%d", pid)) {
+                log_error("Could not determine pid of running process.");
+                exit(EXIT_FAILURE);
+            }
+
+            // add a newline
+            strcat(pidstr, "\n");
+
+            // open the pidfile
+            int pidfd = open(pidfile->c_str(), O_CREAT | O_WRONLY, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
```
KarlStraussberger, line 356: There is a function `writeTextFile` in `util/tools` which does the file work for you.

KarlStraussberger (changes requested):

src/main.cc
flya320: Good point, thanks!

src/main.cc
flya320: I will look into it within the next week.

whyman:
Can we get this rebased @flya320? Then happy to merge it, we can fix it up once its in with further PRs. 

Otherwise keeping up with master branch changes is going to be annoying :smile:

flya320:
Sorry to say, what does rebase mean? What do I have to do?

thorsten

whyman:
Your fork was taken at some time in the past, and the main branch has moved on since then (so we now have conflicts).

Rebasing is the process of updating your branch to be taken from the latest code (moving the "base")

The high level steps are:
1) Pull the latest master code
2) checkout your branch
3) git rebase master
4) Push any conflicts
5) git rebase --continue
6) git push --force-with-lease

whyman:
Though, if you prefer, we can "squash" these commits into one and merge it without trouble.

flya320:
What is easier for you?

whyman:
Squashing is easiest as I can just press this green button

flya320:
Lets do squash then! Sorry that I do not have a clue about the inner workings of github.

Thorsten

whyman:
@flya320 if you are feeling generous a patch for the documentation (/docs) would be great :smile:

flya320:
Yes, I will add a doc patch. May take a few days though...

flya320:
Do you only need a patch for the website docs or is the manpage also kept up to date?

PR #1350, opened 2021-03-29
Reverse search clients by agent to prefer configured items

### LINKED ISSUE #1346: Matching Samsung client by userAgent doesn't work
I'm trying to match a client by user agent like this:

`<client userAgent="SEC_HHP_[TV]" flags="SAMSUNG|SAMSUNG_BOOKMARK_SEC"/>`

Unfortunately the already existing match pattern `SEC_HHP_` will always override it.
If I'm specifying an ip the client works.

I also tried `SEC_HHP_` but that doesn't override it either.

| IP Address | Host Name | Profile | User Agent | Match Type | Match Pattern | Client Type | First Seen | Last Seen | Quirk Flags |
| ---------- | --------- | ------- | ---------- | ---------- | ------------- | ----------- | ---------- | --------- | ----------- |
| 10.0.0.60 | samsung-tv.lan | Manual Setup for IP 10.0.0.60 | SEC_HHP_[TV]WZ/1.0 DLNADOC/1.50 | IP | 10.0.0.60 | None | Fri Mar 26 22:35:33 2021 | Fri Mar 26 22:35:33 2021 | SAMSUNG\|SAMSUNG_BOOKMARK_SEC

KarlStraussberger:
fixes #1346

### DIFF
--- a/src/util/upnp_clients.cc
+++ b/src/util/upnp_clients.cc
@@ -33,7 +33,7 @@
 
 #include <upnp.h>
 
-// table of supported clients (sequence of entries matters!)
+// table of supported clients (reverse search, sequence of entries matters!)
 static const auto bultinClientInfo = std::array<ClientInfo, 8> {
     {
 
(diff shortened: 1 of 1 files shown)


whyman:
This looks sensible. Though I would have probably reordered the array :smile: .

I thought the issue in #1346 is that the config should win over the built in matches?

Disclaimer: I havent had a coffee yet today.

KarlStraussberger:
> 
> 
> This looks sensible. Though I would have probably reordered the array 😄 .
> 
> I thought the issue in #1346 is that the config should win over the built in matches?
> 
> Disclaimer: I havent had a coffee yet today.

I'm an hour and two coffees ahead of you 😄 Made the fix while having the second one. The configured clients are appended to the vector. So if the buildin values match, the configured ones are never checked.

PR #1352, opened 2021-03-31
web/clients: remove C string functions

neheb:
They are not available under MinGW. It's also easier with fmt.


neheb:
I verified that format string to be identical to asctime. But I have no idea what trimString does. I tried it on godbolt but got garbage output.

KarlStraussberger:
Thanks.

`trimString` should not be necessary for fmt

PR #1355, opened 2021-03-31
DB Version Update: New value is the first argument

KarlStraussberger:


src/database/mysql/mysql_database.cc
```diff
@@ -248,9 +248,9 @@ void MySQLDatabase::init()
             for (const auto& upgradeCmd : upgrade) {
                 _exec(upgradeCmd);
             }
-            _exec(fmt::format(MYSQL_UPDATE_VERSION, version, version + 1).c_str());
-            log_info("Database upgrade successful.");
+            _exec(fmt::format(MYSQL_UPDATE_VERSION, version + 1, version).c_str());
             dbVersion = fmt::to_string(version + 1);
+            log_info("Database upgrade to version {} successful.", dbVersion.c_str());
```
whyman, line 253: For future ref you can pass std::string directly to fmt.

KarlStraussberger: I remembered some discussion 😬 we had a year ago, that's why I added c_str since.

PR #1357, opened 2021-04-03
random fixes

neheb:


whyman:
You can rebase on master to fix the CI

src/file_request_handler.cc
```diff
@@ -92,12 +92,12 @@ void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
...
     auto res_id_it = params.find(URL_RESOURCE_ID);
-    if (res_id_it != params.end() && res_id_it->second != URL_VALUE_TRANSCODE_NO_RES_ID)
-        res_id = std::stoi(res_id_it->second);
-    else
-        res_id = SIZE_MAX;
+    auto res_id = size_t { [&]() {
+        if (res_id_it != params.end() && res_id_it->second != URL_VALUE_TRANSCODE_NO_RES_ID)
+            return size_t(std::stoi(res_id_it->second));
+        return std::numeric_limits<size_t>::max();
+    }() };
```
KarlStraussberger, line 100: I prefer initializing with max value and the following if. The nesting is quite unreadable to me.

src/file_request_handler.cc
```diff
@@ -231,12 +231,12 @@ std::unique_ptr<IOHandler> FileRequestHandler::open(const char* filename, enum U
...
```
KarlStraussberger, line 239: Same here. Init with max and check afterwards

src/file_request_handler.cc
```diff
@@ -264,13 +264,13 @@ std::unique_ptr<IOHandler> FileRequestHandler::open(const char* filename, enum U
...
-        else {
+        auto res_handler = int { [&]() {
+            if (!rh.empty())
+                return std::stoi(rh);
+
             auto resource = obj->getResource(res_id);
-            res_handler = resource->getHandlerType();
-        }
+            return resource->getHandlerType();
+        }() };
```
KarlStraussberger, line 273: A simple `cond ? truevalue : falsevalue` seems clearer to me here.

PR #1358, opened 2021-04-03
change C macros to numeric_limits

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

PR #1359, opened 2021-04-03
Actions: CodeQL: Update package list before installs

whyman:


whyman:
Sseems like this is borked in #1357 and #1358

PR #1360, opened 2021-04-03
Allow search for albums and genre and other containers

### LINKED ISSUE #1281: Extend supported devices by Denon RCD-N9
I have a Denon RCD-N9 system and i am successfully using it with gerbera since a couple of years. So Denon RCD-N9 should probably be added to the list of supported devices.

### LINKED ISSUE #1287: File listing issue in Denon HEOS app
@KarlStraussberger in #1281 I saw you mentioned the Denon HEOS multiroom system. Yesterday I happened to try Gerbera with the HEOS app (on Android). I noticed that, although the Gerbera server is found, it doesn't list my albums, artists, etc. in the main search view, i.e. when I go to any of the options (except 'Door mappen bladeren', which means 'Browse through folders; see below):
Screenshot_20210306-235127_HEOS_small

For example, if I click on "Albums", I get an empty listing ("Geen informatie beschikbaar' means 'No information available"):
Screenshot_20210306-235135_HEOS_small

If, however, I go to the "Browse through folders" menu item I see:
Screenshot_20210306-235232_HEOS_small
From there I can e.g. go to "Audio" and find my audio files. So after some digging through the HEOS app, I can find my media, but not from the main screen. @KarlStraussberger do you observe this too? 

I run Gerbera on a server with Ubuntu 20.04 (an LXC container) with [the `.deb` package created by @whyman in this comment on #1264](https://github.com/gerbera/gerbera/issues/1264#issuecomment-787873485). 

I also tried to install MiniDLNA/ReadyMedia and Universal Media Server (in separate containers) and those do show the artists, albums, etc. when selecting Albums etc. in the main screen of the HEOS app. This leads me to believe it could be some misconfiguration of Gerbera on my side, or maybe a difference in the way the information is advertised on the network (I have no knowledge of UPnP, so this is a wild guess).

KarlStraussberger:
Container search is looking for 'dc:title' which has to be set on import

Search operator `derivedfrom` should also match base class (e.g. object.container.album.musicAlbum)

fixes #1287 
fixes #1281

### DIFF
--- a/doc/supported-devices.rst
+++ b/doc/supported-devices.rst
@@ -86,6 +86,8 @@ Denon
 -  AVR-4308
 -  S-52
 -  ASD-3N
+-  RCD-N9
+-  HEOS Multiroom, all devices with buildin HEOS by Denon or Marantz
 
 D-Link
 ~~~~~~
--- a/scripts/js/common.js
+++ b/scripts/js/common.js
@@ -358,6 +358,7 @@ function addAudio(obj) {
     chain.album.meta[M_ALBUM] = album;
     chain.artist.meta[M_ARTIST] = artist;
     chain.artist.meta[M_ALBUMARTIST] = artist;
+    chain.genre.meta[M_GENRE] = genre;
     chain.year.meta[M_DATE] = date;
     chain.composer.meta[M_COMPOSER] = composer;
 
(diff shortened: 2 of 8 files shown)


KarlStraussberger:
Happy Easter. 🥚 🐰

PR #1361, opened 2021-04-04
replace find() end() comparisons with count()

neheb:
Shorter


ViniDalvino (approved):
yea that seem pretty pogger

PR #1363, opened 2021-04-05
massive std::chrono conversion

neheb:

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -120,7 +120,7 @@ static constexpr bool IS_FORBIDDEN_CDS_ID(int id) { return id <= CDS_ID_FS_ROOT;
 #define DEFAULT_RESOURCES_CASE_SENSITIVE YES
 #define DEFAULT_UPNP_STRING_LIMIT (-1)
 #define DEFAULT_SESSION_TIMEOUT 30
-#define SESSION_TIMEOUT_CHECK_INTERVAL (5 * 60)
+static constexpr auto SESSION_TIMEOUT_CHECK_INTERVAL = std::chrono::minutes(5);
 #define DEFAULT_PRES_URL_APPENDTO_ATTR "none"
 #define DEFAULT_ITEMS_PER_PAGE_1 10
 #define DEFAULT_ITEMS_PER_PAGE_2 25
@@ -143,7 +143,7 @@ static constexpr bool IS_FORBIDDEN_CDS_ID(int id) { return id <= CDS_ID_FS_ROOT;
 #endif
 
 #ifdef ONLINE_SERVICES
-#define CFG_DEFAULT_UPDATE_AT_START 10 // seconds
+static constexpr auto CFG_DEFAULT_UPDATE_AT_START = std::chrono::seconds(10);
 #endif
 
 #define DEFAULT_TRANSCODING_ENABLED NO
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -169,10 +169,10 @@ void ContentManager::run()
         try {
             auto sc = std::make_shared<SopCastService>(self);
 
-            int i = config->getIntOption(CFG_ONLINE_CONTENT_SOPCAST_REFRESH);
+            auto i = std::chrono::seconds(config->getIntOption(CFG_ONLINE_CONTENT_SOPCAST_REFRESH));
             sc->setRefreshInterval(i);
 
-            i = config->getIntOption(CFG_ONLINE_CONTENT_SOPCAST_PURGE_AFTER);
+            i = std::chrono::seconds(config->getIntOption(CFG_ONLINE_CONTENT_SOPCAST_PURGE_AFTER));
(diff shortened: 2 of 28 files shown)


neheb:
Basic testing shows nothing wrong. Marking as ready.

src/common.h
```diff
@@ -120,7 +120,7 @@ static constexpr bool IS_FORBIDDEN_CDS_ID(int id) { return id <= CDS_ID_FS_ROOT;
 #define DEFAULT_RESOURCES_CASE_SENSITIVE YES
 #define DEFAULT_UPNP_STRING_LIMIT (-1)
 #define DEFAULT_SESSION_TIMEOUT 30
-#define SESSION_TIMEOUT_CHECK_INTERVAL (5 * 60)
+constexpr auto SESSION_TIMEOUT_CHECK_INTERVAL = std::chrono::seconds(5 * 60);
```
whyman, line 123: ::minutes?

src/content/content_manager.cc
```diff
@@ -238,15 +238,15 @@ void ContentManager::run()
...
+            timer->addTimerSubscriber(this, std::chrono::minutes(1), param, true);
         }
     }
 #endif
 
     for (size_t i = 0; i < autoscan_timed->size(); i++) {
         std::shared_ptr<AutoscanDirectory> adir = autoscan_timed->get(i);
         auto param = std::make_shared<Timer::Parameter>(Timer::Parameter::timer_param_t::IDAutoscan, adir->getScanID());
-        log_debug("Adding timed scan with interval {}", adir->getInterval());
+        log_debug("Adding timed scan with interval {}", adir->getInterval().count());
```
whyman, line 249: Format supports printing `chrono` stuff if you include the header

src/database/sql_database.cc
```diff
@@ -1797,7 +1797,7 @@ void SQLDatabase::addAutoscanDirectory(std::shared_ptr<AutoscanDirectory> adir)
       << quote(AutoscanDirectory::mapScanmode(adir->getScanMode())) << ','
       << mapBool(adir->getRecursive()) << ','
       << mapBool(adir->getHidden()) << ','
-      << quote(adir->getInterval()) << ','
+      << quote(adir->getInterval().count()) << ','
```
whyman, line 1800: count is not guaranteed to be seconds right?, its just the count of the "Interval"

src/util/upnp_quirks.cc
```diff
@@ -61,7 +61,7 @@ void Quirks::restoreSamsungBookMarkedPosition(const std::shared_ptr<CdsItem>& it
 {
     if ((pClientInfo->flags & QUIRK_FLAG_SAMSUNG_BOOKMARK_SEC) == 0 && (pClientInfo->flags & QUIRK_FLAG_SAMSUNG_BOOKMARK_MSEC) == 0)
         return;
-    auto positionToRestore = item->getBookMarkPos();
+    auto positionToRestore = item->getBookMarkPos().count();
```
whyman, line 64: BookMarkPos can be either MS or seconds I think? How are you handling that?

whyman (commented):
Some initial thoughts

src/common.h
neheb: Good point.

src/content/content_manager.cc
neheb: Right. The ostream header also supports printing fs::path directly. I'm intentionally avoiding including it to avoid bloating the code.

src/database/sql_database.cc
neheb: I believe it is. getInterval() returns std::chrono::seconds. My reading of https://en.cppreference.com/w/cpp/chrono/duration/count indicates count() just converts it to an integer.

src/util/upnp_quirks.cc
neheb: No idea about this one.

src/content/content_manager.cc
```diff
@@ -325,7 +325,7 @@ void ContentManager::shutdown()
             if (dir != nullptr) {
                 auto dirEnt = fs::directory_entry(dir->getLocation());
                 if (dirEnt.is_directory()) {
-                    auto t = to_time_t(dirEnt.last_write_time());
+                    auto t = std::chrono::duration_cast<std::chrono::seconds>(dirEnt.last_write_time().time_since_epoch());
```
KarlStraussberger, line 328: to_time_t should be obsolete and removed.

neheb: Removed.

KarlStraussberger:
Why dont't you use `zero` in all places instead of `seconds(0)`?

neheb:
Oversight :). I’ll fix later

neheb:
Fixed.

neheb:
Removed most timespec calls. The remaining ones are fairly self contained.

KarlStraussberger:
Thank you for your work. Let's merge this for now and allow users to benefit from it.

PR #1365, opened 2021-04-07
auto&& conversions

neheb:
Shorter.

### DIFF
--- a/src/cds_objects.cc
+++ b/src/cds_objects.cc
@@ -66,7 +66,7 @@ void CdsObject::copyTo(const std::shared_ptr<CdsObject>& obj)
     obj->setAuxData(auxdata);
     obj->setFlags(objectFlags);
     obj->setSortPriority(sortPriority);
-    for (auto& resource : resources)
+    for (auto&& resource : resources)
         obj->addResource(resource->clone());
 }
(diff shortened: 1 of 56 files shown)


whyman:
Are `auto&&` const by default? Or does this just remove const-ness?

neheb:
auto&& binds to anything, including const references. I do believe it evaluates to const ref first if possible.

whyman:
I think I prefer to keep the explicit const, if only for the humans.

Otherwise looks fine to me.

neheb:
I'll do some testing.

neheb:
Actually now that I think about it, that wouldn't make sense. Adding const would prevent copy elision in certain situations.

phi-whiterabbit:
If i may ask: What exactly is the point of such a change, except to obfuscate the code by hiding from the reader that all of these are references to a const object?

phi-whiterabbit:
What i mean is: All the places where const auto & was used it was clear to the reader that it is a reference to a const object + having something different there like a reference to a mutable object would mit make sense.

Now i have to do the type deduction when reading.

neheb:
Not all of them are references to a const object. Some of them are normal references.

PR #1367, opened 2021-04-07
c_str() conversions

neheb:


whyman:
If anything I want to do the opposite 😊

PR #1368, opened 2021-04-07
import constexpr map library and replace pair arrays

neheb:
Avoids doing manual std::array<std::pair>>. Also supports normal methods
supported by std::map except for insertions.


neheb:
This probably needs work. Converting to draft.

neheb:
Yeah never mind.

PR #1369, opened 2021-04-07
clang-tidy: replace typedef with using

neheb:
Found with modernize-use-using


whyman:
Thanks!

Issue #917, opened 2020-06-16
1.5.0: Transcoding: Playback ends before file is complete


ectotropic:
New issue since moving to 1.5.0 recently, where when transcoding the playback of files ends before the file is actually finished. Prior to this version things worked perfectly.

Log file of the relevant section attached, with debug info enabled. (I've trimmed it down to just the relevant parts).

Best I can figure the transcoder is killed before it has actually finished transcoding, and the amount of file that is left unplayed is related to the buffer size set for transcoding.

System is Ubuntu 20.04 (also a recent update, and potentially related?), and compiling gerbera from source (using the 1.5.0 release tag), and compiling the associated requirements too) 

Transcoding is done via a shell script which (using exec) runs a perl script which deals with the actual transcoding - the output logs from those indicate the shell script is always terminated by gebera before the script has completed. I tried editing the buffer size and it seems to change when the playback stops - as does changing what transcoding does (i.e. remuxing only seems to cause earlier stops than actually transcoding). Either of those may be red-herrings though.

Had a look at the source, but following the chain of events that leads to a 'kill' command is not easy - normally I'd stick a breakpoint in, but doing remote linux debugging is not something I know anything about.

My hunch is that there's something happening with the FIFO and/or buffering where it's stopping serving the file *after* the transcoder has closed the pipe, but *before* the pipe data has all been sent to the device. Although I could easily be wrong! Whatever is happening, the transcoding script, is *always* killed before it has exited completely (although I think it's probably in the process of doing so when it is killed).

Sadly, it's a huge time-sink to test this for me, so although I'm happy to provide extra information as need be, I may not be able to run many test on this.

[gerbera_edit.log](https://github.com/gerbera/gerbera/files/4787170/gerbera_edit.log)

ectotropic:
Finally managed to verify that this is also an issue with the latest developement version, and also does seem to be related to buffer size: larger the buffer, the more of a file is truncated.

ectotropic:
More info: seems that it's actually truncating **both** ends of the file when transcoding. The truncation of the start of the file seems not related to the buffer size, but the first section of the video definitely does not show - missed this earlier because it's a pretty short skip and with most files I have it doesn't actually matter (it's just a lead-in pre-roll thing). If the transcoding buffer is small this is similar to how the end of files are truncated: it's pretty easy to miss as it is mostly not anything relevant (credits, etc.).

whyman:
Interesting, sounds almost like an off by one to me...

[...]

PR #1372, opened 2021-04-07
Fix transcoded media ends before file is complete

dyseg:
Also removed ATTR_TRANSCODING_PROFILES_PROFLE_USECHUNKEDENC option, which wasn't used.


src/iohandler/process_io_handler.cc
```diff
@@ -240,26 +240,6 @@ size_t ProcessIOHandler::read(char* buf, size_t length)
...
-                if (mainProc->getStatus() == EXIT_SUCCESS)
-                    ret = 0;
-
-            } else {
-                mainProc->kill();
-            }
-        } else
-            ret = 0;
-
-        killAll();
```
KarlStraussberger, line 259: Are you sure, we do not need this kind of cleanup. I am afraid of memory leaks or dead threads.

KarlStraussberger:
Otherwise looks good to me.

KarlStraussberger:
Fixes #917

src/iohandler/process_io_handler.cc
dyseg: I looked into it again. The transcoding process is always cleaned up eventually by calling `ProcessIOHandler::close()` either by the `BufferedIOHandler` class which holds the instance of `ProcessIOHandler`, or from the destructor. But with this modification it is only cleaned up when we transferred all content to the client, until then the transcoding process is in zombie state. For how much time it depends on the buffer settings, but likely for 10-20-30 seconds. I'll push changes to prevent this shortly.

neheb:
This breaks npupnp support. UPNP_USING_CHUNKED is not implemented.

dyseg:
> This breaks npupnp support. UPNP_USING_CHUNKED is not implemented.

You're right. Should we restore the original value in case of `npupnp` until a solution is found? I just checked and with `-1` and `npupnp` the transcoding is again ends before it should be. But at least this part of the code would compile.
```
#if defined(USING_NPUPNP)
        UpnpFileInfo_set_FileLength(info, -1);
#else
        UpnpFileInfo_set_FileLength(info, UPNP_USING_CHUNKED);
#endif
```

KarlStraussberger:
> > This breaks npupnp support. UPNP_USING_CHUNKED is not implemented.
> 
> You're right. Should we restore the original value in case of `npupnp` until a solution is found? I just checked and with `-1` and `npupnp` the transcoding is again ends before it should be. But at least this part of the code would compile.
> ```
> #if defined(USING_NPUPNP)
>         UpnpFileInfo_set_FileLength(info, -1);
> #else
>         UpnpFileInfo_set_FileLength(info, UPNP_USING_CHUNKED);
> #endif
> ```

Can you provide a PR, please?

neheb:
@dyseg I posted an issue at npupnp upstream. If implemented, no. Although the minimum version should be updated.

neheb:
Maybe just adding ifndef UPNP_USING_CHUNKED is enough.

PR #1379, opened 2021-04-08
Cleanup path handling

KarlStraussberger:
Simplify map access


src/util/tools.h
```diff
@@ -87,17 +87,12 @@ std::string reduceString(std::string str, char ch);
...
 
-/// \brief Get last write time of the specified file or path, if it does not exist it will throw an exception
-/// \param path file or directory to be checked.
-/// \return last modification time of the path or directory
-time_t getLastWriteTime(const fs::path& path);
-
 /// \brief Checks if the given file is a regular file (imitate same behaviour as std::filesystem::is_regular_file)
-bool isRegularFile(const fs::path& path);
 bool isRegularFile(const fs::path& path, std::error_code& ec) noexcept;
+bool isRegularFile(const fs::directory_entry& dirEnt, std::error_code& ec) noexcept;
```
neheb, line 92: Why not replace this with a template?

src/config/config_setup.cc
```diff
@@ -184,7 +184,9 @@ bool ConfigPathSetup::checkAgentPath(std::string& optValue)
 {
     std::string tmp_path;
     if (fs::path(optValue).is_absolute()) {
-        if (!isRegularFile(optValue) && !fs::is_symlink(optValue)) {
+        std::error_code ec;
+        fs::directory_entry dirEnt(optValue, ec);
+        if (!isRegularFile(dirEnt, ec) && !dirEnt.is_symlink(ec)) {
```
neheb, line 189: I would rather use fs::is_symlink(dirEnt, ec) here and elsewhere.

It's more generic and helps reduce changes when refactoring.

src/content/content_manager.cc
```diff
@@ -315,8 +323,9 @@ void ContentManager::shutdown()
             log_debug("AutoScanDir {}", i);
             std::shared_ptr<AutoscanDirectory> dir = autoscan_inotify->get(i);
             if (dir != nullptr) {
-                if (fs::is_directory(dir->getLocation())) {
-                    auto t = getLastWriteTime(dir->getLocation());
+                auto dirEnt = fs::directory_entry(dir->getLocation());
+                if (dirEnt.is_directory()) {
+                    auto t = to_time_t(dirEnt.last_write_time());
                     dir->setCurrentLMT(dir->getLocation(), t);
```
neheb, line 329: setCurrentLMT should probably be taking std::chrono::seconds or something.

src/util/tools.cc
```diff
@@ -230,23 +217,23 @@ bool isRegularFile(const fs::path& path, std::error_code& ec) noexcept
...
-    int ret = stat(path.c_str(), &statbuf);
+    int ret = stat(dirEnt.path().c_str(), &statbuf);
     if (ret != 0) {
         throw_std_runtime_error("{}: {}", std::strerror(errno), path.c_str());
     }
 
     return statbuf.st_size;
 #else
-    return fs::file_size(path);
+    return dirEnt.file_size();
```
neheb, line 236: What's the difference here?

src/util/tools.h
KarlStraussberger: Because the code is not generic.

src/config/config_setup.cc
KarlStraussberger: The point here is the sequence of `stats`. directory_entry encapsulates this.

src/content/content_manager.cc
KarlStraussberger: Good point. The focus was on filesystem operations. I put it on my todo list.

src/util/tools.cc
KarlStraussberger: The scanning loop already has `directory_entry` and the constructor does the `stat`s. By using the method we avoid doing it all over again. The static method only takes `path`. That's the core of the cleanup: use `directory_entry` if already created.

whyman:
LGTM! Needs a rebase though, otherwise happy for it to go in. Looks good.

whyman (approved):

whyman:
@KarlStraussberger it doesnt like the rebase, are you happy to sort or shall we squash instead?

KarlStraussberger:
> @KarlStraussberger it doesnt like the rebase, are you happy to sort or shall we squash instead?

My PR got corrupted while on a rather slow remote connection, so rebasing gets some trouble.

PR #1380, opened 2021-04-08
Clean compilation and errno fix

whorfin:
Simple change, does what it says on the tin for clean compilation


neheb (approved):

src/config/config_manager.cc
```diff
@@ -1321,12 +1321,12 @@ void ConfigManager::load(const fs::path& userHome)
     setOption(root, CFG_SERVER_ALIVE_INTERVAL);
     setOption(root, CFG_IMPORT_MAPPINGS_MIMETYPE_TO_UPNP_CLASS_LIST);
 
-    auto useInotify = setOption(root, CFG_IMPORT_AUTOSCAN_USE_INOTIFY)->getBoolOption();
-
     args["hiddenFiles"] = getBoolOption(CFG_IMPORT_HIDDEN_FILES) ? "true" : "false";
     setOption(root, CFG_IMPORT_AUTOSCAN_TIMED_LIST, &args);
 
 #ifdef HAVE_INOTIFY
+    auto useInotify = setOption(root, CFG_IMPORT_AUTOSCAN_USE_INOTIFY)->getBoolOption();
```
KarlStraussberger, line 1328: The property may not be used here, but the config option has to be set, otherwise we get a runtime issue when inotify is not compiled in. Add `[[maybe_unused]]` in front.

KarlStraussberger (changes requested):

src/config/config_manager.cc
whorfin: What is the runtime issue?  References to `CFG_IMPORT_AUTOSCAN_USE_INOTIFY` are all inside code which is similarly ifdef'd out as far as I can tell.  
Did I miss something?

KarlStraussberger: It used to be 🕐 . Looks like we managed to move all other refs into the `#ifdef`

So you may go the other way and move the enum symbol and the ConfigSetup into the `#ifdef` as well

whorfin: groovy
I think you need to do something to tell github you are no longer requesting a change to the change, so the PR can continue as-is?

KarlStraussberger:
Before merge, I ask you to move `CFG_IMPORT_AUTOSCAN_USE_INOTIFY` into `#ifdef HAVE_INOTIFY`, please. (it's in config,h.

whorfin:
Good point, I missed that - done

KarlStraussberger:
> Good point, I missed that - done

You also have to move its other usage in `config_manager.cc`

whorfin:
too many builds and trees; this time for sure

KarlStraussberger (approved):

PR #1386, opened 2021-04-09
fix UPNP_USING_CHUNKED breaks npupnp compatibility

dyseg:
Temporary fix until implemented in npupnp (expected in version 4.1.3).

### DIFF
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -172,7 +172,11 @@ void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
                 mimeType += fmt::format(";channels={}", nrch);
         }
 
+#ifdef UPNP_USING_CHUNKED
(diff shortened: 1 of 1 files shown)


neheb:
4.1.3 was released at https://www.lesbonscomptes.com/upmpdcli/pages/downloads.html

git was not updated yet.

dyseg:
Okay, so it is indeed defined in `npupnp 4.1.3`, gerbera can be compiled, which is great.
But it is not used in the source code anywhere. Meanwhile in `pupnp`:
https://github.com/pupnp/pupnp/search?q=UPNP_USING_CHUNKED

Most probably the same functionality is also implemented in `npupnp`, but should be "triggered" in a different way. To be clear: the transcoded media playback still ends before it should be with `npupnp 4.1.3`.

neheb:
ping @medoc92

medoc92:
This is fixed in libnpupnp 4.1.3
Rosen Penev writes:

Issue #1377, opened 2021-04-08
gerbera 1.8.0 does not compile with libnpupnp


phi-whiterabbit:
When trying to compile gerbera 1.8.0 using -DWITH_NPUPNP=1, then compilation fails with

src/server.cc:139:87: error: ‘UpnpGetServerUlaGuaPort6’ was not declared in this scope; did you mean ‘UpnpGetServerPort6’?                                                                                                                                                                139 | GLA: Server bound to: {}:{}", UpnpGetServerUlaGuaIp6Address(), UpnpGetServerUlaGuaPort6());

I am using libnpupnp-4.0.14.

neheb:
libnpupnp 4.0.20 fixes this.

phi-whiterabbit:
> libnpupnp 4.0.20 fixes this.

Hm?? You mean 4.1.2 isn't it?

neheb:
Yeah. Latest version was specifically made to fix this.

phi-whiterabbit:
I'll give it a try.
However, the cmake dependency should then reflect this, so that things are failing earlier, no?

phi-whiterabbit:
The doc still lists 4.0.11 as minimal version, as well as the CMakeLists.txt. The minimal version of libupnp is incorrect in the doc, but correct in the CMakeLists.txt.

neheb:
PRs welcome :).

phi-whiterabbit:
Gerbera compiles successfully and works for me using libnpupnp-4.1.2.

PR #1388, opened 2021-04-11
Fixed required versions of libupnp and libnpupnp.

phi-whiterabbit:
Documented required versions of libupnp and libnpupnp are too old.
The version of libnpupnp required by CMakeLists.txt file is too old (see #1377).


whyman:
Thanks!

PR #1392, opened 2021-04-11
Docker.md: add docker-compose instructions

neheb:


hmoffatt:
The docker-compose instructions refer to a named volume called gebera-config, which is not defined in the instructions.

neheb:
doesn't it get automatically created?

hmoffatt:
No:
```
ERROR: Named volume "gerbera-config:/var/run/gerbera:rw" is used in service "gerbera" but no declaration was found in the volumes section.
```

hmoffatt:
You need to declare a volume with:

```
volumes:
  gerbera-config: {}

```

PR #1395, opened 2021-04-12
string_view conversions

neheb:

### DIFF
--- a/src/content/scripting/script.cc
+++ b/src/content/scripting/script.cc
@@ -272,7 +272,7 @@ Script::Script(std::shared_ptr<ContentManager> content,
 
             duk_push_object(ctx);
             setProperty(ConfigSetup::removeAttribute(ATTR_AUTOSCAN_DIRECTORY_LOCATION), adir->getLocation());
-            setProperty(ConfigSetup::removeAttribute(ATTR_AUTOSCAN_DIRECTORY_MODE), AutoscanDirectory::mapScanmode(adir->getScanMode()));
+            setProperty(ConfigSetup::removeAttribute(ATTR_AUTOSCAN_DIRECTORY_MODE), AutoscanDirectory::mapScanmode(adir->getScanMode()).data());
             setIntProperty(ConfigSetup::removeAttribute(ATTR_AUTOSCAN_DIRECTORY_INTERVAL), adir->getInterval());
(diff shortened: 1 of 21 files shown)


src/content/scripting/script.cc
```diff
@@ -803,7 +803,7 @@ std::string Script::convertToCharset(const std::string& str, charset_convert_t c
         return _f2i->convert(str);
     case J2I:
         return _j2i->convert(str);
-    default:
+    case I2I:
```
whyman, line 806: Doesnt this change the behaviour of the method?

src/database/mysql/mysql_database.cc
```diff
@@ -284,6 +284,21 @@ std::string MySQLDatabase::quote(std::string value) const
     return ret;
 }
 
+std::string MySQLDatabase::quote(std::string_view value) const
```
whyman, line 287: Wouldn't a generic method that takes a char array and length be better? this is pretty much a copy of the above?

whyman:
You cant assume that `data()` gives you a null terminated string, it often wont. This feels dodgy to me,

src/content/scripting/script.cc
neheb: No. Initially I moved the default section out of the switch but then got an unhandled switch statement warning. Note how the return how matches the case name.

neheb:
Sure I can. All these functions return a static null terminated string.

src/database/mysql/mysql_database.cc
KarlStraussberger: It should be possible to get rid of the above method complete (with a bit of work on the calls

src/database/sqlite3/sqlite_database.cc
```diff
@@ -284,6 +284,14 @@ void Sqlite3Database::_exec(const char* query)
...
 }
 
+std::string Sqlite3Database::quote(std::string_view value) const
+{
+    char* q = sqlite3_mprintf("'%q'", value.data());
+    std::string ret = q;
+    sqlite3_free(q);
+    return ret;
+}
+
```
KarlStraussberger, line 294: Please try to get rid of the original method

neheb: I think this is what you meant.

PR #1403, opened 2021-04-13
Implement UPnP SortCriteria

KarlStraussberger:
Reworked select statements in `sql_database.cc` on the way

### DIFF
--- a/src/database/database.h
+++ b/src/database/database.h
@@ -63,17 +63,19 @@ protected:
 
     int startingIndex;
     int requestedCount;
+    std::string sortCrit;
 
     // output parameters
     int totalMatches;
 
 public:
     BrowseParam(int objectID, unsigned int flags)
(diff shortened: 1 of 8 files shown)


whyman:
Ive been wanted to change that enum for so long!

I dont suppose you feel like adding some more DB/SQL tests while it is fresh in your memory?

KarlStraussberger:
> Ive been wanted to change that enum for so long!

I simply could not bear the DEFINE anymore :smile:

> I dont suppose you feel like adding some more DB/SQL tests while it is fresh in your memory?

I was already thinking about how to write _simple_ tests, but that would require further work to access all the definitions which is not on my schedule a the moment. Maybe that comes up with the **UPnP-Filter** thing because it requires more flexible metadata browse results (Kodi sends nice `Filter` requests).

PR #1404, opened 2021-04-13
Avoid needless process elevation requirement on illumos

### LINKED ISSUE #1382: server-threads default too aggressive
It looks like a recent change defaults to running threads in system scheduling scope.  If gerbera is not launched as root, at least on my OS config, that results in errors:
```
2021-04-08 15:49:04  error: Could not start thread UpdateThread: Not owner
2021-04-08 15:49:04  error: Could not start thread ContentTaskThread: Not owner
```
...and then gerbera is crippled - rescan doesn't happen, etc

The culprit seems to be a default of "yes" for system-threads; I was able to fix the problem by changing my config.xml with an amended `<system>` declaration:
```
  <server system-threads="no">
```

As running such services as root, or system-privileged user is ... unwise, I would argue that a default of "yes" is needlessly aggressive, and will likely end in tears for many users.  As far as I can tell, it's also a change in behavior, as `pthread_create()` had nullptr attr previously

whorfin:
The default scoping on solaroid systems is in fact PTHREAD_SCOPE_SYSTEM
plus, setting PTHREAD_EXPLICIT_SCHED requires elevated priveleges, thus the default gerbera
behavior would require elevated priveleges, which is poor security practice
fixes #1382 for illumos

### DIFF
--- a/src/util/thread_runner.h
+++ b/src/util/thread_runner.h
@@ -128,6 +128,9 @@ protected:
     /// \brief start the thread
     void startThread() override
     {
+#ifndef SOLARIS
+        // default scoping on solaroid systems is in fact PTHREAD_SCOPE_SYSTEM
+        // plus, setting PTHREAD_EXPLICIT_SCHED requires elevated priveleges
         if (config->getBoolOption(CFG_THREAD_SCOPE_SYSTEM)) {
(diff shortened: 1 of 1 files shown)


KarlStraussberger:
Thank you

PR #1408, opened 2021-04-15
Add support for Windows Media Player

### LINKED ISSUE #1407: Windows Media Player does not display files
The only issue I have with 1.7 is Win7_64 Pro WMP - It shows the Gerbera UPnP server but no files. I have temporarily turned off Windows Firewall for testing. Is there a compatibility issue?

_Originally posted by @geno11x11 in https://github.com/gerbera/gerbera/issues/1399#issuecomment-819731127_

KarlStraussberger:
Rework sql statement generation

fixes #1407

### DIFF
--- a/doc/supported-devices.rst
+++ b/doc/supported-devices.rst
@@ -172,6 +172,12 @@ Medion
 
 -  MD 85651
 
+Microsoft
+~~~~~~~~~
+
+-  Windows Explorer (Browse in directory and play with selectable player)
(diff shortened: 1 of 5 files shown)


whyman:
So WMP was using SEARCH? or? What bit made it work?

KarlStraussberger:
Yes, it runs a full search on all items (that takes some time..). The search uses `@refID exists false` which was not implemented at all. So I added the other `@...` requests as well - you never know 😄

PR #1413, opened 2021-04-21
algorithm cleanups

neheb:
In a different branch, I converted most algorithms to use the new ranges library in C++20. These were things I noticed.


KarlStraussberger:
Great stuff, thank you

PR #1423, opened 2021-04-26
change some size_t to bool

neheb:
These are only used in boolean context.


neheb:
Interestingly since pugixml doesn;t provide empty() methods std::distance must be used. Somewhat annoying.

neheb:
Interesting...

The initial version that failed used != 0 for the std::distance calls.

That's very strange that it can go negative.

Issue #1425, opened 2021-04-28
Support for multiple values in import.js script


imolb:
The id3v2.4.0 standard allows multiple values:
https://id3.org/id3v2.4.0-frames, chapter 4.2
```
4.2.   Text information frames

   [...] All text
   information frames supports multiple strings, stored as a null
   separated list, where null is reperesented by the termination code
   for the charater encoding.
```

Calling the tagreader.cpp example from the taglib 1.12 with the mp3 file attached in #1410, it looks like the taglib library also supports the return of multiple values for the RELEASETYPE attribute:

```
/tagreader sample_releasetype_multi.mp3 
******************** "sample_releasetype_multi.mp3" ********************
-- TAG (basic) --
title   - "sample"
artist  - ""
album   - ""
year    - "0"
comment - ""
track   - "0"
genre   - ""
-- TAG (properties) --
RELEASETYPE - "album"
RELEASETYPE - "compilation"
TITLE       - "sample"
-- AUDIO --
bitrate     - 178
sample rate - 44100
channels    - 2
length      - 0:10
```

To support multiple values in the whole database might be a huge issue.
But in a first step I'd appreciate to have access to the multiple values in the import.js script of Gerbera.

But nevertheless thanks for this great software!

KarlStraussberger:
I appreciate your request for the feature in import scripts, but it means we have to invent some way transfering them to and from javascript stored in a crude way internally. I prefer doing things the proper way - add support for multiple values in resources.

PR #1427, opened 2021-04-29
Concatenate multivalue field with entrySeparator

imolb:
Fix: Concatenate multivalue field with entrySeparator and not space

In ID3v2.4 tags multiple values are allowed (see also #1425).
These values are stored as null separated list in the mp3 file tag.
If taglib is called with textFrame->toString() the multiple values are concatenated with a space.
To concatenate these values with the defined entrySeparater this call is changed to textFrame->fieldList().toString(entrySeparator).

For TXXX tags the first element in the fieldList is the subtag name and extracted separately. The other entries are concatenated with the entrySeparator.

Note: Unfortunately I'm not an C++ programmer, so please revise and correct my code if not optimal. Also I just did only a limited test depth regarding mp3 files and taglib versions.

### DIFF
--- a/src/metadata/taglib_handler.cc
+++ b/src/metadata/taglib_handler.cc
@@ -445,7 +445,7 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
                 const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
                 if (textFrame == nullptr)
                     continue;
-                TagLib::String frameContents = textFrame->toString();
+                TagLib::String frameContents = textFrame->fieldList().toString(entrySeparator);
                 if (!value.empty())
                     value += entrySeparator;
                 if (!legacyEntrySeparator.empty())
@@ -466,12 +466,21 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
                 const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
                 if (textFrame == nullptr)
                     continue;
-                const TagLib::String frameContents = textFrame->toString();
-                std::string value(frameContents.toCString(true));
-
-                size_t subTagEnd = value.find(']');
-                std::string subTag = value.substr(1, subTagEnd - 1); // Cut out brackets
-                std::string content = value.substr(subTagEnd + 2); // Skip bracket and space
+                std::string content = "";
+                std::string subTag = "";
+                TagLib::StringList fieldList = textFrame->fieldList();
+                for(TagLib::StringList::ConstIterator idx = fieldList.begin(); idx != fieldList.end(); ++idx) {
+                    if (idx == fieldList.begin()){
+                        // first element is subTag name
+                        subTag = (*idx).toCString(true);
+                    } else {
+                        content += (*idx).toCString(true) + entrySeparator;
+                    }
+                }
+        
+                // remove entrySeparator at the end again
+                if (content.length() - entrySeparator.length() > 0)
+                    content = content.substr(0, content.length() - entrySeparator.length());
                 // log_debug("TXXX Tag: {}", subTag.c_str());
 
                 if (desiredSubTag == subTag) {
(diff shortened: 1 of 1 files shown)


src/metadata/taglib_handler.cc
```diff
@@ -466,12 +466,21 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
-                const TagLib::String frameContents = textFrame->toString();
-                std::string value(frameContents.toCString(true));
-
-                size_t subTagEnd = value.find(']');
-                std::string subTag = value.substr(1, subTagEnd - 1); // Cut out brackets
-                std::string content = value.substr(subTagEnd + 2); // Skip bracket and space
+                std::string content = "";
+                std::string subTag = "";
+                TagLib::StringList fieldList = textFrame->fieldList();
+                for(TagLib::StringList::ConstIterator idx = fieldList.begin(); idx != fieldList.end(); ++idx) {
```
KarlStraussberger, line 472: You can use `for (auto&& item : textFrame->fieldList()) {` here. `(*idx)` becomes `item` then.

src/metadata/taglib_handler.cc
```diff
@@ -466,12 +466,21 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
+                    if (idx == fieldList.begin()){
+                        // first element is subTag name
+                        subTag = (*idx).toCString(true);
+                    } else {
+                        content += (*idx).toCString(true) + entrySeparator;
+                    }
+                }
+        
+                // remove entrySeparator at the end again
+                if (content.length() - entrySeparator.length() > 0)
```
KarlStraussberger, line 482: better use `(content.length() > entrySeparator.length())` here.

src/metadata/taglib_handler.cc
```diff
@@ -466,12 +466,21 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
-                std::string value(frameContents.toCString(true));
-
-                size_t subTagEnd = value.find(']');
-                std::string subTag = value.substr(1, subTagEnd - 1); // Cut out brackets
-                std::string content = value.substr(subTagEnd + 2); // Skip bracket and space
+                std::string content = "";
+                std::string subTag = "";
+                TagLib::StringList fieldList = textFrame->fieldList();
+                for(TagLib::StringList::ConstIterator idx = fieldList.begin(); idx != fieldList.end(); ++idx) {
+                    if (idx == fieldList.begin()){
```
KarlStraussberger, line 473: Can be `if (content.empty()) {`

src/metadata/taglib_handler.cc
```diff
@@ -466,12 +466,21 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
...
+                        // first element is subTag name
+                        subTag = (*idx).toCString(true);
+                    } else {
+                        content += (*idx).toCString(true) + entrySeparator;
```
KarlStraussberger, line 477: `content = fmt::format("{}{}{}", content, item.toCString(true), entrySeparator);`

KarlStraussberger (changes requested):
Thanks for your PR and your efforts with C++. A really good first run!

I added some suggestions to make the code a bit more elegant. Please, also watch out for spaces as the failing lint check will block the merge

src/metadata/taglib_handler.cc
imolb: Many thanks for the several hints!
I've updated the code accordingly.

src/metadata/taglib_handler.cc
```diff
@@ -466,12 +466,20 @@ void TagLibHandler::extractMP3(TagLib::IOStream* roStream, const std::shared_ptr
                 const auto textFrame = dynamic_cast<const TagLib::ID3v2::TextIdentificationFrame*>(frame);
                 if (textFrame == nullptr)
                     continue;
-                const TagLib::String frameContents = textFrame->toString();
-                std::string value(frameContents.toCString(true));
+                std::string content = "";
+                std::string subTag = "";
+                for (auto&& item : textFrame->fieldList()) {
```
KarlStraussberger, line 471: My fault, but may you rename `item` to `field`. `item` is already used for the `CdsItem` argument.

KarlStraussberger (changes requested):

src/metadata/taglib_handler.cc
imolb: Fixed (and also fixed other logical failure in IF statement)

KarlStraussberger:
Thank you

PR #1430, opened 2021-05-02
Misc improvements

### LINKED ISSUE #1410: TXXX AuxData extraction from MP3 showing subtag in value
With #421 the user defined text frame TXXX was integrated.

I've two issues with this functions:
1. The name of the subtag is included in the value.
2. Multiple separated values are concatenated.

To 1) With MusicBrainz I've added the tag "TXXX:MusicBrainz Album Type" with the value "album". Gebera extracts the value "MusicBrainz Album Type album" for this tag. I would have expected only "album".
To 2) MusicBrainz allows to define several values and separates these with 0x00 00 FF FE. Gerbera seems to pack the values together and separates these with a space. (Maybe the multiple values might be outside of the mp3 specification.)

[sample_releasetype.zip](https://github.com/gerbera/gerbera/files/6332001/sample_releasetype.zip)

sample_releasetype_one.mp3:
`debug: [taglib_handler.cc:478] extractMP3(): Adding auxdata: TXXX:MusicBrainz Album Type with value MusicBrainz Album Type album`

sample_releasetype_multi.mp3
`debug: [taglib_handler.cc:478] extractMP3(): Adding auxdata: TXXX:MusicBrainz Album Type with value MusicBrainz Album Type album compilation`

Tested with recent v1.8.0-33-g48db6739.

### LINKED ISSUE #1429: Slash in track metadata causes undesired containers to be generated with JS import
I have an album named _Shake/Shiver/Moan_. On the filesystem it's been converted to _Shake-Shiver-Moan_ for obvious reasons but the correct name with slashes is present in the metadata.

While writing an import script I noticed that additional containers are generated due to the slashes in the album title. I've confirmed that this also occurs on the default JS import script (i.e. `addAudio`). It does not occur with the builtin importer.

image

I see the documentation makes mention of escaping the slashes under addCdsObject
> - containerChain (string) –
>     - A string, defining where the object will be added in the database hierarchy. The containers in the chain
> are separated by a slash ‘/’, for example, a value of ‘/Audio/All Music’ will add the object to the Audio, All Music container in the server hierarchy. Make sure to properly escape the slash characters in container names. You will find more information on container chain escaping later in this chapter. 
>     - A string, containing the container id as obtained from addContainerTree. In this case the third parameter is not used.

However I can't find any reasonable example of escaping. `createContainerChain` only operates on an array strings which I can't use with `addContainerTree`. Almost all the example scripts use `addContainerTree` over `createContainerChain`. Is the only approach to manually call `escapeSlash` for each metadata field?

KarlStraussberger:
- escape slashes in virtual path
- remove leading TXXX tag
- save last modified for timed_autoscans
- cleanup zlib includes

last weeks' collection 
- fixes #1429
- fixes #1410

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -113,7 +113,7 @@ void ContentManager::run()
     threadRunner = std::make_unique<ThreadRunner<std::condition_variable_any, std::recursive_mutex>>("ContentTaskThread", ContentManager::staticThreadProc, this, config);
 
     if (!threadRunner->isAlive()) {
-        throw_std_runtime_error("Could not start task thread");
+        throw_std_runtime_error("Could not start ContentTaskThread thread");
     }
 
     auto config_timed_list = config->getAutoscanListOption(CFG_IMPORT_AUTOSCAN_TIMED_LIST);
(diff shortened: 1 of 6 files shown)


whyman:
Nice!

PR #1438, opened 2021-05-08
random stuff

neheb:


src/server.cc
```diff
@@ -542,7 +542,7 @@ std::unique_ptr<RequestHandler> Server::createRequestHandler(const char* filenam
 int Server::registerVirtualDirCallbacks()
 {
     log_debug("Setting UpnpVirtualDir GetInfoCallback");
-    int ret = UpnpVirtualDir_set_GetInfoCallback([](const char* filename, UpnpFileInfo* info, const void* cookie, const void** requestCookie) -> int {
+    int ret = UpnpVirtualDir_set_GetInfoCallback([](auto filename, auto info, auto cookie, auto) {
```
KarlStraussberger, line 545: I prefer the full signature, otherwise the code becomes unreadable to me

KarlStraussberger (changes requested):

src/server.cc
whyman: +1, please keep the types.

PR #1439, opened 2021-05-08
Support creation_time on FFmpeg handler. 

mlaerd:
Because I want to classify my videos by recorded date just like my photos, I support `creation_time` on FFmpeg handler.

I confirmed that I can get `creation_time` as `M_DATE` on following video formats.
- MP4
- MOV
- AVI

I see 2 pattern of `creation_time`.
- "2021-05-08T00:00:00.000000Z" (UTC)
- "2021-05-08 00:00:00" (local time)

If `creation_time` is described by UTC, I convert it to local time.
This conversion requires `timegm()` function, which is nonstandard GNU extensions that are also present on the BSDs.

I confirmed that Gerbera supported OSs are supported `timegm()`.
- [Linux](https://www.man7.org/linux/man-pages/man3/timegm.3.html)
- [Free BSD](https://www.unix.com/man-page/FreeBSD/3/ctime/)
- [Net BSD](https://www.unix.com/man-page/netbsd/3/offtime)
- [Mac OS X](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/localtime.3.html)


mlaerd:
I check again.
Please wait for a moment.

mlaerd:
I checked again and confirmed that this PR works correctly.

KarlStraussberger:
Your mapping to `M_DATE` which means you overwrite `date` - or `creation_time` is overwritten. I'd prefer a new metadata property in case both are present. We may map the result the the same upnp-property in the end which we can results in better control of the output.

mlaerd:
> Your mapping to `M_DATE` which means you overwrite `date` - or `creation_time` is overwritten. I'd prefer a new metadata property in case both are present. We may map the result the the same upnp-property in the end which we can results in better control of the output.

Thanks for your advice. I define a new metadata `M_CREATION_DATE`.

Furthermore, I want to update builtin_layout and common.js to add 'Date' and 'Year' directory on addVideo() just like addImage().
This runs correctly on my development, however js e2e test of the PR failed.
Could you help me?

src/metadata/metadata_handler.h
```diff
@@ -123,6 +124,7 @@ static constexpr auto mt_keys = std::array<std::pair<metadata_fields_t, const ch
     { M_ARTIST, "upnp:artist" },
     { M_ALBUM, "upnp:album" },
     { M_DATE, "dc:date" },
+    { M_CREATION_DATE, "dc:date" },
```
KarlStraussberger, line 127: You need to avoid overwriting here also. I recommend `upnp:recordedStartDateTime`. There's no real equivalent in upnp spec. I'm not sure whether there is some sort of property mapping in UPNP_XML

KarlStraussberger:
The test failure is quite often. Needs a kick and then succeeds.

src/metadata/metadata_handler.h
mlaerd: > You need to avoid overwriting here also. I recommend `upnp:recordedStartDateTime`. There's no real equivalent in upnp spec. I'm not sure whether there is some sort of property mapping in UPNP_XML

I found [DCMI Metadata Terms](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/) linked from [/UPnP-av spec](http://upnp.org/specs/av/UPnP-av-ContentDirectory-v1-Service.pdf).
Refer to this, `dc:created` is suitable.
I fixed it.

KarlStraussberger:
Thank you!

PR #1448, opened 2021-05-14
Make transactions for SQLite thread safe

### LINKED ISSUE #1446: Importing m3u files with errors can cause sqlite database errors
Environment: Gerbera 1.8.1 release, Docker image from hub.docker.com

Symptoms: Importing my audio collection, I noticed a database error turning up in the user interface. 

error1

The problem seems to be with M3U files that are malformed or with incorrect paths. I had several M3U files that were encoded with Windows ANSI encoded filenames, where Gerbera seems to expect UTF-8. The problem only seems to happen when there's several bad M3U files; only one doesn't seem to be sufficient. Also it may not fail at the same point on each import.

Attached are the logs from the Docker container, the config.xml file, and an example of a problem M3U file. The log file in particular is interesting, because it first indicates a transaction failure, then fails to rollback, then closes the connection prematurely.

[gerbera.log](https://github.com/gerbera/gerbera/files/6464811/gerbera.log)
[config.xml.txt](https://github.com/gerbera/gerbera/files/6464815/config.xml.txt)
[The Beach Boys - Smiley Smile.m3u.txt](https://github.com/gerbera/gerbera/files/6464825/The.Beach.Boys.-.Smiley.Smile.m3u.txt)

KarlStraussberger:
Fixes #1446

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,9 @@ jobs:
           - name: gcc 9
             cc: gcc-9
             cxx: g++-9
-          - name: clang 9
-            cc: clang-9
-            cxx: clang++-9
+          - name: clang 11
+            cc: clang-11
+            cxx: clang++-11
     env:
       CC: ${{ matrix.compiler.cc }}
       CXX: ${{ matrix.compiler.cxx }}
--- a/src/database/mysql/mysql_database.cc
+++ b/src/database/mysql/mysql_database.cc
@@ -119,7 +119,7 @@ MySQLDatabase::MySQLDatabase(std::shared_ptr<Config> config)
 
 MySQLDatabase::~MySQLDatabase()
 {
-    AutoLock lock(mysqlMutex); // just to ensure, that we don't close while another thread
+    SqlAutoLock lock(sqlMutex); // just to ensure, that we don't close while another thread
     // is executing a query
 
     if (mysql_connection) {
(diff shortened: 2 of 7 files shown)


KarlStraussberger:
@whyman Do you have any clue why clang test fails?

whyman:
It looks like ctest itself is crashing looking for the gtests. Very strange as we havent changed anything in that area. Not sure what we can do to resolve it.

I wonder if there is an issue with some conan package?

KarlStraussberger:
> It looks like ctest itself is crashing looking for the gtests. Very strange as we havent changed anything in that area. Not sure what we can do to resolve it.
> 
> I wonder if there is an issue with some conan package?

Thanks for confirming that I'm not completely off track today 🍷 😺

whyman:
The mac and BSD build also use clang (of different versions) and they seem fine.... strange

KarlStraussberger:
@whyman I think it is an early effect of [[ubuntu] Clang 9 will be removed on May, 17th and default will be set to Clang 11](https://github.com/actions/virtual-environments/issues/3235)

KarlStraussberger:
> 
> 
> @whyman I think it is an early effect of [[ubuntu] Clang 9 will be removed on May, 17th and default will be set to Clang 11](https://github.com/actions/virtual-environments/issues/3235)

Confirmed! But my permissions are no sufficient to change the rules for PR to main branch

whyman:
@KarlStraussberger You are now an admin :tada:

PR #1450, opened 2021-05-15
clang fixes

neheb:


neheb:
weird CI failure. These fixes are just stylistic.

KarlStraussberger:
Unbuntu 20.04 is removing `clang-9`. Hold on until we changed CI check.

KarlStraussberger:
Please rebase that new check is getting active.

neheb:
```
--- a/src/config/config_setup.cc
+++ b/src/config/config_setup.cc
@@ -149,9 +149,6 @@ size_t ConfigSetup::extractIndex(const std::string& item)
     return i;
 }

-const std::string_view ConfigSetup::ROOT_NAME = "config";
-const std::string_view ConfigSetup::ATTRIBUTE = "attribute::";
-
 void ConfigStringSetup::makeOption(const pugi::xml_node& root, const std::shared_ptr<Config>& config, const std::map<std::string, std::string>* arguments)
 {
     bool trim = true;
diff --git a/src/config/config_setup.h b/src/config/config_setup.h
index b4f97fa1..75023d33 100644
--- a/src/config/config_setup.h
+++ b/src/config/config_setup.h
@@ -91,8 +91,8 @@ protected:
     }

 public:
-    static const std::string_view ROOT_NAME;
-    static const std::string_view ATTRIBUTE;
+    static constexpr std::string_view ROOT_NAME = "config";
+    static constexpr std::string_view ATTRIBUTE = "attribute::";

     pugi::xpath_node_set getXmlTree(const pugi::xml_node& element) const;

--
```

seems to not work. wonder why.

KarlStraussberger:
What's the message you get, "not working" leaves some space for interpretation.

neheb:
> What's the message you get, "not working" leaves some space for interpretation.

CI failure. Specifically CI validation / cpp (gcc 9, gcc-9, g++-9) (pull_request)

neheb:
https://github.com/gerbera/gerbera/runs/2595757508?check_suite_focus=true

KarlStraussberger:
That's the js test timing issue - normally we rerun the test and then it passes. No need to push again - we never know whether you're finished or kicking the tests

neheb:
hmm works now. weird.

KarlStraussberger:
``std::data`` may be more generic (which is helpful in generic code), but it makes code less readable and it ends up in calling `.data()`, so please revert.

neheb:
Reverted.

PR #1453, opened 2021-05-17
Improve handling of default config values

### LINKED ISSUE #1447: adding directory to system directories causes error
Hi,

Using the GUI config screen, I added a directory: /authentic-theme to hide it in Filesystem.
That worked, but it instantly makes the config screen inaccesible.

'systemctl status gerbera' shows:
"May 13 08:50:24 NAS gerbera[3571]: 2021-05-13 08:50:24  error: unhandled option /import/system-directories/add-path[26] != /import/system-directories"

The database table grb_config_value shows these 2 records:
item                                     |                                             key                    |                 item_value      |       status
/import/system-directories/add-path[26]          |                /import/system-directories |                       |       manual
/import/system-directories/add-path[26]/attribute::name | /import/system-directories | /authentic-theme | unchanged

Removing these records restored the previous situation. PLease check!

And it seems that none of the config.xml settings are copied to the database at update or start. 
Doing so and make the database/config screen leading, except maybe for the database access settings, would help a lot to make configuration easier, offerings dropdown choices for options, etc!

KarlStraussberger:
Moved defaults and setup of config into separate file.

Implement detection of used detault value.

Display hint on ui that default value is shown (value is not in config.xml)

Goal is to fix #1447


src/config/config_definition.cc
```diff
@@ -960,3 +961,64 @@ const std::map<config_option_t, std::vector<config_option_t>> ConfigManager::par
...
+    if (save)
+        return nullptr;
+
+    throw_std_runtime_error("Error in config code: {} tag not found", option);
+}
+
+std::shared_ptr<ConfigSetup> ConfigDefinition::findConfigSetupByPath(const std::string& key, bool save, const std::shared_ptr<ConfigSetup>& parent)
+{
+    auto co = std::find_if(complexOptions.begin(), complexOptions.end(), [&](auto&& s) { return s->getUniquePath() == key; });
+
```
neheb, line 1024: I would remove this newline as is done above.

src/config/config_setup.h
```diff
@@ -116,6 +117,7 @@ class ConfigSetup {
     ConfigSetup(config_option_t option, const char* xpath, const char* help, bool required = false, const char* defaultValue = "")
         : cpath(buildCpath(xpath))
         , required(required)
+        , useDefault(false)
```
neheb, line 105: This is a redundant member init. You're already initializing it to false.

src/config/config_definition.cc
```diff
@@ -960,3 +961,64 @@ const std::map<config_option_t, std::vector<config_option_t>> ConfigManager::par
...
+    if (parent != nullptr) {
+        auto attrKey = key.substr(parent->getUniquePath().length());
+        if (attrKey.find_first_of(']') != std::string::npos) {
+            attrKey = attrKey.substr(attrKey.find_first_of(']') + 1);
+        }
+        if (attrKey.find_first_of("attribute::") != std::string::npos) {
+            attrKey = attrKey.substr(attrKey.find_first_of("attribute::") + 11);
+        }
+        co = std::find_if(complexOptions.begin(), complexOptions.end(), [&](auto&& s) { return s->getUniquePath() == attrKey && (parentOptions.find(s->option) == parentOptions.end() || parentOptions.at(s->option).end() != std::find(parentOptions.at(s->option).begin(), parentOptions.at(s->option).end(), s->option)); });
+
```
neheb, line 1039: Same here.

src/config/config_definition.h
```diff
@@ -0,0 +1,95 @@
...
+    }
+    static std::string removeAttribute(config_option_t option)
+    {
+        auto attr = std::string(mapConfigOption(option));
+        if (attr.size() > ATTRIBUTE.size() && attr.substr(0, ATTRIBUTE.size()) == ATTRIBUTE)
+            attr = attr.substr(ATTRIBUTE.size());
+        return attr;
+    }
+
+protected:
```
neheb, line 90: These should be private I think. See https://clang.llvm.org/extra/clang-tidy/checks/misc-non-private-member-variables-in-classes.html

KarlStraussberger: Will add these to my next PR which is based on all theses changes (I don't want to mess around in branches)

PR #1460, opened 2021-05-21
clang: remove CTAD in some places

neheb:
This throws -Wctad-maybe-unsupported under clang as gcc's libstdc++
seems to not implement it properly.


neheb:
```
[1]       AssertionError: expected '' to equal 'Successfully removed item'
[1]       + expected - actual
```

Heisenbug I'm guessing.

PR #1471, opened 2021-05-24
clang-analyzer and misc

neheb:


neheb:
I can't figure out GCC's fanalyzer. It complains in the last commit (before and after) about nullptr being in the make_unique function (here and elsewhere). I cannot figure out how that's the case.

PR #1472, opened 2021-05-25
mostly pass by value changes

neheb:


src/config/config_manager.h
```diff
@@ -168,7 +168,7 @@ class ConfigManager : public Config, public std::enable_shared_from_this<Config>
 
     std::unique_ptr<std::vector<std::shared_ptr<ConfigOption>>> options;
 
-    std::shared_ptr<ConfigOption> setOption(const pugi::xml_node& root, config_option_t option, const std::map<std::string, std::string>* arguments = nullptr);
+    std::shared_ptr<ConfigOption> setOption(pugi::xml_node root, config_option_t option, const std::map<std::string, std::string>* arguments = nullptr);
```
KarlStraussberger, line 172: I don't think we should pass the whole XML structure by value.

KarlStraussberger (changes requested):

src/config/config_manager.h
neheb: What’s suspicious is that it’s not even moved. This change might have been automated by clang-tody. I’ll look again.

neheb: Oh I see. It's trivially copyable.

KarlStraussberger (approved):

PR #1483, opened 2021-05-29
initialize several unique_ptrs

neheb:
GCC's -fanalyzer complains that null pointers are created otherwise.

Some cleanup posibilities showed up.

### DIFF
--- a/src/content/update_manager.cc
+++ b/src/content/update_manager.cc
@@ -50,7 +50,7 @@ UpdateManager::UpdateManager(std::shared_ptr<Config> config, std::shared_ptr<Dat
     : config(std::move(config))
     , database(std::move(database))
     , server(std::move(server))
-    , objectIDHash(std::make_unique<std::unordered_set<int>>())
+    , objectIDHash(std::make_unique<std::unordered_set<int>>(MAX_OBJECT_IDS - 1))
     , shutdownFlag(false)
     , flushPolicy(FLUSH_SPEC)
     , lastContainerChanged(INVALID_OBJECT_ID)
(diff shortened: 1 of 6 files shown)


src/config/config_definition.cc
```diff
@@ -130,7 +130,7 @@
 #define DEFAULT_MARK_PLAYED_ITEMS_STRING "*"
 
 /// \brief default values for CFG_IMPORT_SYSTEM_DIRECTORIES
-static std::vector<std::string_view> excludes_fullpath {
+static std::vector<std::string> excludes_fullpath {
```
KarlStraussberger, line 133: What's the problem with `string_view`?

neheb: Some problem in the constructor. Partial log: https://gist.github.com/neheb/73d3206c19a68ab3b2fd9d7bc35dd2d0

FWIW, this does not increase binary size. These strings are short enough to fit the short string optimization.

neheb:
Added extra commit to mostly fix the rest of the unique_ptr warnings. Something tells me it shouldn't work.

src/content/update_manager.cc
```diff
@@ -50,7 +50,7 @@ UpdateManager::UpdateManager(std::shared_ptr<Config> config, std::shared_ptr<Dat
     : config(std::move(config))
     , database(std::move(database))
     , server(std::move(server))
-    , objectIDHash(std::make_unique<std::unordered_set<int>>())
+    , objectIDHash(std::make_unique<std::unordered_set<int>>(MAX_OBJECT_IDS))
```
neheb, line 53: this is probably wrong.

KarlStraussberger:
> Added extra commit to mostly fix the rest of the unique_ptr warnings. Something tells me it shouldn't work.

I'm waiting for this after creating 1.8.2.

neheb:
OK. I split up the PR to keep the last one here.

PR #1488, opened 2021-05-31
initialize several unique_ptrs

neheb:
GCC's -fanalyzer complains that null pointers are created otherwise.

Some cleanup posibilities showed up.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -581,8 +581,6 @@ std::shared_ptr<CdsObject> SQLDatabase::loadObjectByServiceID(const std::string&
 
 std::unique_ptr<std::vector<int>> SQLDatabase::getServiceObjectIDs(char servicePrefix)
 {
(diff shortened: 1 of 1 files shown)


src/database/sql_database.cc
```diff
@@ -1604,9 +1603,8 @@ std::unique_ptr<Database::ChangedContainers> SQLDatabase::_purgeEmptyContainers(
     log_debug("start upnp: {}; ui: {}",
         join(maybeEmpty->upnp, ',').c_str(),
         join(maybeEmpty->ui, ',').c_str());
-    auto changedContainers = std::make_unique<ChangedContainers>();
     if (maybeEmpty->upnp.empty() && maybeEmpty->ui.empty())
-        return changedContainers;
+        return nullptr;
```
neheb, line 1607: These should be equivalent. The only place that uses this has a nullptr check: https://github.com/gerbera/gerbera/blob/master/src/content/content_manager.cc#L804

PR #1493, opened 2021-05-31
move some initializations up

neheb:


src/cds_objects.h
```diff
@@ -122,7 +122,7 @@ class CdsObject {
 
 public:
     /// \brief Constructor. Sets the default values.
-    explicit CdsObject();
+    explicit CdsObject() = default;
```
neheb, line 125: this caused a small size reduction.

neheb: Oh I see. It's not even needed.

PR #1494, opened 2021-05-31
Revert "initialize several unique_ptrs"

KarlStraussberger:
Reverts gerbera/gerbera#1483

Breaks ubuntu:18.04

https://github.com/gerbera/gerbera/runs/2714005392

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -70,19 +70,19 @@ std::shared_ptr<Quirks> ActionRequest::getQuirks() const
 
 std::unique_ptr<pugi::xml_document> ActionRequest::getRequest() const
 {
-    pugi::xml_document request;
+    auto request = std::make_unique<pugi::xml_document>();
 #if defined(USING_NPUPNP)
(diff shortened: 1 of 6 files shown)


KarlStraussberger:
@neheb Have a look at this, please

neheb:
The second commit shouldn't be reverted.

The first is indeed problematic. It's effectively a naive way to avoid allocating a null pointer, which later on becomes non NULL.

KarlStraussberger:
> 
> 
> The second commit shouldn't be reverted.
> 
> The first is indeed problematic. It's effectively a naive way to avoid allocating a null pointer, which later on becomes non NULL.
> 
> edit: oh I see why it fails. PUGIXML has support for move semantics, which seem to not be available with Ubuntu's version.

I can only revert the whole PR. Sorry, at least it was the fastest way to get the build running again. Better to split them up. also.

neheb:
Not a big deal honestly.

PR #1499, opened 2021-06-01
fixed problems with the path of the config

qq7te:
For some reason the path to the config file seems to be inconsistent between what the code looks like and what the script files setup. I hope this can be useful

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,4 +48,4 @@ EXPOSE 49494
 EXPOSE 1900/udp
 
 ENTRYPOINT ["/sbin/tini", "--", "docker-entrypoint.sh"]
-CMD ["gerbera","--port", "49494", "--config", "/var/run/gerbera/config.xml"]
+CMD ["gerbera","--port", "49494", "--config", "/var/run/gerbera/.config/gerbera/config.xml"]
--- a/scripts/docker/docker-entrypoint.sh
+++ b/scripts/docker/docker-entrypoint.sh
@@ -1,15 +1,16 @@
 #!/usr/bin/env sh
 
-if [ ! -f /var/run/gerbera/config.xml ]; then
-  # Generate a config file with home set
-  gerbera --create-config --home /var/run/gerbera > /var/run/gerbera/config.xml
+if [ ! -f /var/run/gerbera/.config/gerbera/config.xml ]; then
+    # Generate a config file with home set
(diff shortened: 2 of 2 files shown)


KarlStraussberger:
The scripts actually overwrite the builtin value to provide a location which is easier to access.

The config file is loaded from any location you point to. Did you encounter any issue?

qq7te:
I just kept creating containers without providing a config file and the result would be an error and the server not starting. 
If you don't see the same problem that I'm seeing, then it's probably a problem with my setup.

I was just trying to contribute back what worked for me. 

so... what happens if you run `docker run --rm gerbera/gerbera:1.8.2`   ? 
this is my output: 
```
Unable to find image 'gerbera/gerbera:1.8.2' locally
1.8.2: Pulling from gerbera/gerbera
540db60ca938: Already exists 
304ec9dcf044: Pull complete 
62bf77780216: Pull complete 
0d0dcc73d0c6: Pull complete 
45c2152897e4: Pull complete 
fb37aa01c016: Pull complete 
5f48be9b6a2b: Pull complete 
c644db531efc: Pull complete 
Digest: sha256:fec406b3ddc7ece408a534c1623718e1976445da2e8dc0cff4a72f8e2606d8c6
Status: Downloaded newer image for gerbera/gerbera:1.8.2
2021-06-08 02:50:16   info: Gerbera UPnP Server 1.8.2 - http://gerbera.io/
2021-06-08 02:50:16   info: Copyright 2016-2021 Gerbera Contributors.
2021-06-08 02:50:16   info: Gerbera is free software, covered by the GNU General Public License version 2
2021-06-08 02:50:16   info: Loading configuration from: /var/run/gerbera/config.xml
2021-06-08 02:50:16   info: Checking configuration...
2021-06-08 02:50:16  error: [/gerbera_build/src/config/config_manager.cc:164] load Error: Directory '/var/run/gerbera/.config/gerbera' does not exist
```

and done. The container does not start.

KarlStraussberger:
I see the issue. It can be solved by adding

```
if [ ! -d /var/run/gerbera/.config/gerbera/ ]; then
    mkdir /var/run/gerbera/.config/gerbera/
fi
```

But actual issue is that `gerbera --create-config --home /var/run/gerbera > /var/run/gerbera/config.xml` does not set `home` correctly

qq7te:
great, thanks for the better fix. Hope to see it in 1.8.3 soon!

PR #1508, opened 2021-06-03
CI: Conan on 18.04

whyman:


KarlStraussberger:
clang-11 is not available in 18.04.
I tried with clang-9 only to get other issues. Looks like the prebuild environment changed and is not consistent at all.

Therefore, I recommend to disable clang for the moment and work on sorting things out once there is a bit more white space on our schedules.

whyman:
Argh, conan is such a nice idea but sucha PITA in practice.

I guess we will have to disable it for now.

KarlStraussberger:
At the moment gcc works on 20.04, so we should just comment clang and disable the CI requirement

KarlStraussberger:
> 
> 
> At the moment gcc works on 20.04, so we should just comment clang and disable the CI requirement

https://github.com/KarlStraussberger/gerbera/blob/refactor-search-handler/.github/workflows/ci.yml

PR #1514, opened 2021-06-05
replace C locale functions with std::locale

neheb:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,17 +406,6 @@ target_link_libraries(libgerbera PUBLIC pugixml::pugixml)
 find_package(SQLite3 3.7.0 REQUIRED)
 target_link_libraries(libgerbera PUBLIC SQLite::SQLite3)
 
-# We should probably move these two out to their own FindLocale
-include(CheckFunctionExists)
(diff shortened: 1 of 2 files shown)


neheb:
failed. wonder why.

KarlStraussberger:
> 
> 
> failed. wonder why.

Hi, you need to rebase you PRs.
We had to remove a CI check because of weird issues in conan packages when compiling with clang.

neheb:
I tested a simplified version locally:

```
#include <string>
#include <locale>

int main()
{
    std::string temp;
    std::locale glo;
    glo.global(std::locale(""));
    std::locale gloc;
    temp = !gloc.name().empty() ? gloc.name() : "empty";
    printf("Global locale is: %s\n", temp.c_str());
}

It works correctly. I have no idea why it's failing here.
```

KarlStraussberger:
It the ConfigGeneratorTest that segfaults. You need to look into it, maybe does some locale stuff also.

neheb:
No it's not it. Gerbera crashes as well. Maybe it's some weird constructor interaction. Anyway, I'm probably not going to bother figuring it out.

PR #1518, opened 2021-06-05
clang-tidy: pass unique_ptr by value

neheb:
Since the copy stuff is deleted, this is safe to do.

Found with google-runtime-references.


neheb:
Funny. https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-conventional partially disagrees with this and recommends using && instead.

PR #1527, opened 2021-06-08
clang-tidy applied to tests

neheb:


test/config/test_configgenerator.cc
```diff
@@ -11,10 +11,10 @@ using namespace ::testing;
 class ConfigGeneratorTest : public ::testing::Test {
 
 public:
-    ConfigGeneratorTest() {};
-    virtual ~ConfigGeneratorTest() {};
+    ConfigGeneratorTest() = default;;
```
KarlStraussberger, line 14: I suppose the second `;` is unwanted and happened by replacing the braces with `default`.

KarlStraussberger (changes requested):

test/config/test_configgenerator.cc
neheb: Sounds about right. The ; was originally not necessary anyway.

neheb:
funny. freebsd build is stuck.

KarlStraussberger:
> funny. freebsd build is stuck.

Sourcehut seems to have stopped sending responses to github

KarlStraussberger (approved):

PR #1528, opened 2021-06-08
clang-tidy: use pointers instead of references

neheb:
Found with google-runtime-references


KarlStraussberger:
I think I prefer the references. Pointer represent dynamic memory (after new). References are scoped objects. Technically it should end up in the same code.
@whyman What do you think?

whyman:
I prefer the references too.

neheb:
OK.

PR #1538, opened 2021-06-12
make bultinClientInfo constexpr

neheb:
Avoids destructor nastiness.

### DIFF
--- a/src/config/client_config.cc
+++ b/src/config/client_config.cc
@@ -37,7 +37,7 @@ ClientConfig::ClientConfig()
     clientInfo->flags = 0;
 }
 
-ClientConfig::ClientConfig(int flags, const std::string& ip, const std::string& userAgent)
+ClientConfig::ClientConfig(int flags, std::string_view ip, std::string_view userAgent)
     : clientInfo(std::make_shared<ClientInfo>())
 {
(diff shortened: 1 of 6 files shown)


neheb:
GCC's libstdc++ is fine. LLVM's libc++ is not. Hmmmmm.

neheb:
As far as I can tell, libc++ has problems assigning to an std::string_view whereas libstdc++ does not. Needs a closer look.

neheb:
Hmm OK. Maybe it's a problem with the types having to match.

As far as the js issue, that seems to be introduced with https://github.com/gerbera/gerbera/commit/89d022b2251d59214b724fc6f6cea7fd2130c76c .

I have absolutely no idea why...

neheb:
meh. heisenbug.

KarlStraussberger:
> 
> Hmm OK. Maybe it's a problem with the types having to match.
> 
> As far as the js issue
> 
> I have absolutely no idea why...

JS issue is around for a year already. I think it is just a problem of timers passing in low priority containers. After kicking the test once again it runs smoothly.

KarlStraussberger:
I'll have to revert most of this change, because it break client detection.

`string_view` refers to the stored pointer even if the target is gone already (e.g. after reading `config.xml`)

PR #1542, opened 2021-06-12
more make_unique changes

neheb:
-fanalyzer claims null pointers are allocated otherwise.

### DIFF
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -313,30 +313,25 @@ void UpnpXMLBuilder::renderResource(const std::string& URL, const std::map<std::
 
 std::unique_ptr<UpnpXMLBuilder::PathBase> UpnpXMLBuilder::getPathBase(const std::shared_ptr<CdsItem>& item, bool forceLocal)
 {
-    auto pathBase = std::make_unique<PathBase>();
     /// \todo resource options must be read from configuration files
 
     std::map<std::string, std::string> dict;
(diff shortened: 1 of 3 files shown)


src/upnp_cds.cc
```diff
@@ -93,15 +93,15 @@ void ContentDirectoryService::doBrowse(const std::unique_ptr<ActionRequest>& req
...
-    param->setRequestedCount(stoiString(requestedCount));
-    param->setSortCriteria(trimString(sortCriteria));
+    param.setStartingIndex(stoiString(startingIndex));
+    param.setRequestedCount(stoiString(requestedCount));
+    param.setSortCriteria(trimString(sortCriteria));
 
     std::vector<std::shared_ptr<CdsObject>> arr;
     try {
-        arr = database->browse(param);
+        arr = database->browse(std::make_unique<BrowseParam>(param));
```
KarlStraussberger, line 104: I'm quite sure this breaks browsing, because browse calls `param->setTotalMatches` which is read further down and I assume this call creates a copy of `param`.

KarlStraussberger (changes requested):

src/upnp_cds.cc
neheb: yes it does.

neheb: Removed.

KarlStraussberger (approved):

PR #1544, opened 2021-06-13
add .clang-tidy file

neheb:

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,50 @@
+---
+Checks:          'clang-diagnostic-*,clang-analyzer-*'
+WarningsAsErrors: ''
+HeaderFilterRegex: '*.h'
+FormatStyle:     Webkit
+CheckOptions:
+  - key:             llvm-else-after-return.WarnOnConditionVariables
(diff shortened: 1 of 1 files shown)


.clang-tidy
```diff
@@ -0,0 +1,52 @@
...
+  - key:             modernize-loop-convert.MaxCopySize
+    value:           '16'
+  - key:             cppcoreguidelines-explicit-virtual-functions.IgnoreDestructors
+    value:           '1'
+  - key:             modernize-use-nullptr.NullMacros
+    value:           'NULL'
+  - key:             llvm-qualified-auto.AddConstToQualified
+    value:           '0'
+  - key:             modernize-loop-convert.NamingStyle
+    value:           CamelCase
```
KarlStraussberger, line 46: We currently have quite a mixture of styles, so we should agree whether to define as standard.

neheb: Removed.

neheb:
I recently looked at https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html

Most of the codebase seems to use camelBack.

PR #1549, opened 2021-06-14
Configuration of folders for resources

### LINKED ISSUE #1501: [Feature Request] Support globs in <add-file>
I've been setting up a media server at home, and I deeply appreciate the efforts made by this community, as everything works almost perfectly. Only real hiccup here is this:

I've been setting up subtitles for the various movies and tv shows I have, and noticed I could add subtitles, if not part of the video itself, using `<resources>` and [`<add-file>`](https://docs.gerbera.io/en/latest/config-import.html#add-file).

While this is great, and the variables replaced in the search string are useful, they still lack some flexibility. I've got some subtitles, for example, that don't use the movie's name (such as 2_English.srt), and I imagine writing a rule for every language I might possibly come across might be inconvenient. This could be easily solved if the `name` argument of the `<add-file>` tag  could take globs, for example, the following config would solve my problems:

```
<add-file name="Subs/*.srt" />
<add-file name="*.srt" />
```

So perhaps something like this could be implemented? Let me know if I can help any further.

KarlStraussberger:
Allows multiple resources of same type

Resources are displayed in web UI and sent to clients (tested with vlc, kodi and foobar2000)

Especially relevant for subtitles

Revert part of #1538

fixes #1501


src/metadata/metacontent_handler.cc
```diff
@@ -231,54 +247,50 @@ std::unique_ptr<IOHandler> ContainerArtHandler::serveContent(std::shared_ptr<Cds
...
+    if (pathList.empty() || pathList[0].empty())
         obj->removeResource(CH_SUBTITLE);
+
+    for (auto&& path : pathList) {
+        log_debug("Running subtitle handler on {} -> {}", obj->getLocation().c_str(), path.c_str());
+        if (!path.empty()) {
+            auto resource = std::make_shared<CdsResource>(CH_SUBTITLE);
+            std::string type = path.extension().string().substr(1);
+
+            std::string mimeType = mime->getMimeType(path, fmt::format("text/{}", type));
```
neheb, line 273: could use a lambda here. But that's probably a micro optimization.

src/metadata/metadata_handler.h
```diff
@@ -176,6 +177,7 @@ static constexpr auto res_keys = std::array<std::pair<resource_attributes_t, con
     { R_FANART_OBJ_ID, "fanArtObject" },
     { R_FANART_RES_ID, "fanArtResource" },
     { R_BITS_PER_SAMPLE, "bitsPerSample" },
+    { R_LANGUAGE, "dc:language" },
```
neheb, line 180: array size is not updated.

KarlStraussberger: `R_MAX` is increased by adding the new enum entry.

neheb: Ah OK

KarlStraussberger: Thanks for the hint: the entry is still missing in `script_names.h`

PR #1552, opened 2021-06-15
replace stringHash with single accumulate call

neheb:
std::accumulate is shorter and generates more efficient assembly.


neheb:
https://godbolt.org/z/c8cb8K1qE

compilers have gotten better looks like.

PR #1554, opened 2021-06-15
remove non const references

neheb:


KarlStraussberger:
That's breaking. The refs are used to pass results out.

Also, pointers should only be used in case of dynamic memory.

PR #1556, opened 2021-06-15
remove several defines

neheb:
These are hard to read. Convert to constexpr variables and use raw
format strings where needed.

Not sure if these are correct transformations. All done manually.

### DIFF
--- a/src/database/sqlite3/sqlite_database.cc
+++ b/src/database/sqlite3/sqlite_database.cc
@@ -35,88 +35,88 @@
 
 #include "config/config_manager.h"
 
-#define DB_BACKUP_FORMAT "{}.backup"
+static constexpr auto DB_BACKUP_FORMAT = "{}.backup";
 
 // updates 1->2
-#define SQLITE3_UPDATE_1_2_1 "DROP INDEX mt_autoscan_obj_id"
-#define SQLITE3_UPDATE_1_2_2 "CREATE UNIQUE INDEX mt_autoscan_obj_id ON mt_autoscan(obj_id)"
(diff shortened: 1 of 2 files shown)


src/database/sqlite3/sqlite_database.cc
```diff
@@ -35,88 +35,88 @@
...
-#define SQLITE3_UPDATE_3_4_1 "CREATE TABLE \"mt_metadata\" ( \
-  \"id\" integer primary key, \
-  \"item_id\" integer NOT NULL, \
-  \"property_name\" varchar(255) NOT NULL, \
-  \"property_value\" text NOT NULL, \
-  CONSTRAINT \"mt_metadata_idfk1\" FOREIGN KEY (\"item_id\") REFERENCES \"mt_cds_object\" (\"id\") \
-  ON DELETE CASCADE ON UPDATE CASCADE )"
-#define SQLITE3_UPDATE_3_4_2 "CREATE INDEX mt_metadata_item_id ON mt_metadata(item_id)"
+static constexpr auto SQLITE3_UPDATE_3_4_1 = R"(CREATE TABLE "mt_metadata" ()"
+                                             R"("id" integer primary key, )"
```
whyman, line 51: Dont these quotes cause issues?

whyman:
If I am honest, I dont see that this any better, they both look pretty bad to me.

src/database/sqlite3/sqlite_database.cc
neheb: My understanding is no. This whole thing is supposed to be on one line I think.

PR #1560, opened 2021-06-16
lambda conversion

neheb:

### DIFF
--- a/src/content/onlineservice/sopcast_content_handler.cc
+++ b/src/content/onlineservice/sopcast_content_handler.cc
@@ -144,18 +144,17 @@ std::shared_ptr<CdsObject> SopCastContentHandler::getObject(const std::string& g
     // I wish they had a mimetype setting
     //auto mappings = config->getDictionaryOption(CFG_IMPORT_MAPPINGS_EXTENSION_TO_MIMETYPE_LIST);
     //std::string mt = getValueOrDefault(mappings, temp);
-    std::string mt;
-    // map was empty, we have to do construct the mimetype ourselves
-    if (temp == "wmv")
-        mt = "video/sopcast-x-ms-wmv";
-    else if (temp == "mp3")
-        mt = "audio/sopcast-mpeg";
-    else if (temp == "wma")
-        mt = "audio/sopcast-x-ms-wma";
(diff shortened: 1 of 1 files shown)


src/content/onlineservice/sopcast_content_handler.cc
```diff
@@ -144,18 +144,18 @@ std::shared_ptr<CdsObject> SopCastContentHandler::getObject(const std::string& g
...
+        if (temp == "wmv")
+            return "video/sopcast-x-ms-wmv";
+        if (temp == "mp3")
+            return "audio/sopcast-mpeg";
+        if (temp == "wma")
+            return "audio/sopcast-x-ms-wma";
+        return "application/sopcast-stream";
+    }();
+    if (mt == "application/sopcast-stream")
         log_warning("Could not determine mimetype for SopCast channel (stream_type: {})", temp.c_str());
```
KarlStraussberger, line 158: Why don't you move that warning into the lambda?

src/content/onlineservice/sopcast_content_handler.cc
```diff
@@ -144,18 +144,18 @@ std::shared_ptr<CdsObject> SopCastContentHandler::getObject(const std::string& g
     // I wish they had a mimetype setting
     //auto mappings = config->getDictionaryOption(CFG_IMPORT_MAPPINGS_EXTENSION_TO_MIMETYPE_LIST);
```
KarlStraussberger, line 145: I suggest to create a new config dictionary `CFG_IMPORT_SOPCAST_MIMETYE_LIST` defaulting to the current values and some string with the default value.

KarlStraussberger (changes requested):

src/content/onlineservice/sopcast_content_handler.cc
neheb: That's beyond the scope here.

src/content/onlineservice/sopcast_content_handler.cc
neheb: Good catch.

KarlStraussberger (approved):

PR #1562, opened 2021-06-16
use sizeof for snprintf

neheb:

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -344,7 +344,7 @@ int main(int argc, char** argv, char** envp)
 
             // convert to a string
             char pidstr[20];
-            if (0 > snprintf(pidstr, 18, "%d", pid)) {
+            if (0 > snprintf(pidstr, sizeof(pidstr) - 2, "%d", pid)) {
                 log_error("Could not determine pid of running process.");
                 exit(EXIT_FAILURE);
             }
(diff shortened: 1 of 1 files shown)


whyman:
Could we not use fmt?

neheb:
> Could we not use fmt?

I have no idea honestly. src/main uses C functions everywhere. I don't know the reason.

KarlStraussberger:
I know: nobody changed it since the first days of mediatomb

neheb:
Replaced with fmt. The whole codeblock looks like it could be replaced with something simpler.

Anyway, this is good enough for now I think.

src/main.cc
```diff
@@ -342,15 +342,13 @@ int main(int argc, char** argv, char** envp)
...
-            char pidstr[20];
-            if (0 > snprintf(pidstr, 18, "%d", pid)) {
+            if (pid <= 1) {
                 log_error("Could not determine pid of running process.");
-                exit(EXIT_FAILURE);
+                std::exit(EXIT_FAILURE);
             }
 
-            // add a newline
-            strcat(pidstr, "\n");
```
KarlStraussberger, line 353: You need to add the newline to the output to retain the file format

KarlStraussberger (changes requested):

KarlStraussberger (approved):

PR #1565, opened 2021-06-16
fix replace string functions

neheb:
The original reference is modified. There's no need to return a copy.

### DIFF
--- a/src/util/xml_to_json.cc
+++ b/src/util/xml_to_json.cc
@@ -112,9 +112,9 @@ std::string Xml2Json::getAsString(const char* str)
 {
     auto escaped = escape(str, '\\', '"');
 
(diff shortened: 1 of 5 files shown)


src/util/xml_to_json.cc
```diff
@@ -112,8 +112,8 @@ std::string Xml2Json::getAsString(const char* str)
 {
     auto escaped = escape(str, '\\', '"');
 
-    escaped = replaceAllString(escaped, "\n", "\\n");
-    escaped = replaceAllString(escaped, "\r", "\\r");
+    replaceAllString(escaped, "\n", "\\n");
+    replaceAllString(escaped, "\r", "\\r");
     return "\"" + escaped + '"';
```
neheb, line 117: looks like should be using fmt.

PR #1570, opened 2021-06-17
Cleanup exifhandler

KarlStraussberger:
Reorganize endless if

Simplify other code

Addresses
https://github.com/gerbera/gerbera/security/code-scanning/202?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/src/metadata/libexif_handler.cc
+++ b/src/metadata/libexif_handler.cc
@@ -62,289 +62,178 @@ static void setJpegResolutionResource(const std::shared_ptr<CdsItem>& item, size
     }
 }
 
+static const std::map<std::string, int> exifTagMap {
+    { "EXIF_TAG_INTEROPERABILITY_INDEX", EXIF_TAG_INTEROPERABILITY_INDEX },
+    { "EXIF_TAG_INTEROPERABILITY_VERSION", EXIF_TAG_INTEROPERABILITY_VERSION },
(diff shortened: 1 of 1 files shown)


src/metadata/libexif_handler.cc
```diff
@@ -62,289 +62,178 @@ static void setJpegResolutionResource(const std::shared_ptr<CdsItem>& item, size
     }
 }
 
+static const std::map<std::string, int> exifTagMap {
```
neheb, line 65: this is somewhat interesting.

So in the case of non-constexpr classes like std::map, clang throws a -Wglobal-constructor and -Wexit-time-destructor warning. Moving the map inside the function eliminates the warning. https://github.com/gerbera/gerbera/pull/1573 is a silent fix for such warnings. https://github.com/neheb/gerbera/commit/834192a9beced7779946a0587cd627a4be2c8cf0 is the same.

KarlStraussberger: I understand that warning, but is it a real issue as long as we stick to maps and arrays of basic types? There are plenty of these in config_definition and I definitively don't want to wrap them into a function.

On the other hand, if you move it into the constructor or any other function it is not initialized before the first call in contrast to being initialized before _main_. So the only safe way is to wrap it in a static function and call this function instead of the global property.

neheb: Right.  It’s a bit annoying without std::span :).

PR #1572, opened 2021-06-17
string ref to string_view conversion

neheb:
Saves size. More efficient too.


KarlStraussberger:
I'm a bit nervous about string_view now. But as long as we do not store them, it should be fine.

PR #1573, opened 2021-06-18
Assign clientInfo in a default constructor

neheb:
Avoids some issue in GCC's -fanalyzer about a null pointer.

### DIFF
--- a/src/util/upnp_clients.cc
+++ b/src/util/upnp_clients.cc
@@ -33,9 +33,10 @@
 
 #include <upnp.h>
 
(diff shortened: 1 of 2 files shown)


KarlStraussberger:
Fine in this case. Need to find a strategy to deal with the other `static`s

neheb:
Note that fanalyzer only complains about this one.

KarlStraussberger:
Seems it only complains about complex types (like `Client`).  So no need to redo the other code 😌

PR #1576, opened 2021-06-18
remove nullptr checks and static members

neheb:
Not needed for std::unique_ptr.


KarlStraussberger:
That has an extreme runtime effect: setup is created for each call instead of once.

neheb:
OK.

PR #1578, opened 2021-06-18
add missing close

neheb:

### DIFF
--- a/src/util/tools.cc
+++ b/src/util/tools.cc
@@ -1202,6 +1202,8 @@ int find_local_port(in_port_t range_min, in_port_t range_max)
             return port;
         }
 
+        close(fd);
+
         retry_count++;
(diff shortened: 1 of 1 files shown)


src/util/tools.cc
```diff
@@ -1205,6 +1205,8 @@ int find_local_port(in_port_t range_min, in_port_t range_max)
         retry_count++;
     } while (retry_count < std::numeric_limits<uint16_t>::max());
 
+    close(fd);
```
KarlStraussberger, line 1208: In my opinion it can / should be placed in the end of the while loop if connect fails.

KarlStraussberger (changes requested):

src/util/tools.cc
neheb: Sounds right.

KarlStraussberger (approved):

PR #1580, opened 2021-06-18
default initialize some io stuff

neheb:

### DIFF
--- a/src/iohandler/io_handler_buffer_helper.cc
+++ b/src/iohandler/io_handler_buffer_helper.cc
@@ -37,27 +37,14 @@
 
 IOHandlerBufferHelper::IOHandlerBufferHelper(std::shared_ptr<Config> config, size_t bufSize, size_t initialFillSize)
     : config(std::move(config))
+    , bufSize(bufSize)
+    , initialFillSize(initialFillSize)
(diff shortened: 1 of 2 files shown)


src/iohandler/io_handler_buffer_helper.cc
```diff
@@ -43,21 +43,9 @@ IOHandlerBufferHelper::IOHandlerBufferHelper(std::shared_ptr<Config> config, siz
     if (initialFillSize > bufSize)
         throw_std_runtime_error("initialFillSize must be lesser than or equal to the size of the buffer");
 
-    this->bufSize = bufSize;
-    this->initialFillSize = initialFillSize;
+    bufSize = bufSize;
+    initialFillSize = initialFillSize;
```
KarlStraussberger, line 47: I'm not sure whether you can drop the `this->`. But you can move the init up to the init list and keep the checks. Makes it all more compact.

KarlStraussberger (changes requested):

src/iohandler/io_handler_buffer_helper.cc
neheb: Fixed.

KarlStraussberger (approved):

PR #1584, opened 2021-06-19
fix std::accumulate

neheb:
UBSan reports a shifting by int error. Looks like auto in this case
evaluates to int, which is not correct.


neheb:
Fixed slightly differently.

From cppref:

```
Common mistakes

If left to type inference, op operates on values of the same type as init which can result in unwanted casting of the iterator elements. For example, std::accumulate(v.begin(), v.end(), 0) likely does not give the result one wishes for when v is std::vector<double>. 
```

PR #1592, opened 2021-06-20
nppnp changes

neheb:
Get rid of the pointless ifdefs and just pass a C string to npupnp. It's
copy constructed anyway.

The overall goal is to reduce the NPUPNP diff to libupnp.


neheb:
Not enough to fix the memory leaks unfortunately.

PR #1596, opened 2021-06-21
clang-tidy: return a new vector

neheb:
clang-tidy complains to use emplace_back instead of push_back. Just
return a vector with an empty element.


whyman:
That surely has to be worse as now we are constructing a new vector?

neheb:
Hmm yeah never mind.

PR #1597, opened 2021-06-21
replace while loops with for loops

neheb:
Easier to read. Avoids setting pointers to the same value twice.

### DIFF
--- a/src/metadata/matroska_handler.cc
+++ b/src/metadata/matroska_handler.cc
@@ -146,24 +146,24 @@ void MatroskaHandler::parseMKV(const std::shared_ptr<CdsItem>& item, std::unique
     file_io_callback ebml_file(item->getLocation().c_str());
     EbmlStream ebml_stream(ebml_file);
 
-    auto el_l0 = ebml_stream.FindNextID(LIBMATROSKA_NAMESPACE::KaxSegment::ClassInfos, ~0);
-    while (el_l0 != nullptr) {
+    for (;;) {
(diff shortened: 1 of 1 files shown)


src/metadata/matroska_handler.cc
```diff
@@ -146,24 +146,24 @@ void MatroskaHandler::parseMKV(const std::shared_ptr<CdsItem>& item, std::unique
     file_io_callback ebml_file(item->getLocation().c_str());
     EbmlStream ebml_stream(ebml_file);
 
-    auto el_l0 = ebml_stream.FindNextID(LIBMATROSKA_NAMESPACE::KaxSegment::ClassInfos, ~0);
-    while (el_l0 != nullptr) {
+    for (;;) {
+        auto el_l0 = ebml_stream.FindNextID(LIBMATROSKA_NAMESPACE::KaxSegment::ClassInfos, ~0);
```
KarlStraussberger, line 150: what about
```
for (
    auto el_l0 = ebml_stream.FindNextID(LIBMATROSKA_NAMESPACE::KaxSegment::ClassInfos, ~0);
    el_l0 != nullptr;
    el_l0 = ebml_stream.FindNextElement(LIBMATROSKA_NAMESPACE::KaxSegment_Context, i_upper_level, ~0, true)
)
```
Seems a lot closer to the original logic: you replaced `FindNextElement` by `FindNextID`

KarlStraussberger (changes requested):

src/metadata/matroska_handler.cc
neheb: Looks uglier. Might as well keep the original loop.

PR #1598, opened 2021-06-21
small error handle

neheb:

### DIFF
--- a/src/transcoding/transcode_ext_handler.cc
+++ b/src/transcoding/transcode_ext_handler.cc
@@ -127,13 +127,17 @@ std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(std::shared_pt
             strcpy(fifo_template, "mt_transcode_XXXXXX");
             location = tempName(config->getOption(CFG_SERVER_TMPDIR), fifo_template);
             log_debug("creating reader fifo: {}", location.c_str());
-            if (mkfifo(location.c_str(), O_RDWR) == -1) {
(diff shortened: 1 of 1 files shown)


whyman:
I personally dislike doing work in if conditions.

Can we do the work, assign a `ret` value and then check it?

The compiler will do the right thing, and its much clearer to read.

neheb:
I mean, the surrounding does similar

whyman:
Fix that too while you are there? 😄

neheb:

src/transcoding/transcode_ext_handler.cc
```diff
@@ -127,13 +127,17 @@ std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(std::shared_pt
             strcpy(fifo_template, "mt_transcode_XXXXXX");
             location = tempName(config->getOption(CFG_SERVER_TMPDIR), fifo_template);
             log_debug("creating reader fifo: {}", location.c_str());
-            if (mkfifo(location.c_str(), O_RDWR) == -1) {
+            auto r = mkfifo(location.c_str(), O_RDWR);
+            if (r == -1) {
```
KarlStraussberger, line 131: I'd prefer `!= 0`. Looks more like error handling.

KarlStraussberger (changes requested):

src/transcoding/transcode_ext_handler.cc
neheb: OK

KarlStraussberger (approved):

PR #1601, opened 2021-06-22
Implemement dynamic containers

### LINKED ISSUE #171: FR: File date based sorting
Hi

I know most people will say that sorting should be the player's feature but in reality that does not work because most players do not  really offer sorting let alone file date sorting.

I used to use UMS and it offered sorting based on the file mod date which was very helpful because that way it was easy to find the lastly added media to the library otherwise one has to keep track of all those files mentally. 

Maybe this can be for meta lists like albums, all audio etc.

thanks

KarlStraussberger:
Allows virtual container that run a search.

fixes #171


whyman:
Nice!

We should add some of these to the default config, I think!

KarlStraussberger:
> 
> 
> Nice!
> 
> We should add some of these to the default config, I think!

Added two lines to config_generator. Still not default, means: old users don't get it.

PR #1602, opened 2021-06-22
switch several for loops to use size_t

neheb:
Array indeces are typically pointers, which are either 32 or 64 bit. Use
size_t to avoid bad code generation.

### DIFF
--- a/src/metadata/ffmpeg_handler.cc
+++ b/src/metadata/ffmpeg_handler.cc
@@ -211,7 +211,7 @@ void FfmpegHandler::addFfmpegResourceFields(const std::shared_ptr<CdsItem>& item
     // video resolution, audio sampling rate, nr of audio channels
     audioset = false;
     videoset = false;
-    for (unsigned int i = 0; i < pFormatCtx->nb_streams; i++) {
+    for (size_t i = 0; i < pFormatCtx->nb_streams; i++) {
         AVStream* st = pFormatCtx->streams[i];
         if ((st != nullptr) && (!videoset) && (as_codecpar(st)->codec_type == AVMEDIA_TYPE_VIDEO)) {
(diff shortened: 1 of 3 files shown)


whyman:
I agree with this idea.

Most of those changes aren't over arrays, and the ones that are are already unsigned.

Do you have an example of this "bad code" that is generated?

neheb:
I haven't looked at the gerbera code, but the basic idea is that with signed integers, the compiler needs to emit sign extension instructions. See https://youtu.be/jKpIZ4UcaNw?t=261 for an example.

Note that for range loops use size_t internally.

That video actually referenced a similar issue with the bzip2 source. Someone recently tested changing the loop variable (actually function parameter) from uint32_t to int32_t, and to size_t. The speed successively got better because of better assembly generation (the variable is used as an array index).

src/content/scripting/script.cc
```diff
@@ -228,7 +228,7 @@ Script::Script(std::shared_ptr<ContentManager> content,
     }
 
     duk_push_object(ctx); // config
-    for (int i = 0; i < int(CFG_MAX); i++) {
+    for (size_t i = 0; i < CFG_MAX; i++) {
         auto scs = ConfigDefinition::findConfigSetup(config_option_t(i), true);
```
KarlStraussberger, line 232: That one is based on an enum, so I think we should keep `int` instead.

neheb: It's cast to config_option_t anyway.

KarlStraussberger: That's why I'm actually thinking about writing an iterator over `config_option_t`. Furthermore: it is not used as an array index.

neheb: Removed.

PR #1605, opened 2021-06-22
replace C file stuff with fmt

neheb:
Simplifies the code and avoids static analyzer warnings.

Locally, the file seems to be created with 0600 permissions and because
of O_EXCL, it will fail if a file is present.

### DIFF
--- a/src/transcoding/transcode_ext_handler.cc
+++ b/src/transcoding/transcode_ext_handler.cc
@@ -127,13 +127,17 @@ std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(std::shared_pt
             strcpy(fifo_template, "mt_transcode_XXXXXX");
             location = tempName(config->getOption(CFG_SERVER_TMPDIR), fifo_template);
             log_debug("creating reader fifo: {}", location.c_str());
-            if (mkfifo(location.c_str(), O_RDWR) == -1) {
+            auto r = mkfifo(location.c_str(), O_RDWR);
+            if (r != 0) {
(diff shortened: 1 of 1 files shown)


neheb:
Maybe this should use ex somehow.

neheb:
```
   /home/runner/work/gerbera/gerbera/src/main.cc:44:10: fatal error: fmt/os.h: No such file or directory
     44 | #include <fmt/os.h>
```

orly?

neheb:
version minimum becomes 7. Unfortunate.

whyman:
There is also `writeTextFile` in `tools`

src/main.cc
```diff
@@ -327,48 +329,28 @@ int main(int argc, char** argv, char** envp)
...
 
-            // exit if the pidfile already exists
-            struct stat s;
-            errno = 0;
-            // for pidfile to be not there stat needs to fail and it needs
-            // to fail due to ENOENT (no such file or dir)
-            if (!(-1 == stat(pidfile->c_str(), &s) && errno == ENOENT)) {
+            try {
+                auto pidf = fmt::output_file(pidfile->c_str(), O_CREAT | O_EXCL | O_WRONLY);
+                // get the pid of our process
```
KarlStraussberger, line 334: Add your comment here `O_EXCL, it will fail if a file is present`

neheb:
Meh replaced with fopen instead. fmt 7.1 is a bit too new.

neheb:
fopen creates a 0644 file though...

neheb:
But so does open...

PR #1610, opened 2021-06-24
Implement EnumIterator

KarlStraussberger:
Looks nice to me because it encapsulates begin and end of enum.
Also casts from enum type to int and back are hidden and save because of to_underlying

### DIFF
--- a/src/config/config_setup.h
+++ b/src/config/config_setup.h
@@ -49,6 +49,8 @@ using DictionaryInitFunction = bool (*)(const pugi::xml_node& value, std::map<st
 using IntCheckFunction = bool (*)(int value);
 using IntMinFunction = bool (*)(int value, int minValue);
 
+using ConfigOptionIterator = EnumIterator<config_option_t, config_option_t::CFG_MIN, config_option_t::CFG_MAX>;
+
 class ConfigValue {
 public:
     std::string key;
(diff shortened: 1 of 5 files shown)


src/web/config_load.cc
```diff
@@ -162,10 +162,10 @@ void web::configLoad::process()
     }
 
     // write all values with simple type (string, int, bool)
-    for (int i = 0; i < int(CFG_MAX); i++) {
-        auto scs = ConfigDefinition::findConfigSetup(config_option_t(i));
+    for (auto&& i : ConfigOptionIterator()) {
```
neheb, line 165: maybe option instead of i?

src/config/config_setup.h
```diff
@@ -49,6 +49,8 @@ using DictionaryInitFunction = bool (*)(const pugi::xml_node& value, std::map<st
 using IntCheckFunction = bool (*)(int value);
 using IntMinFunction = bool (*)(int value, int minValue);
 
+typedef EnumIterator<config_option_t, config_option_t::CFG_MIN, config_option_t::CFG_MAX> ConfigOptionIterator;
```
neheb, line 52: using instead of typedef is better.

src/util/tools.h
```diff
@@ -247,6 +248,43 @@ std::string join(const C& container, const D& delimiter)
     return buf.str();
 }
 
+/// \brief Iterator over values of a sequential enum between begin and end
+template <typename C, C beginVal, C endVal>
+class EnumIterator {
+private:
+    typedef typename std::underlying_type<C>::type val_t;
```
neheb, line 255: using instead of typedef is better.

PR #1615, opened 2021-06-25
small lambda conversion

neheb:
Easier to read.


whyman:
Personally I think thats less readable :shrug: 

A simple if would be more obvious than the ternary.

neheb:
I think this is better.

PR #1619, opened 2021-06-25
switch to C++ ffmpegthumbnailer API

neheb:
Small simplification.


whyman:
Nice! Does it work?

KarlStraussberger:
no problems so far.

neheb:
> Nice! Does it work?

I haven't done through testing but it should work. The names are similar anyway. Except for the overlay strip thing. I had to look at the code to see how it was done.

neheb:
I wonder if that lock is needed anymore. The surrounding code looks like it could be cleaned up further. It looks like it's made to deal with C stuff.

PR #1628, opened 2021-06-27
default init clientInfo

neheb:
Avoids constructor shenanigans.


KarlStraussberger:
I'm not sure we should propagate this implementation detail to the header

Issue #1621, opened 2021-06-25
freebsd build with mysql enabled is broken


druban:
Not sure if it affects other builds, but this patch fixes it:
```
diff --git a/src/database/mysql/mysql_database.cc b/src/database/mysql/mysql_database.cc
index f845230a..14fdaf0d 100644
--- a/src/database/mysql/mysql_database.cc
+++ b/src/database/mysql/mysql_database.cc
@@ -32,6 +32,7 @@
 #ifdef HAVE_MYSQL
 #include "mysql_database.h"
 
+#include <array>
 #include <cstdlib>
 
 #include <netinet/in.h>
```

Otherwise build fails w/:
```
/home/dima/dev/gerbera/src/database/mysql/mysql_database.cc:103:31: error: implicit instantiation of undefined template 'std::__1::array<std::__1::vector<const char *, std::__1::allocator<const char *> >, 10>'
static const auto dbUpdates = std::array<std::vector<const char*>, 10> { {
                              ^
/usr/include/c++/v1/__tuple:219:64: note: template is declared here
template <class _Tp, size_t _Size> struct _LIBCPP_TEMPLATE_VIS array;
                                                               ^
[ 54%] Building CXX object CMakeFiles/libgerbera.dir/src/database/sql_database.cc.o
1 error generated.
```

KarlStraussberger:
Thank you.

PR #1633, opened 2021-06-27
SQLDatabase: Refactor init and upgrade

### LINKED ISSUE #1622: mysql.sql needs to reflect db version properly
mysql.sql has been update to new (11) layout, yet it still stores db_version as 10. Following patch fixes it:

```
diff --git a/src/database/mysql/mysql.sql b/src/database/mysql/mysql.sql
index 2a584d14..7d483402 100644
--- a/src/database/mysql/mysql.sql
+++ b/src/database/mysql/mysql.sql
@@ -46,7 +46,7 @@ CREATE TABLE `mt_internal_setting` (
   `value` varchar(255) NOT NULL,
   PRIMARY KEY  (`key`)
 ) ENGINE=MyISAM CHARSET=utf8;
-INSERT INTO `mt_internal_setting` VALUES ('db_version','10');
+INSERT INTO `mt_internal_setting` VALUES ('db_version','11');
 CREATE TABLE `mt_autoscan` (
   `id` int(11) NOT NULL auto_increment,
   `obj_id` int(11) default NULL,
```

KarlStraussberger:
fixes #1622

### DIFF
--- a/src/database/mysql/mysql.sql
+++ b/src/database/mysql/mysql.sql
@@ -46,7 +46,6 @@ CREATE TABLE `mt_internal_setting` (
   `value` varchar(255) NOT NULL,
   PRIMARY KEY  (`key`)
 ) ENGINE=MyISAM CHARSET=utf8;
-INSERT INTO `mt_internal_setting` VALUES ('db_version','10');
 CREATE TABLE `mt_autoscan` (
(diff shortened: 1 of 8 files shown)


KarlStraussberger:
Also fixes #1621

whyman:
Much better!

PR #1640, opened 2021-06-28
clang-tidy: remove implicit bool conversions

neheb:
Found with readability-implicit-bool-conversion

Manually edited smart pointers to use normal bool as it's properly
supported and a lot shorter.


whyman:
I like the `if (cs != nullptr)` to `if (cs)`.

But the rest of it makes it less clear to me. The implcit conversions are safe and easy to read.

neheb:
Rebased. Removed a bunch of 0 comparisons.

neheb:
This is not working right it seems. Probably made a mistake somewhere...

neheb:
Found the bug:

```
   2 --- a/src/metadata/taglib_handler.cc
   3 +++ b/src/metadata/taglib_handler.cc
   4 @@ -62,7 +62,7 @@ TagLibHandler::TagLibHandler(const std::shared_ptr<Context>& context)
   5
   6  void TagLibHandler::addField(metadata_fields_t field, const TagLib::File& file, const TagLib::Tag* tag, const std::shared_ptr<CdsItem>& item) const
   7  {
   8 -    if (tag == nullptr)
   9 +    if (!tag)
  10          return;
  11
  12      if (tag->isEmpty())
```

So this works perfectly fine with smart pointers as there's an explicit overload, but not with C pointers. Great...

neheb:
Splitting this up in parts just to be safe.

Issue #781, opened 2020-04-09
Request transcoding help: DTS to AC3 in .mkv


ras07:
My TV will play .mkv streams with AC3 audio, but not DTS. I've tried various transcoding options but I don't really know what I'm doing. I read the transcoding doc section but I'm afraid much of it went over my head. Can someone suggest edits to the config.xml file that will get me there?

pamapa:
To keep it simple i convert all videos with DTS before hand and do not need and transcoding options. This converts only the audio stream and leaves the image stream as is:
```
audio_opts="-c:a ac3" # convert all
ffmpeg -i "$input" \
   -map 0 \
   -c:v copy \
   $audio_opts \
   -c:s copy \
   "$output"
```

ras07:
Thanks - I'd sure love to be able to do this on the fly though, rather than either losing the original file or keeping two copies of everything. Can anyone point me in a direction that lets me use @pamapa 's ffmpeg parameters to accomplish transcoding?

jkiusAlarus:
Hello
you can maybe have a look to the issue https://github.com/gerbera/gerbera/issues/370 
there is a script that you can use to transcode you mkv using ffmpeg.
regards
Laurent

JoeNyland:
I have the same requirement, too: My TV will not play DTS from an .mkv container, but it will play AC3. I'd love to see some support in Gerbera for transcoding audio for this use case. The script linked above seems a bit too complicated and we'd lose the ability of seeking transcoded files.

jmgh77:
I am new to Gerbera and have the same issue with my LG WebOS TV (OLED65CX6LA with WebOS 5.0).  So I dig in the transcode option and put this in my config.xml file:

  <transcoding enabled="yes">
    <mimetype-profile-mappings>
      <transcode mimetype="video/x-matroska" using="nodts" />
    </mimetype-profile-mappings>

      <profile name="nodts" enabled="yes" type="external">
        <mimetype>audio/ac3</mimetype>
        <accept-url>no</accept-url>
        <first-resource>yes</first-resource>
        <hide-original-resource>no</hide-original-resource>
        <agent command="ffmpeg" arguments="-y -i %in -f ac3 %out" />
        <buffer size="1048576" chunk-size="131072" fill-size="262144" />
      </profile>

[...]

PR #1641, opened 2021-06-28
Add codec info to resource data

KarlStraussberger:
found this while searching for CaptionInfo.

Maybe does some trick to #781


whyman:
Nice

Koga-Miyata:
I installed and compiled Gerbera from source into a blank Debian Testing, to see if (at least my concerns) from #781 were fixed. For me on a LGC9 transcoding with settings suggested in #781 won't start. As far as I can tell the original file, not the transcode is beeing delivered.

KarlStraussberger:
I don't see any issues with transcoding. Can you try running `gerbera --debug` to see what happens on your box when you want to stream the movie.

PR #1644, opened 2021-06-29
Move db upgrade commands to config file

KarlStraussberger:
It's getting close to impossible to maintain all that code in source


src/database/mysql/mysql-upgrade.xml
```diff
@@ -0,0 +1,60 @@
...
+            `item_value` varchar(255) NOT NULL,
+            `status` varchar(20) NOT NULL)
+            ENGINE=MyISAM CHARSET=utf8
+        </script>
+        <script>CREATE INDEX grb_config_value_item ON grb_config_value(item)</script>
+    </version>
+    <version from="6" to="7">
+        <script>DROP TABLE mt_cds_active_item</script>
+    </version>
+    <version from="7" to="8" remark="part_number">
```
whyman, line 45: Maybe just version_number and work on the assumption you need apply previous versions? Every update we have is +1 anyway.

whyman:
Something to consider is that this lets someone with write access, and the ability to restart the process to run arbitrary SQL as the gerbera user. Slightly more risky than the create which is only called on an empty DB.

Is there anything simple we can do to mitigate that risk? Permissions check maybe? Is it worth the hassle? 🤔

src/database/mysql/mysql-upgrade.xml
KarlStraussberger: It's documentary anyway

KarlStraussberger:
> 
> 
> Something to consider is that this lets someone with write access, and the ability to restart the process to run arbitrary SQL as the gerbera user. Slightly more risky than the create which is only called on an empty DB.
> 
> Is there anything simple we can do to mitigate that risk? Permissions check maybe? Is it worth the hassle? 🤔

Added some hashes for the sections in XML file. So in order to update the file you need to run gerbera to get the hash in the warning and then enter the hash in the code.

whyman:
Nice!

I guess `hashies` should be `hashes`? (if you mean a selection of hash items)

KarlStraussberger:
> 
> 
> Nice!
> 
> I guess `hashies` should be `hashes`? (if you mean a selection of hash items)

I wanted it to sound like a nickname :smile:

src/util/tools.h
```diff
@@ -276,7 +276,7 @@ class EnumIterator {
     EnumIterator begin() { return *this; }
     EnumIterator end()
     {
-        static const EnumIterator endIter = ++EnumIterator(endVal);
+        static const EnumIterator endIter = EnumIterator(endVal);
```
KarlStraussberger, line 279: Founf that when testing: Threw an exception when loading config page

src/web/config_load.cc
```diff
@@ -59,7 +59,7 @@ void web::configLoad::addTypeMeta(pugi::xml_node& meta, const std::shared_ptr<Co
 {
     auto info = meta.append_child("item");
     info.append_attribute("item") = cs->getUniquePath().c_str();
-    info.append_attribute("id") = fmt::to_string(cs->option).c_str();
+    info.append_attribute("id") = fmt::format("{:03d}", cs->option).c_str();
```
KarlStraussberger, line 62: Regression: Broke array values on config page

src/web/config_load.cc
```diff
@@ -163,14 +163,15 @@ void web::configLoad::process()
...
         try {
+            auto scs = ConfigDefinition::findConfigSetup(option);
+            auto item = values.append_child("item");
+            createItem(item, scs->getItemPath(-1), option, option, scs);
+
             log_debug("    Option {:03d} {} = {}", option, scs->getItemPath(), scs->getCurrentValue().c_str());
             setValue(item, scs->getCurrentValue());
         } catch (const std::runtime_error& e) {
+            log_warning("Option {:03d} {}", option, e.what());
         }
```
KarlStraussberger, line 175: Avoid exceptions if more issues in config setup are introduced

src/upnp_xml.cc
```diff
@@ -441,6 +441,7 @@ bool UpnpXMLBuilder::renderSubtitle(const std::string& virtualURL, const std::sh
     for (auto&& res : item->getResources()) {
         if (res->isMetaResource(VIDEO_SUB, CH_SUBTITLE)) {
             url = renderOneResource(virtualURL, item, res, realCount);
+            url.append(renderExtension("", res->getAttribute(R_RESOURCE_FILE)));
```
KarlStraussberger, line 444: Make header identical to resource

PR #1652, opened 2021-07-03
use make_pair in std::array

neheb:
Allows using CTAD and getting rid of extra {}.

CTAD is a good idea. The first entry explains why. The array was updated at some point but the size was not. That can cause issues when accessing entries outside of the array.


KarlStraussberger:
I prefer having the enum types in array declaration. Otherwise you always have to look up the exact types in order to understand the code. In cases when you add enum props you cannot grep for the type to check the usages.

neheb:
I don't agree but added anyway.

neheb:
added some extra ones.

neheb:
switched to std::pair. It's the exact same as make_pair turns out.

PR #1655, opened 2021-07-03
replace return nullptr with return {}

neheb:
Equivalent and shorter.


whyman:
I prefer the nullptr version, what do you think @KarlStraussberger ?

KarlStraussberger:
> I prefer the nullptr version, what do you think @KarlStraussberger ?

Ditto

PR #1657, opened 2021-07-04
remove std from std::next

neheb:
Redundant.


KarlStraussberger:
I'd keep it because we always use namespaces. Avoids compiltion issues in exotic environments.

neheb:
Right but in this case, it's quarenteed to work because of ADL: https://en.cppreference.com/w/cpp/language/adl

Note that all of these are in accumulate calls.

PR #1660, opened 2021-07-06
remove std from move in several places

neheb:
Because of ADL, it's implicit


whyman:
I personally prefer keeping it, its more human friendly IMO. The machine understands either way.

KarlStraussberger:
> 
> 
> I personally prefer keeping it, its more human friendly IMO. The machine understands either way.

It's the same with me. I don't want to think or try whether it works with or without, I just add it and leave it to the compiler to do its work.

PR #1665, opened 2021-07-11
several string_view conversions

neheb:
Format fixes were also done along the way.

### DIFF
--- a/src/database/search_handler.h
+++ b/src/database/search_handler.h
@@ -429,7 +429,6 @@ protected:
     std::shared_ptr<ASTNode> parseRelationshipExpression();
     std::shared_ptr<ASTNode> parseParenthesis();
     std::shared_ptr<ASTQuotedString> parseQuotedString();
-    void checkIsExpected(TokenType tokenType, const std::string& tokenTypeDescription);
 
 private:
     std::shared_ptr<SearchToken> currentToken;
--- a/src/metadata/taglib_handler.cc
+++ b/src/metadata/taglib_handler.cc
@@ -264,11 +264,11 @@ void TagLibHandler::fillMetadata(std::shared_ptr<CdsObject> obj)
     log_debug("TagLib handler done.");
 }
 
-bool TagLibHandler::isValidArtworkContentType(const std::string& art_mimetype)
+bool TagLibHandler::isValidArtworkContentType(std::string_view art_mimetype)
 {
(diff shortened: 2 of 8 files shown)


src/metadata/matroska_handler.cc
```diff
@@ -256,11 +256,11 @@ std::string MatroskaHandler::getContentTypeFromByteVector(const LIBMATROSKA_NAME
 #endif
 }
 
-void MatroskaHandler::addArtworkResource(const std::shared_ptr<CdsItem>& item, const std::string& art_mimetype)
+void MatroskaHandler::addArtworkResource(const std::shared_ptr<CdsItem>& item, std::string_view art_mimetype)
 {
```
KarlStraussberger, line 260: This change produces an extra object because its used in call to renderProtocolInfo as string&

src/database/search_handler.h
```diff
@@ -429,7 +429,7 @@ class SearchParser {
     std::shared_ptr<ASTNode> parseRelationshipExpression();
     std::shared_ptr<ASTNode> parseParenthesis();
     std::shared_ptr<ASTQuotedString> parseQuotedString();
-    void checkIsExpected(TokenType tokenType, const std::string& tokenTypeDescription);
+    void checkIsExpected(TokenType tokenType, std::string_view tokenTypeDescription);
```
KarlStraussberger, line 433: That method is not called anywhere. Drop it.

src/metadata/taglib_handler.cc
```diff
@@ -281,11 +281,11 @@ std::string TagLibHandler::getContentTypeFromByteVector(const TagLib::ByteVector
 #endif
 }
 
-void TagLibHandler::addArtworkResource(const std::shared_ptr<CdsItem>& item, const std::string& art_mimetype)
+void TagLibHandler::addArtworkResource(const std::shared_ptr<CdsItem>& item, std::string_view art_mimetype)
 {
```
KarlStraussberger, line 285: Same call to renderProtocolInfo here

KarlStraussberger (changes requested):

src/metadata/matroska_handler.cc
neheb: Removed.

src/database/search_handler.h
neheb: Removed.

KarlStraussberger (approved):

PR #1668, opened 2021-07-13
string_view conversions

neheb:


KarlStraussberger:
All places convert from string to string_view. string_view is helpful if you want to apply comparison to char* objects, otherwise it's an extra object created.

PR #1673, opened 2021-07-18
small cleanups

neheb:
not sure about the dynamic_cast commit.


neheb:
lint seems to be failing on unrelated stuff I think? I can't reproduce locally.

PR #1675, opened 2021-07-18
move make_shared outside of initializer list

neheb:

### DIFF
--- a/src/config/client_config.cc
+++ b/src/config/client_config.cc
@@ -30,32 +30,35 @@
 #include "util/upnp_quirks.h"
 
 ClientConfig::ClientConfig()
-    : clientInfo(std::make_shared<ClientInfo>())
 {
-    clientInfo->name = "Unknown";
-    clientInfo->type = ClientType::Unknown;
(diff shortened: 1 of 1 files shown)


src/config/client_config.cc
```diff
@@ -30,32 +30,35 @@
...
-    clientInfo->matchType = ClientMatchType::None;
-    clientInfo->match = "";
+    auto cInfo = ClientInfo {
+        "Unknown",
+        ClientType::Unknown,
+        QUIRK_FLAG_NONE,
+        ClientMatchType::None,
+        "",
+    };
+    clientInfo = std::make_shared<ClientInfo>(cInfo);
```
KarlStraussberger, line 41: What about a `move` here to avoid copying cInfo?

neheb: I think I tried this and found that adding std::move made no difference to the compiled size. Will need to look at it again.

neheb: Yeah no difference to size. Added move anyway. clang-tidy doesn't complain so it should be fine.

PR #1679, opened 2021-07-18
lambda conversion

neheb:
Slightly shorter and smaller size.


neheb:
haha that CodeQL warning is bogus.

neheb:
Oh I see. clang is smart enough to figure it out. GCC is not.

PR #1682, opened 2021-07-19
Factor out transactions to reduce overhead when disabled

KarlStraussberger:


whyman:
At some point I will get some time to look at transactions 😞

KarlStraussberger:
No worries. I moved it aside now 😢 altough I like the idea of protecting changes in a transaction.

Mostly it's colliding with multithreading: If we run transactions, all operations must be in sequence which costs the time we win by running writing operations in memory (part of the transaction) before writing everything to the disk.

I did not check its possibility, but we may open a reading and a writing connection in parallel. Reading done as ever, writing using transactions. Only getting into trouble when reading while writing.

PR #1687, opened 2021-07-21
change db variable to pointer

neheb:
Allows removing a const_cast and a bunch of address-of operators.


KarlStraussberger:
If you don't initalize the pointer with new we will get a SEGFAULT

neheb:
hmm good point. It's currently initialized to nullptr.

PR #1689, opened 2021-07-21
move make_unique down

neheb:
Saves some size.


KarlStraussberger:
You create an extra copy of the vector. There should be a `move` to avoid this

neheb:
Added move. It seems to work here.

PR #1695, opened 2021-07-22
use string_view for upnp_headers

neheb:
They're copied to an std::string at the onset.


KarlStraussberger:
It's fine with the key which is a `#define "..."`. Value is string in code, so your converting down.

whyman:
I still get worried about all usage of `string_view` without `length`

PR #1701, opened 2021-07-24
Allow setting resource order

### LINKED ISSUE #875: Inconsistent thumbnail / Fanart for video
aka VLC always shows ffmpeg_thumbnailer previews instead of fanart from folder.jpg.

Gerbera generates 2 items for a video file with `folder.jpg` in the same folder:
```xml
<upnp:albumArtURI xmlns:dlna="urn:schemas-dlna-org:metadata-1-0" dlna:profileID="JPEG_TN">http://10.0.0.4:49152/content/media/object_id/62/res_id/1/rct/aa</upnp:albumArtURI>
<res protocolInfo="http-get:*:image/jpeg:DLNA.ORG_PN=JPEG_TN;DLNA.ORG_OP=01;DLNA.ORG_CI=0" resolution="128x85">http://10.0.0.4:49152/content/media/object_id/62/res_id/2/rh/6/ext/file.jpg</res>
```

Where the URL of the first one corresponds to `fanart_handler` (as I understand via `CH_FANART` stored in the DB). The second one is ResourceHandler(rh)==CH_FFTH(6).

When VLC gets this it first initializes its `psz_albumArt` field with the value from `upnp:albumArtURI` element and then **overwrites** it with URL from each resource which is image (i.e. `protocolInfo` starts with `http-get:*:image/`).

Of course this could be considered a bug in VLC (as if it should prefer albumArtURI), but having one logical thing implemented in 2 ways is not good either.

To me the most logical seems to skip ffmpeg_thumbnailer part for objects which already have fanart. Maybe make it customizable.

### LINKED ISSUE #1201: [Question] HEVC decoding on live for old TVs
Hey guys, you did great job with Gerbera, thank you !
I have LG Smart tv that can play media files through default player. Natively it doesn't support HEVC files.
Is there any possibility to somehow decode on the fly on server side HEVC file and transform it to H.264 ? So TV can play it as it was H.264 file ?

### LINKED ISSUE #1645: AVC
Hello

Trying to transcode in order to see some mkv in my ps3, could not understand how to configure the profile section using video/avc mimetype. On PS3 it shows "unkown data"

Using video/mpeg and ffmpeg mpeg2 it worked. 

Any suggestions ?

Regards

KarlStraussberger:
fixes #875

- Allow client specific transcoding

closes #1645
closes #1201


src/upnp_xml.cc
```diff
@@ -424,33 +431,30 @@ std::string UpnpXMLBuilder::renderOneResource(const std::string& virtualURL, con
 bool UpnpXMLBuilder::renderItemImage(const std::string& virtualURL, const std::shared_ptr<CdsItem>& item, std::string& url)
 {
     bool artAdded = false;
-    int realCount = 0;
-    for (auto&& res : item->getResources()) {
+    auto orderedResources = getOrderedResources(item);
+    for (auto&& res : orderedResources) {
```
neheb, line 435: This looks like it can be converted to std::any_of

src/upnp_xml.cc
```diff
@@ -424,33 +431,30 @@ std::string UpnpXMLBuilder::renderOneResource(const std::string& virtualURL, con
...
-        realCount++;
     }
     return artAdded;
 }
 
 bool UpnpXMLBuilder::renderSubtitle(const std::string& virtualURL, const std::shared_ptr<CdsItem>& item, std::string& url)
 {
     bool srtAdded = false;
-    int realCount = 0;
     for (auto&& res : item->getResources()) {
```
neheb, line 451: Same here

src/upnp_xml.cc
```diff
@@ -473,6 +477,26 @@ std::string UpnpXMLBuilder::renderExtension(const std::string& contentType, cons
...
+    std::vector<std::shared_ptr<CdsResource>> orderedResources {};
+    auto&& resources = item->getResources();
+    for (auto&& oh : orderedHandler) {
+        std::for_each(resources.begin(), resources.end(), [&](auto res) { if (oh == res->getHandlerType()) orderedResources.push_back(res); });
+    }
+
+    // Append resources not listed in orderedHandler
+    for (auto&& res : resources) {
+        auto ch = res->getHandlerType();
+        auto chIndex = std::find_if(orderedHandler.begin(), orderedHandler.end(), [ch](auto&& entry) { return ch == entry; });
```
neheb, line 492: std::none_of seems better here.

src/upnp_xml.cc
KarlStraussberger: I don't see how any_of can work here, find_if does

src/upnp_xml.cc
KarlStraussberger: Thank you

src/upnp_xml.cc
neheb: Yeah, my bad. Need the iterator.

src/upnp_xml.cc
```diff
@@ -473,6 +477,26 @@ std::string UpnpXMLBuilder::renderExtension(const std::string& contentType, cons
     return "";
 }
 
+std::vector<std::shared_ptr<CdsResource>> UpnpXMLBuilder::getOrderedResources(const std::shared_ptr<CdsItem>& item)
+{
+    // Order resources according to index defined by orderedHandler
+    std::vector<std::shared_ptr<CdsResource>> orderedResources {};
```
neheb, line 483: This is somewhat of a minor point, but there's no need to use {} with vector like this.

whyman:
Allow client specific transcoding! 👏

whyman:
How do you use it?

KarlStraussberger:
Just add the `client-flags` attribute to the transcoding `profile`. If your client is not SAMSUNG you may invent a flag, manually assign it in `<clients>` config, e.g. `0x100` and pick it up here.

PR #1704, opened 2021-07-25
use merge directly

neheb:
This is C++17 now.


KarlStraussberger:
```
Attempts to extract ("splice") each element in source and insert it into *this using the comparison object of *this. If there is an element in *this with key equivalent to the key of an element from source, then that element is not extracted from source. No elements are copied or moved, only the internal pointers of the container nodes are repointed. All pointers and references to the transferred elements remain valid, but now refer into *this, not into source. 
```

So this change is breaking

PR #1705, opened 2021-07-25
use map.emplace

neheb:
allows using std::move.

### DIFF
--- a/src/cds_resource.cc
+++ b/src/cds_resource.cc
@@ -69,14 +69,14 @@ void CdsResource::mergeAttributes(const std::map<std::string, std::string>& addi
     }
 }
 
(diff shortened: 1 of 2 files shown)


KarlStraussberger:
How does emplace behave if an entry is already there?

neheb:
From cppref:

```
The element may be constructed even if there already is an element with the key in the container, in which case the newly constructed element will be destroyed immediately. 
```

KarlStraussberger:
This also is a breaking change!

whyman:
I am confused how it makes sense to emplace if you are "moving"? There is no construction going on with a move?

neheb:
Replaced with insert_or_assign.

PR #1706, opened 2021-07-26
remove pointless {}

neheb:
These are default constructed.

### DIFF
--- a/src/cds_objects.h
+++ b/src/cds_objects.h
@@ -103,7 +103,7 @@ protected:
     std::chrono::seconds utime {};
 
     /// \brief File size on disk (in bytes).
-    off_t sizeOnDisk { 0 };
+    off_t sizeOnDisk {};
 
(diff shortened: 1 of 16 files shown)


neheb:
std::chrono stuff must remain as it. Tested with godbolt.

src/util/jpeg_resolution.cc
```diff
@@ -76,7 +76,7 @@ static int Get16m(const void* Short)
 
 static int ioh_fgetc(const std::unique_ptr<IOHandler>& ioh)
 {
-    uchar c[1] = { 0 };
+    uchar c[1] = {};
```
KarlStraussberger, line 79: This sets the first entry to 0. I'd keep that for clarity.

neheb: It sets both to 0. Changed anyway.

PR #1708, opened 2021-07-27
replace reference parameter with std::pair

neheb:
Simpler and allows to use structured bindings.

### DIFF
--- a/src/util/upnp_quirks.cc
+++ b/src/util/upnp_quirks.cc
@@ -56,8 +56,8 @@ void Quirks::addCaptionInfo(const std::shared_ptr<CdsItem>& item, const std::uni
     if (!startswith(item->getMimeType(), "video"))
         return;
 
-    std::string url;
-    if (UpnpXMLBuilder::renderSubtitle(context->getServer()->getVirtualUrl(), item, url)) {
+    auto [url, subAdded] = UpnpXMLBuilder::renderSubtitle(context->getServer()->getVirtualUrl(), item);
+    if (subAdded) {
         headers->addHeader("CaptionInfo.sec", url);
     }
 }
--- a/src/web/containers.cc
+++ b/src/web/containers.cc
@@ -74,8 +74,8 @@ void web::containers::process()
         int autoscanType = cont->getAutoscanType();
         ce.append_attribute("autoscan_type") = mapAutoscanType(autoscanType).data();
 
-        std::string url;
-        if (UpnpXMLBuilder::renderContainerImage(server->getVirtualUrl(), cont, url)) {
+        auto [url, artAdded] = UpnpXMLBuilder::renderContainerImage(server->getVirtualUrl(), cont);
+        if (artAdded) {
             ce.append_attribute("image") = url.c_str();
         }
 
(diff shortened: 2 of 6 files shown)


neheb:
Actually... std::string can just be used.

neheb:
Hmm that's probably not correct.

src/util/upnp_quirks.cc
```diff
@@ -56,8 +56,8 @@ void Quirks::addCaptionInfo(const std::shared_ptr<CdsItem>& item, const std::uni
     if (!startswith(item->getMimeType(), "video"))
         return;
 
-    std::string url;
-    if (UpnpXMLBuilder::renderSubtitle(context->getServer()->getVirtualUrl(), item, url)) {
+    auto [url, r] = UpnpXMLBuilder::renderSubtitle(context->getServer()->getVirtualUrl(), item);
```
KarlStraussberger, line 59: Please use `subAdded` instead of single character var

src/web/containers.cc
```diff
@@ -74,8 +74,8 @@ void web::containers::process()
         int autoscanType = cont->getAutoscanType();
         ce.append_attribute("autoscan_type") = mapAutoscanType(autoscanType).data();
 
-        std::string url;
-        if (UpnpXMLBuilder::renderContainerImage(server->getVirtualUrl(), cont, url)) {
+        auto [url, r] = UpnpXMLBuilder::renderContainerImage(server->getVirtualUrl(), cont);
```
KarlStraussberger, line 77: Same here

src/web/edit_load.cc
```diff
@@ -195,8 +195,8 @@ void web::edit_load::process()
         mimeType.append_attribute("value") = objItem->getMimeType().c_str();
         mimeType.append_attribute("editable") = true;
 
-        std::string url;
-        if (UpnpXMLBuilder::renderItemImage(server->getVirtualUrl(), objItem, url)) {
+        auto [url, r] = UpnpXMLBuilder::renderItemImage(server->getVirtualUrl(), objItem);
```
KarlStraussberger, line 198: Same

src/web/items.cc
```diff
@@ -133,8 +133,8 @@ void web::items::process()
         std::string res = UpnpXMLBuilder::getFirstResourcePath(objItem);
         item.append_child("res").append_child(pugi::node_pcdata).set_value(res.c_str());
 
-        std::string url;
-        if (UpnpXMLBuilder::renderItemImage(server->getVirtualUrl(), objItem, url)) {
+        auto [url, r] = UpnpXMLBuilder::renderItemImage(server->getVirtualUrl(), objItem);
```
KarlStraussberger, line 136: Same

neheb:

neheb:
To clarify, my original idea was to check if string is not empty, but not sure if that would break things.

KarlStraussberger:
It's fine the way it is now. Never rely on particular values for control - separation of concerns

neheb:
Interestingly enough, both commits have the same compiled size. Guess gcc sees through this and optimizes it.

KarlStraussberger:
> 
> 
> Interestingly enough, both commits have the same compiled size. Guess gcc sees through this and optimizes it.

Sometimes it's just by accident. Logic and readability should always dominate over compiled sized, especially if we consider different compilers on different platforms. Our _work_ is logic - theirs to make small code out of it.

PR #1710, opened 2021-07-28
getTimespecAfterMillis: return ret instead of void

neheb:
This function needed to be like this when using timespec. No longer.

### DIFF
--- a/src/content/update_manager.cc
+++ b/src/content/update_manager.cc
@@ -177,9 +177,6 @@ void UpdateManager::threadProc()
             }
             bool sendUpdates = true;
             if (sleepMillis >= MIN_SLEEP && objectIDHash->size() < MAX_OBJECT_IDS) {
-                std::chrono::milliseconds timeout;
-                getTimespecAfterMillis(sleepMillis, timeout);
-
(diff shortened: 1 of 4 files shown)


src/content/update_manager.cc
```diff
@@ -177,8 +177,7 @@ void UpdateManager::threadProc()
             }
             bool sendUpdates = true;
             if (sleepMillis >= MIN_SLEEP && objectIDHash->size() < MAX_OBJECT_IDS) {
-                std::chrono::milliseconds timeout;
-                getTimespecAfterMillis(sleepMillis, timeout);
+                auto timeout = getTimespecAfterMillis(sleepMillis);
```
KarlStraussberger, line 181: Seems this call is not needed altogether.

neheb: Funny. Now this functions is only used in one location. Might make sense to remove and just move it to updateNextNotify

KarlStraussberger: Looking at the complexity of it -> do it

neheb:
Removed.

PR #1711, opened 2021-07-29
unique_ptr removals

neheb:


neheb:
What a strange clang error

```
/usr/home/build/gerbera/src/database/sql_database.cc:693:20: error: chosen constructor is explicit in copy-initialization
            return {};
                   ^~
/usr/include/c++/v1/sstream:765:14: note: explicit constructor declared here
    explicit basic_ostringstream(ios_base::openmode __wch = ios_base::out);
```

Weird that it picks the wrong constructor.

src/database/sql_database.cc
```diff
@@ -2555,10 +2555,10 @@ std::unique_ptr<std::ostringstream> SQLDatabase::sqlForInsert(const std::shared_
     std::ostringstream qb;
     qb << "INSERT INTO " << TQ(tableName) << " (" << fields.str() << ") VALUES (" << values.str() << ')';
 
-    return std::make_unique<std::ostringstream>(std::move(qb));
+    return qb;
 }
```
KarlStraussberger, line 2559: Isn't the stream copied here? What happens to its contents?

Maybe resolve by calling `.str()` and making the whole thing return `std::string` (I don't like the streams!)

src/util/upnp_headers.cc
```diff
@@ -124,6 +124,6 @@ std::unique_ptr<std::map<std::string, std::string>> Headers::readHeaders(UpnpFil
         ret.insert(add);
     }
 
-    return std::make_unique<std::map<std::string, std::string>>(std::move(ret));
+    return ret;
 #endif
```
KarlStraussberger, line 128: We should keep the unique_ptr here to avoid copying the map.

src/database/sql_database.cc
neheb: No this is an implicit move as the types match.

src/content/autoscan_inotify.h
```diff
@@ -164,7 +164,7 @@ class AutoscanInotify {
         int wd;
     };
 
-    std::unique_ptr<std::unordered_map<int, std::shared_ptr<Wd>>> watches;
+    std::unordered_map<int, std::shared_ptr<Wd>> watches;
```
KarlStraussberger, line 168: The argument (0) should be moved here (`std::make_unique<std::unordered_map<int, std::shared_ptr<Wd>>>(0)`)

KarlStraussberger (changes requested):
I like most of the changes because the reduce complexity.

src/content/autoscan_inotify.h
neheb: IIRC this was added by me to avoid an fanalyzer false positive. 0 is the default value.

src/util/upnp_headers.cc
neheb: Removed.

neheb:
Removed the readHeaders patch.

KarlStraussberger (approved):

PR #1712, opened 2021-07-29
Change generation of SQL statements to format

KarlStraussberger:


whyman:
Im not a fan of the ostringstreams either.

I think a bit of `fmt::join` might help here to remove the other loops?

src/database/sql_database.cc
```diff
@@ -680,7 +680,7 @@ void SQLDatabase::updateObject(const std::shared_ptr<CdsObject>& obj, int* chang
 
     beginTransaction("updateObject");
     for (auto&& addUpdateTable : data) {
-        auto qb = [this, &obj, &addUpdateTable]() -> std::unique_ptr<std::ostringstream> {
+        auto qb = [this, &obj, &addUpdateTable]() {
```
neheb, line 683: Is this needed without unique_ptr?

KarlStraussberger: It's just a string now!

KarlStraussberger:
> 
> 
> Im not a fan of the ostringstreams either.
> 
> I think a bit of `fmt::join` might help here to remove the other loops?

Thanks for the hint. I'll look into it!

src/database/sql_database.cc
neheb: I meant "this" in the capture list.

src/database/sql_database.cc
```diff
@@ -285,64 +310,45 @@ void SQLDatabase::init()
...
-        buf << " ";
-        this->sql_browse_query = buf.str();
+        auto join1 = fmt::format("{0}{2}{1} {0}{3}{1} ON {4} = {0}{3}{1}.{0}{5}{1}", table_quote_begin, table_quote_end, CDS_OBJECT_TABLE, REF_ALIAS, browseColumnMapper->mapQuoted(BrowseCol::ref_id), browseColMap.at(BrowseCol::id).second);
+        auto join2 = fmt::format("{} ON {} = {}", asColumnMapper->tableQuoted(), asColumnMapper->mapQuoted(AutoscanCol::obj_id), browseColumnMapper->mapQuoted(BrowseCol::id));
+        this->sql_browse_query = fmt::format("SELECT {} FROM {} LEFT JOIN {} LEFT JOIN {} ", fmt::join(buf, ", "), browseColumnMapper->tableQuoted(), join1, join2);
     }
     // Statement for UPnP search
     {
-        std::ostringstream colBuf;
+        std::vector<std::string> colBuf;
```
neheb, line 327: reserve should be called

src/database/sql_database.cc
```diff
@@ -2526,39 +2526,32 @@ void SQLDatabase::generateResourceDBOperations(const std::shared_ptr<CdsObject>&
...
 {
     std::string tableName = addUpdateTable->getTableName();
     auto dict = addUpdateTable->getDict();
 
-    std::ostringstream fields;
-    std::ostringstream values;
+    std::vector<std::string> fields;
+    std::vector<std::string> values;
 
     for (auto it = dict.begin(); it != dict.end(); it++) {
```
neheb, line 2509: Range loop

src/database/sql_database.cc
```diff
@@ -2526,39 +2526,32 @@ void SQLDatabase::generateResourceDBOperations(const std::shared_ptr<CdsObject>&
...
-std::ostringstream SQLDatabase::sqlForInsert(const std::shared_ptr<CdsObject>& obj, const std::shared_ptr<AddUpdateTable>& addUpdateTable) const
+std::string SQLDatabase::sqlForInsert(const std::shared_ptr<CdsObject>& obj, const std::shared_ptr<AddUpdateTable>& addUpdateTable) const
 {
     std::string tableName = addUpdateTable->getTableName();
     auto dict = addUpdateTable->getDict();
 
-    std::ostringstream fields;
-    std::ostringstream values;
+    std::vector<std::string> fields;
+    std::vector<std::string> values;
```
neheb, line 2507: reserve should be called

src/database/sql_database.cc
```diff
@@ -1041,18 +1032,18 @@ std::shared_ptr<CdsObject> SQLDatabase::findObjectByPath(fs::path fullpath, bool
     else
         dbLocation = addLocationPrefix(LOC_DIR_PREFIX, fullpath);
 
-    std::ostringstream qb;
-    qb << sql_browse_query
-       << " WHERE " << TQBM(BrowseCol::location_hash) << '=' << quote(stringHash(dbLocation))
-       << " AND " << TQBM(BrowseCol::location) << '=' << quote(dbLocation)
-       << " AND " << TQBM(BrowseCol::ref_id) << " IS NULL"
-                                                " LIMIT 1";
+    std::vector<std::string> where {
```
neheb, line 1035: I could be wrong but this could probably be replaced with

```
auto where = std::array {
```

src/database/sql_database.cc
```diff
@@ -1032,7 +1031,7 @@ std::shared_ptr<CdsObject> SQLDatabase::findObjectByPath(fs::path fullpath, bool
     else
         dbLocation = addLocationPrefix(LOC_DIR_PREFIX, fullpath);
 
-    std::vector<std::string> where {
+    auto where = std::array {
```
neheb, line 1034: Local testing shows this to increase size rather than reduce it...

I guess std::array mostly makes sense in constexpr context.

neheb:
somewhat related:

src/util/tools.h contains a join template that can be replaced with fmt::join. Saves some size.

KarlStraussberger:
I suppose the in the end that join can be deleted including some others.

neheb:
Implemented on top of this PR: https://github.com/neheb/gerbera/commit/21ddf58ad25f2da846e90d10a556c4e470ce4dd7

KarlStraussberger:
I'll pull this into my change... once done here

src/database/sql_database.cc
```diff
@@ -2448,8 +2387,8 @@ std::string SQLDatabase::sqlForUpdate(const std::shared_ptr<CdsObject>& obj, con
     auto dict = addUpdateTable->getDict();
 
     std::vector<std::string> fields;
-    for (auto it = dict.begin(); it != dict.end(); it++) {
-        fields.push_back(fmt::format("{0}{2}{1} = {3}", table_quote_begin, table_quote_end, it->first, it->second));
+    for (auto&& [field, value] : dict) {
```
neheb, line 2380: clang-tidy complains of a missing reserve.

KarlStraussberger:
Finally! SqlDatabase is stream free!

And some other places as well

src/util/tools.h
```diff
@@ -34,11 +34,7 @@
 #include <chrono>
 #include <filesystem>
 #include <map>
-#include <memory>
-#include <optional>
```
neheb, line 38: This is still needed.

```
src/util/tools.h:162:6: error: no template named 'optional' in namespace 'std'
std::optional<std::vector<std::byte>> readBinaryFile(const fs::path& path);
```

KarlStraussberger: Seems to heavily depend on the compiler. Readded

KarlStraussberger:
Ran my import twice tonight and got similar import times than before. No obvious things broken, though. But also no significant speed improvement (seems there is nothing left to improve)

whyman:
The ostreams are ugly, but they are fast when used correctly. So nice to see we havent lost much from the fmt changes.

We should work on some benchmark tests. Argh! I need more time 😿

KarlStraussberger:
> We should work on some benchmark tests. Argh! I need more time 😿

Once found, send it over 💯

KarlStraussberger:
> The ostreams are ugly, but they are fast when used correctly. So nice to see we havent lost much from the fmt changes.

That's what I wanted to see last night. To h... with the work on the weekend if performance goes there. But I was lucky to see no lack. Furthermore there are several places to optimize now (resizes etc.)

PR #1718, opened 2021-07-30
use auto instead of const char*

neheb:
Too much const otherwise.

### DIFF
--- a/src/util/tools.cc
+++ b/src/util/tools.cc
@@ -75,7 +75,7 @@
 
 #define WHITE_SPACE " \t\r\n"
 
-static constexpr const char* HEX_CHARS = "0123456789abcdef";
+static constexpr auto HEX_CHARS = "0123456789abcdef";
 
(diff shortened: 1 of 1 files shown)


whyman:
I like the types. auto makes you think more 🙃

KarlStraussberger:
Me too, as long as isn't obvious - like with the char here. Now you can even read it as string.

whyman:
Yeah nothing against this change particularly, just as a decidedly part time C++ user it adds to the complexity

neheb:
It's perfectly fine for simple types and right side type assignment. Also if you don't care about the type of something.

But otherwise yes.

KarlStraussberger:
> Yeah nothing against this change particularly, just as a decidedly part time C++ user it adds to the complexity
> 
> Same with the automatic move magic. C++ is complicated 😅😄😬

That's what keeps our little grey things below the hair alive

PR #1723, opened 2021-07-30
remove const and use move for map

neheb:


src/metadata/libexif_handler.cc
```diff
@@ -165,7 +165,7 @@ static const std::map<std::string, int> exifTagMap {
 
 static int getTagFromString(const std::string& tag)
 {
-    auto result = getValueOrDefault(exifTagMap, tag, -1);
+    auto result = getValueOrDefault(std::move(exifTagMap), tag, -1);
```
KarlStraussberger, line 168: Are you sure this works. From my understanding only for thr first call.

PR #1729, opened 2021-08-01
move unique_ptr removals

neheb:
Some background:

Because of -fanalyzer warnings, I changed these make_unique calls such that the function creates the vector or whatever and then moves it into make_unique. But doing that is quite silly as it's more efficient return the object. NRVO is prevented when make_unique is used. Compiled binary sees a small size reduction locally.

Runtime test seems to be fine as well.

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -694,8 +694,7 @@ void ContentManager::_rescanDirectory(const std::shared_ptr<AutoscanDirectory>& 
         if (!asSetting.followSymlinks && dirEnt.is_symlink()) {
             int objectID = database->findObjectIDByPath(newPath);
             if (objectID > 0) {
-                if (list)
-                    list->erase(objectID);
+                list.erase(objectID);
                 removeObject(adir, objectID, false);
             }
             log_debug("link {} skipped", newPath.c_str());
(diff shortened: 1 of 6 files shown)


KarlStraussberger:
The point is, that now the vector is copied instead. This has a runtime effect.

matthiasstraka:
I recently came across the same line and would have suggested the same modifications. 

> The point is, that now the vector is copied instead. This has a runtime effect.

Returning a vector<int> directly requires no copy at all. The old implementation with make_unique<vector>(local_vec) does indeed make a copy (only make_unique<>(std::move(local_vec)) would eliminate the vector copy but require additional memory allocation for the unique_ptr).

The generated code difference can be seen [here](https://godbolt.org/z/7T5YaGhax).

neheb:
@matthiasstraka Thanks for making a godbolt example.

I was confused as to how this was a copy. NVRO optimizes the copy away. According to cppref, it's a move when optimization is disabled.

https://en.cppreference.com/w/cpp/language/copy_elision

neheb:
Rebased.

also note that these were found with

```
git grep make_unique\<std
```

I trust the std stuff to do the right thing.

PR #1734, opened 2021-08-02
final unique_ptr removals

neheb:
With these,

```
git grep make_unique\<std src
```

returns 0.


whyman:
I dont understand why this is needed?

Are we not now copying the map? And we have lost the ownership schematics.

neheb:
As stated in the previous PR (with a godbilt example), no. unique_ptr prevents NVRO aka copy elision.

neheb:
It simplifies the code. The result ends up the same anyway.

PR #1735, opened 2021-08-02
move orderedHandler out of the class

neheb:
It's not really needed.


whyman:
If it were me I would make it a regular member, is it static just to avoid loading the config multiple times @KarlStraussberger ?

KarlStraussberger:
It's static because `UpnpXMLBuilder` instance is not available in `upnp_quirks` and web pages. We could propagate the object there and remove the static - which would be great!

KarlStraussberger:
PR #1762  clarifies what the code should be

PR #1739, opened 2021-08-02
increase code readabiliy, preallocate vectors when possible

matthiasstraka:
The database code can be reduced in several places, especially row unique_ptrs only make sense to live inside the body of if/while statements.
Result vectors can and should be preallocated when possible using `getNumRows()` to avoid repeated reallocation.
In `SQLDatabase::getPathIDs`, the `LIMIT 1` clause has no effect since `WHERE` selects a unique id anyway.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -601,8 +601,6 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
         cdsObjectSql["mime_type"] = quote(item->getMimeType().substr(0, 40));
     }
 
-    std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> returnVal;
-
     // check for a duplicate (virtual) object
     if (hasReference && op != Operation::Update) {
         std::ostringstream q;
@@ -618,7 +616,7 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
         auto res = select(q);
         // if duplicate items is found - ignore
         if (res && (res->nextRow()))
-            return returnVal;
+            return {};
     }
 
     if (obj->getParentID() == INVALID_OBJECT_ID) {
@@ -626,7 +624,10 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
     }
 
     cdsObjectSql["parent_id"] = fmt::to_string(obj->getParentID());
-    returnVal.push_back(std::make_shared<AddUpdateTable>(CDS_OBJECT_TABLE, cdsObjectSql, op));
+
+    std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> returnVal = {
+        std::make_shared<AddUpdateTable>(CDS_OBJECT_TABLE, cdsObjectSql, op)
+    };
 
     if (!hasReference || obj->getMetadata() != refObj->getMetadata()) {
         generateMetadataDBOperations(obj, op, returnVal);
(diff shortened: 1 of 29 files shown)


src/database/sql_database.cc
```diff
@@ -618,15 +616,18 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
...
     }
 
     if (obj->getParentID() == INVALID_OBJECT_ID) {
         throw_std_runtime_error("Tried to create or update an object {} with an illegal parent id {}", obj->getLocation().c_str(), obj->getParentID());
     }
 
     cdsObjectSql["parent_id"] = fmt::to_string(obj->getParentID());
-    returnVal.push_back(std::make_shared<AddUpdateTable>(CDS_OBJECT_TABLE, cdsObjectSql, op));
+
+    std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> returnVal = {
```
neheb, line 628: = { not needed.

Alternatively, you could rewrite to

```
auto returnVal = std::vector {

CTAD will take care of the template argument.
```

src/database/sql_database.cc
```diff
@@ -1471,16 +1459,16 @@ std::string SQLDatabase::incrementUpdateIDs(const std::unordered_set<int>& ids)
...
+    rows.reserve(res->getNumRows());
+    while (auto row = res->nextRow()) {
         std::ostringstream s;
         s << row->col(0) << ',' << row->col(1);
         rows.emplace_back(s.str());
     }
 
     if (rows.empty())
-        return "";
+        return {};
```
neheb, line 1471: I'd add a newline.

matthiasstraka: Why a newline? We want to return an empty string (i.e. default constructed)

neheb: No I mean between both return statements.

matthiasstraka: Am I seeing the wrong line which you are talking about? There in only one modified return.

neheb:
```
        return {};
<-------newline here
    return join(rows, ",");
```

matthiasstraka: I only saw that after switching to full file view. Yes, a new line would look better. I can improve this tomorrow.

KarlStraussberger:
Thank you for the contribution.

Please be aware that there is a PR pending for database code. So some of your changes will have to be reworked.

src/database/sql_database.cc
```diff
@@ -855,12 +851,9 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::browse(const std::unique_pt
...
 
     std::vector<std::shared_ptr<CdsObject>> result;
-
-    std::unique_ptr<SQLRow> row;
-    while ((row = sqlResult->nextRow())) {
-        auto obj = createObjectFromRow(row);
-        result.push_back(obj);
-        row = nullptr; // clear out content from unique_ptr
+    result.reserve(sqlResult->getNumRows());
+    while (auto row = sqlResult->nextRow()) {
```
KarlStraussberger, line 855: For readability we prefer not to mix assignments with checks and leave optimization to the compiler.

KarlStraussberger (changes requested):

src/database/sql_database.cc
matthiasstraka: It's now fixed wherever suitable

matthiasstraka:
> Please be aware that there is a PR pending for database code. So some of your changes will have to be reworked.
I'll review the code after the other PR is merged

src/database/sql_database.cc
```diff
@@ -1012,7 +1021,8 @@ std::vector<std::string> SQLDatabase::getMimeTypes()
 
     std::vector<std::string> arr;
     arr.reserve(res->getNumRows());
-    while (auto row = res->nextRow()) {
+    std::unique_ptr<SQLRow> row;
+    while (row = res->nextRow()) {
```
neheb, line 1025: This will definitely cause compiler warnings as extra parentheses are needed.

neheb:
I’d rebase and squash the last two commits into the first as you’re undoing your own changes.

matthiasstraka:
I think the current state of my PR has exceeded my knowledge of git. I will close it and restart with more precise commits (cherry picked by topic) and without the need for merges.

PR #1740, opened 2021-08-02
replace std::vector with std::list

neheb:
It seems entries get erased in various spots.

There's a small size reduction. No idea why.


neheb:
Hmmm vector is not that slow though...

whyman:
Vectors are faster than the linked lists most of the time because of cache locality.

PR #1744, opened 2021-08-03
use newer spdlog and fmt for debian

KarlStraussberger:
Support for `fmt::join` seems broken in spdlog 1.8.1


whyman:
I guess we should update the minimums in the CMake/tables in the docs

KarlStraussberger:
> 
> 
> I guess we should update the minimums in the CMake/tables in the docs

Good point, if that merge goes green

PR #1746, opened 2021-08-03
Update required versions for spdlog and libfmt

KarlStraussberger:


cmake/GenCompileInfo.cmake
```diff
@@ -52,6 +52,7 @@ function(generate_compile_info)
                 OUTPUT_STRIP_TRAILING_WHITESPACE
         )
     else()
+        set(GIT_TAG "")
```
KarlStraussberger, line 55: This did not fix the funny GERBERA_VERSION on freebsd (1.3.5-*)

whyman: Hmmm, I think that is likely a CI issue

PR #1747, opened 2021-08-03
Code refactoring and performance enhancements in sql_database

matthiasstraka:
This is essentially the same PR as [PR 1739](https://github.com/gerbera/gerbera/pull/1739) but with cleaner commits grouped by function. All previous review remarks are already included in these commits.


matthiasstraka:
One change I did not yet make but which I would recommend: collapse `table_quote_begin` and `table_quote_end` to `table_quote` which would reduce the SQL generation code significantly (especially now that fmt is used). Currently, begin is the same character as end for both MySQL and SQLite.

KarlStraussberger (commented):
Thanks for helping to improve gerbera. I found one point which could have negative effect on preformance

src/database/sql_database.cc
```diff
@@ -2214,13 +2217,13 @@ std::vector<int> SQLDatabase::getPathIDs(int objectID)
     std::vector<int> pathIDs;
     while (objectID != CDS_ID_ROOT) {
         pathIDs.push_back(objectID);
-        auto res = select(fmt::format("{} {} LIMIT 1", sel, quote(objectID)));
+        auto res = select(fmt::format("{} {}", sel, quote(objectID)));
```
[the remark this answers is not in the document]
KarlStraussberger: > That LIMIT should stay because the result list may be long and we do only need the first

If you readd that `LIMIT` the PR could be merged.

src/database/sql_database.cc
```diff
@@ -1518,11 +1519,12 @@ std::unique_ptr<Database::ChangedContainers> SQLDatabase::removeObjects(const st
     std::vector<int32_t> containers;
     std::unique_ptr<SQLRow> row;
     while ((row = res->nextRow())) {
-        int objectType = std::stoi(row->col(1));
+        const int32_t objectID = row->col_int(0);
+        const int objectType =   row->col_int(1);
```
KarlStraussberger, line 1523: There is a double space here

src/database/sql_database.cc
```diff
@@ -1672,24 +1672,24 @@ std::unique_ptr<Database::ChangedContainers> SQLDatabase::_recursiveRemove(
                 throw DatabaseException("", fmt::format("Sql error: {}", sql));
             containerIds.clear();
             while ((row = res->nextRow())) {
-                int objectType = std::stoi(row->col(1));
+                const int objectType = row->col_int(1);
```
KarlStraussberger, line 1675: Add `const int objId = row->col_int(0);` and use it in the following lines;

KarlStraussberger (changes requested):

KarlStraussberger:
If it wasn't for the double spaces there is no need for a change, but if you have to change just make it even better

matthiasstraka:
That should be all of the changes (I messed up an attempted squash but that should not be an issue I hope)

KarlStraussberger (approved):

KarlStraussberger:
As there have been changes in the master branch in the meantime I cannot automatically rebase. Either you rebase yourself or I squash merge.

matthiasstraka:
If possible, I'd like you to squash merge my commits together. My fork is currently in a somewhat messed-up state which might cause problems during a rebase.

KarlStraussberger:
Thank you

PR #1754, opened 2021-08-04
replace several inserts with std::copy

neheb:
For whatever reason, the latter is more efficient.

Maybe because std::copy internally is implemented as a push_back loop
and insert is done in some other way.

### DIFF
--- a/src/config/config_setup.cc
+++ b/src/config/config_setup.cc
@@ -625,7 +625,7 @@ std::vector<std::string> ConfigArraySetup::getXmlContent(const pugi::xml_node& o
     if (result.empty()) {
         log_debug("{} assigning {} default values", xpath, defaultEntries.size());
         useDefault = true;
-        result.assign(defaultEntries.begin(), defaultEntries.end());
+        result = defaultEntries;
     }
     if (notEmpty && result.empty()) {
         throw_std_runtime_error("Invalid array {} empty '{}'", xpath, optValue);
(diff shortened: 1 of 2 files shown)


src/config/config_setup.cc
```diff
@@ -625,9 +625,9 @@ std::vector<std::string> ConfigArraySetup::getXmlContent(const pugi::xml_node& o
     if (result.empty()) {
         log_debug("{} assigning {} default values", xpath, defaultEntries.size());
         useDefault = true;
-        result.assign(defaultEntries.begin(), defaultEntries.end());
+        return defaultEntries;
     }
-    if (notEmpty && result.empty()) {
+    if (notEmpty) {
```
KarlStraussberger, line 630: This changes the logic if defaultEntries are not set!

src/config/config_setup.cc
```diff
@@ -831,11 +831,9 @@ std::map<std::string, std::string> ConfigDictionarySetup::getXmlContent(const pu
     if (result.empty()) {
         log_debug("{} assigning {} default values", xpath, defaultEntries.size());
         useDefault = true;
-        for (auto&& entry : defaultEntries) {
-            result.insert(entry); // this should be an insert without a loop, but Debian 10 doesn't compile.
-        }
+        return defaultEntries;
     }
-    if (notEmpty && result.empty()) {
+    if (notEmpty) {
```
KarlStraussberger, line 836: Logic changed here as well

KarlStraussberger (changes requested):

src/config/config_setup.cc
neheb: Hmm I just assumed defaultEntries was some static vector or something. I'll revert.

neheb:
Fixed.

PR #1755, opened 2021-08-04
prefer xml-node children over xpath

matthiasstraka:
pugi XML select_nodes evaluates an xpath expression while children(name) simply uses a filtered iterator, which is faster.
Fixed minor pointer move and explicit emplace (since the if branch is only executed when dynId is not in dynamicContainers)


KarlStraussberger:
Thank you.

PR #1756, opened 2021-08-04
add missing move

neheb:


neheb:
removing the variable has the same effect.

PR #1758, opened 2021-08-05
{} conversions

neheb:


src/content/content_manager.cc
```diff
@@ -1121,7 +1121,7 @@ std::pair<int, bool> ContentManager::addContainerTree(const std::vector<std::sha
         update_manager->containerChanged(result);
         session_manager->containerChangedUI(result);
     }
-    return std::pair(result, isNew);
+    return { result, isNew };
```
KarlStraussberger, line 1124: Two months ago, you replaced `{ result, isNew }` by `std::make_pair` then by `std::pair`. Now your turning it back. That keeps us busy but does not produce any progress. What's your strategy here? What's the benefit?

neheb: So I run coverity on this codebase from time to time. It complains here that the functions has no return statement, which is a false positive. I think it gets confused by the loop above. My strategy originally was to use std::make_pair and std::pair but that didn't fix it. Meaning the original {} return was just fine.

PR #1760, opened 2021-08-05
libfmt formatters for quoted SQL identifiers

matthiasstraka:
I very much like the new fmt-based SQL statement generation code - but the format strings for quoting literals make the statements hard to read.

So I came up with [custom formatters](https://fmt.dev/latest/api.html#formatting-user-defined-types) for a SQLIdentifier struct that encapsulates the identifier name and quotation-traits and allows for a very readable quoted name formatting. One drawdown is that all identifiers need to be outside the SQL template.

There is only one usage example in this PR which I'd like to discuss before changing additional code (also ignore code formatting for now). Is this this a good idea?

On the other hand, why is the current code so pedantic on quoting all identifiers while both MySQL and SQLite should be able to handle all used identifiers without quotes.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -344,7 +344,7 @@ void SQLDatabase::init()
     for (auto&& resAttrId : ResourceAttributeIterator()) {
         auto attrName = MetadataHandler::getResAttrName(resAttrId);
         resourceTagMap.emplace_back(fmt::format("res@{}", attrName), to_underlying(ResourceCol::attributes) + resAttrId);
-        resourceColMap[to_underlying(ResourceCol::attributes) + resAttrId] = { RES_ALIAS, attrName };
+        resourceColMap[to_underlying(ResourceCol::attributes) + resAttrId] = std::pair(RES_ALIAS, attrName);
     }
 
     browseColumnMapper = std::make_shared<EnumColumnMapper<BrowseCol>>(table_quote_begin, table_quote_end, ITM_ALIAS, CDS_OBJECT_TABLE, browseSortMap, browseColMap);
@@ -359,9 +359,10 @@ void SQLDatabase::init()
         std::vector<std::string> buf;
         buf.reserve(browseColMap.size());
         for (auto&& [key, col] : browseColMap) {
-            buf.push_back(fmt::format("{0}{2}{1}.{0}{3}{1}", table_quote_begin, table_quote_end, col.first, col.second));
+            buf.push_back(fmt::format("{}.{}", identifier(col.first), identifier(col.second)));
         }
-        auto join1 = fmt::format("LEFT JOIN {0}{2}{1} {0}{3}{1} ON {4} = {0}{3}{1}.{0}{5}{1}", table_quote_begin, table_quote_end, CDS_OBJECT_TABLE, REF_ALIAS, browseColumnMapper->mapQuoted(BrowseCol::ref_id), browseColMap.at(BrowseCol::id).second);
+        auto join1 = fmt::format("LEFT JOIN {0} {1} ON {2} = {1}.{3}",
+            identifier(CDS_OBJECT_TABLE), identifier(REF_ALIAS), browseColumnMapper->mapQuoted(BrowseCol::ref_id), identifier(browseColMap.at(BrowseCol::id).second));
         auto join2 = fmt::format("LEFT JOIN {} ON {} = {}", asColumnMapper->tableQuoted(), asColumnMapper->mapQuoted(AutoscanCol::obj_id), browseColumnMapper->mapQuoted(BrowseCol::id));
         this->sql_browse_query = fmt::format("SELECT {} FROM {} {} {} ", fmt::join(buf, ", "), browseColumnMapper->tableQuoted(), join1, join2);
     }
@@ -370,7 +371,7 @@ void SQLDatabase::init()
         std::vector<std::string> colBuf;
         colBuf.reserve(searchColMap.size());
         for (auto&& [key, col] : searchColMap) {
-            colBuf.push_back(fmt::format("{0}{2}{1}.{0}{3}{1}", table_quote_begin, table_quote_end, col.first, col.second));
+            colBuf.push_back(fmt::format("{}.{}", identifier(col.first), identifier(col.second)));
         }
         this->sql_search_columns = fmt::format("{}", fmt::join(colBuf, ", "));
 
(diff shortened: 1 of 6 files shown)


src/database/sql_format.h
```diff
@@ -0,0 +1,60 @@
+/*GRB*
+  Gerbera - https://gerbera.io/
+
+  sql_format.h - this file is part of Gerbera.
+
+  Copyright (C) 2018-2021 Gerbera Contributors
```
whyman, line 6: This should just be 2021 as its new code

whyman:
This is a nice concept.

If each DB were updated to provide an implementation of the interface then we avoid a separate traits object.

matthiasstraka:
> If each DB were updated to provide an implementation of the interface then we avoid a separate traits object.

This is my idea also. Maybe we can use a more generic traits class outside SQLIdentifier which is filled with various database-related capabilities that can be queried. The current code-base has a wide variety of quoting code which should be unified in the long-run.

In the end, I can image that the final version looks something like
`auto t = getSQLTraits();`
`fmt::format("SELECT {} FROM {} WHERE {} = {}", SQLIdentifier('colname', t), SQLIdentifier('table', t), SQLIdentifier("id", t), SQLLiteral(123, t))`
I'll keep looking into this and will update this PR as I progress.

KarlStraussberger:
I like the idea a lot.

The only thing I'm concerned is performance! How much time do we loose?
I was in the same position with the format change: Change it all and then drop it because it is incredibly slow.

The quoting was in MediaTomb already and there are some cases where it is saver (e.g. the resource columns can contain ":") so it's wiser to keep it.

Maybe you can make some methods for traits `t.Identifier('colname')` ---> `Identifier(identifier, traits) { return SQLIdentifier(identifier, traits); }`

matthiasstraka:
Performance-wise I would love to see the sql code base to use prepared statements (at least for commonly executed scripts). I did a trace while browsing my files using MariaDB and there are a lot of repeated constant statements that would benefit greatly.

KarlStraussberger:
There is no fixed codebase because the statement contain IDs which differ for each database. The core of the problem is ContentManager which runs through the whole hierarchy most of the time. I know some really crazy back and forth with IDs instead of caching some objects. But that's stuff for longer winter nights.

src/database/sql_database.h
```diff
@@ -172,6 +173,9 @@ class SQLDatabase : public Database {
     /// \brief migrate resources from mt_cds_objects to grb_resource before removing the column (DBVERSION 13)
     bool doResourceMigration();
     void migrateResources(int objectId, const std::string& resourcesStr);
+    
+    /// \brief returns a fmt-prinable identifier name
+    SQLIdentifier identifier(const std::string_view& name) const { return SQLIdentifier(name, table_quote_begin); }
```
KarlStraussberger, line 178: You should add `table_quote_end` also, e.g. M$SqlServer uses `[...]`.

KarlStraussberger (changes requested):

KarlStraussberger:
I recommend running `git rebase` to reduce the change set.

src/database/sql_database.h
matthiasstraka: I now use both quotes again. Thanks for the mssql reminder. I like the way the current state of the code looks and will transform the remaining statements as I find the time to do so.

src/database/sql_format.h
```diff
@@ -0,0 +1,60 @@
...
+{
+    struct Traits
+    {
+        Traits(char quote) : quote_begin(quote), quote_end(quote) {}
+        Traits(char quote_begin, char quote_end) : quote_begin(quote_begin), quote_end(quote_end) {}
+        char quote_begin;
+        char quote_end;
+    };
+
+    SQLIdentifier(std::string_view&& name, const Traits& traits)
```
neheb: AFAIK string_view is trivially copyable. Why &&?

src/database/sql_format.h
```diff
@@ -0,0 +1,60 @@
...
+
+  $Id$
+*/
+
+/// \file sql_format.h
+
+#ifndef __SQL_FORMAT_H__
+#define __SQL_FORMAT_H__
+
+#include <fmt/format.h>
```
neheb, line 28: I think this should be fmt/core.h

src/database/sql_database.h
```diff
@@ -188,6 +189,9 @@ class SQLDatabase : public Database {
...
```
neheb: there's no need to take a string_view by reference. It's cheap to copy.

matthiasstraka: This is already fixed in a previous commit

src/database/sql_format.h
matthiasstraka: This is already fixed in a previous commit

src/database/sql_format.h
matthiasstraka: Fix applied locally, will push when there are more changes. Thanks for pointing it out.

src/database/sql_database.cc
```diff
@@ -1143,13 +1147,13 @@ int SQLDatabase::createContainer(int parentID, std::string name, const std::stri
     std::string dbLocation = addLocationPrefix((isVirtual ? LOC_VIRT_PREFIX : LOC_DIR_PREFIX), virtualPath);
 
     auto fields = std::vector {
```
KarlStraussberger, line 1149: I'm not sure whether `fmt::join` works with `SQLIdentifier`. But in case, you could make the fields vector a `vector<SQLIdentifier>` and updated the insert method.

KarlStraussberger (commented):
It's looking really great and makes the code a lot more readable.

src/database/sql_database.cc
matthiasstraka: I'm just getting back to where I was before my vaccation - and I am pretty sure that I successfully tested this before. Today, I managed to get a gerbera building outside Docker so I can properly debug now. This will speed up things, especially when doing more risky changes. There are still many places left where I want to apply this identifier(...) method.

KarlStraussberger:
I don't want to put pressure on you, but it's number one candidate to make 1.9.1 ready 😄

matthiasstraka:
I was not expecting this to be fully ready so soon. But we can always go for an initial merge and I can continue later on (probably a good idea to avoid conflicts). I hope I can find some time tomorrow to review my current PR and un-draft it when I'm pleased.
1.9.1 is important since it fixes the issue with mysql databases which got me stated in custom-builing and contributing to the project 😊

KarlStraussberger (approved):

KarlStraussberger:
Thank you for that work. I've ran a full import into my library and did not see any performance issue.

PR #1765, opened 2021-08-06
remove some double pointer magic

neheb:
Return the value instead.

### DIFF
--- a/src/util/upnp_clients.cc
+++ b/src/util/upnp_clients.cc
@@ -141,7 +141,7 @@ void Clients::addClientByDiscovery(const struct sockaddr_storage* addr, const st
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


whyman:
Are you not creating a pointer to an object allocated on the stack and then returning it? What happens to the object then?

I guess RVO is saving you here, because in my head this shouldnt work :shrug:

neheb:
I guess so. But neither coverity nor compilers are complaining.

neheb:
Just saw this pattern used in another codebase. No issues there.

KarlStraussberger:
The semantics of the former code is to leave `pClientInfo` untouched if nothing was found. With your change it is set to `nullptr`.

I'd love to keep this if we need to call `getInfo` for a second time outside the constructor.

PR #1766, opened 2021-08-06
return by value

neheb:
The loop can be converted to reference.


neheb:
This was a bad change. I converted the vector to a unique_ptr and found that it was being copied anyway.

PR #1767, opened 2021-08-07
remove shared_ptr from vector

neheb:
Same size.


whyman:
If we remove the shared pointer and babe a method to return to the vector then who owns it?

neheb:
converted to unique_ptr instead.

whyman:
Thanks!

PR #1771, opened 2021-08-09
Add M3U8 support

FlashSystems:
The M3U8 format allows special characters within the file names and descriptions. It uses an UTF8 BOM to indicate the byte order. This commit strips the BOM in common.js to make sure the rest of the file is interpreted correctly. Without this change the BOM will be part of the first line of the playlist and lead to an error message or a missing playlist entry.

The core of this commit is stripping a BOM from the first line of the playlist if it exists and mapping the m3u8 extension to `audio/x-mpegurl`. I use foobar2000 to generate the playlists and some of my filenames contain characters that can not be encoded correctly within a m3u playlist.

I opted to create a new test for m3u8 playlists that is almost a copy of the m3u case. I opted for a dedicated test because the it's a different file format and to make this commit cleaner. Maybe this test could also be rolled into the m3u test. But my ctest knowledge is too limited to say anything about that.

### DIFF
--- a/scripts/js/common.js
+++ b/scripts/js/common.js
@@ -876,6 +876,11 @@ function readM3uPlaylist(playlist_title, playlistChain, playlistDirChain) {
     var line = readln();
     var playlistOrder = 1;
 
+    // Remove a BOM if there is one
+    if (line.charCodeAt(0) === 0xFEFF) {
+        line = line.substr(1);
+    }
+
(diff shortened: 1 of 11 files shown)


KarlStraussberger:
Don't worry about duplicating tests. Each tests has to be responsible for exactly one case.

Thank you for your contribution. Great job.

PR #1772, opened 2021-08-10
convert Quirks to unique_ptr

neheb:
Avoids copies.


whyman:
Avoids copies how?

whyman:
I actually think removing this pointer could work here, assuming those are the only accesses.

Assuming that all get calls are const refs, it shoud be fine.

neheb:
unique_ptr has a deleted copy constructor/operator.

shared_ptr is a lot more complex than just allowing copies. It also does ref counting, which is pointless in this case.

whyman:
Huh?

neheb:
Tried removing the unique_ptr, too many build failures.

It's easier to keep as is.

PR #1773, opened 2021-08-10
remove single argument std::string

neheb:
There's no point.

### DIFF
--- a/src/web/config_load.cc
+++ b/src/web/config_load.cc
@@ -306,7 +306,7 @@ void web::configLoad::process()
 
         item = values.append_child("item");
         createItem(item, cs->getItemPath(pr, ATTR_TRANSCODING_PROFILES_PROFLE, ATTR_TRANSCODING_PROFILES_PROFLE_TYPE), cs->option, ATTR_TRANSCODING_PROFILES_PROFLE_TYPE);
-        setValue(item, std::string((entry->getType() == TR_External) ? "external" : "none"));
+        setValue(item, (entry->getType() == TR_External ? "external" : "none"));
 
         item = values.append_child("item");
         createItem(item, cs->getItemPath(pr, ATTR_TRANSCODING_PROFILES_PROFLE, ATTR_TRANSCODING_PROFILES_PROFLE_MIMETYPE), cs->option, ATTR_TRANSCODING_PROFILES_PROFLE_MIMETYPE);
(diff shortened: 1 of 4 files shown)


src/web/config_load.cc
```diff
@@ -306,7 +306,7 @@ void web::configLoad::process()
 
         item = values.append_child("item");
         createItem(item, cs->getItemPath(pr, ATTR_TRANSCODING_PROFILES_PROFLE, ATTR_TRANSCODING_PROFILES_PROFLE_TYPE), cs->option, ATTR_TRANSCODING_PROFILES_PROFLE_TYPE);
-        setValue(item, std::string((entry->getType() == TR_External) ? "external" : "none"));
+        setValue(item, (entry->getType() == TR_External ? "external" : "none"));
```
KarlStraussberger, line 309: What version of `setValue` is called now? `string_view` or `string`?

neheb: Most likely string_view.

KarlStraussberger: So without extra code cost you could add `std::string_view` instead of `std::string` that we do not call path-version by accident

neheb: Wait where's the string_view overload? I only see

```
src/config/config_generator.h:    std::shared_ptr<pugi::xml_node> setValue(const std::string& tag, const std::string& value = "", bool makeLastChild = false);
src/config/config_generator.h:    std::shared_ptr<pugi::xml_node> setValue(config_option_t option, const std::string& value = "");
src/config/config_generator.h:    std::shared_ptr<pugi::xml_node> setValue(config_option_t option, config_option_t attr, const std::string& value);
src/config/config_generator.h:    std::shared_ptr<pugi::xml_node> setValue(config_option_t option, const std::string& key, const std::string& value);
src/config/config_generator.h:    std::shared_ptr<pugi::xml_node> setValue(config_option_t option, config_option_t dict, config_option_t attr, const std::string& value);
src/config/config_generator.h:    static std::shared_ptr<pugi::xml_node> setValue(const std::shared_ptr<pugi::xml_node>& parent, config_option_t option, const std::string& value);
src/config/config_options.h:    void setValue(size_t keyIndex, const std::string& value);
src/web/pages.h:    static void setValue(pugi::xml_node& item, const T& value);
```

neheb: after https://github.com/gerbera/gerbera/pull/1777 this will call std::string.

PR #1777, opened 2021-08-10
simplify setValue template

neheb:
if constexpr can be used here.

fs::path was kept as it can be implemented as std::wstring.

### DIFF
--- a/src/web/config_load.cc
+++ b/src/web/config_load.cc
@@ -85,26 +85,10 @@ void Web::ConfigLoad::createItem(pugi::xml_node& item, const std::string& name, 
 template <typename T>
 void Web::ConfigLoad::setValue(pugi::xml_node& item, const T& value)
 {
-    static_assert(fmt::has_formatter<T, fmt::format_context>::value, "T must be formattable");
-    item.append_attribute("value") = fmt::to_string(value).c_str();
-}
-
-template <>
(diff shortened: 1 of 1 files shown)


neheb:
I'm unsure about the first commit.

So as written, it's compatible with std::string and std::string_view. Although passing std::string_view by const ref is a bit weird.

There's also the issue of fs::path. The if statement can be simplified by just calling c_str, which would break std::string_view(not a big deal) and fs::path that uses std::wstring.

KarlStraussberger:
The idea behind these templates was to let the compiler select the correct option and avoid cascated `if`s at runtime!

neheb:
constexpr ifs are special. They are also evaluated at compile time. It's the same code generated as before.

KarlStraussberger:
But the code looks a lot more complicated and requires a lot more attention.

neheb:
OK, removed. Now just keeping the first commit. I assume that's not complicated?

neheb:
Rebased to bump the CI.

neheb:
Rebased again.

neheb:
Anything missing?

KarlStraussberger:
My conviction that the new version is really better. It look to C-ish to me. And I'm afraid that for future argument types the if statement gets longer and longer.
Sorry.

PR #1779, opened 2021-08-11
fmt: reduce minimum version to 6.1.2

neheb:
This is what is used on Debian 10.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,7 +396,7 @@ else()
     target_link_libraries(libgerbera PUBLIC UPNP::UPNP)
 endif()
 
-find_package(fmt 7.1.3 REQUIRED)
+find_package(fmt 6.1.2 REQUIRED)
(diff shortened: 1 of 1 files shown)


whyman:
I think spdlog 1.8.5 requires 7.1.3. hence the version being set.

neheb:
spdlog 1.8.5 doesn't seem to be a requirement. In any case, Debian 10 compiles with this commit.

whyman:
Its 1.8.2 in the CMake, but 1.8.5 in the README, typo I guess.

KarlStraussberger:
I changed CMake to 1.8.2 because alpine is using that - even though the version check did not catch for whatever reason.

Most build scripts install 1.8.5 to ensure the correct build settings with fmtlib (which is also installed). I'd love to keep the newer version requirements because we do not know how the older versions change runtime behaviour.

PR #1787, opened 2021-08-12
more move with push_back

neheb:


src/upnp_xml.cc
```diff
@@ -50,7 +50,7 @@ UpnpXMLBuilder::UpnpXMLBuilder(const std::shared_ptr<Context>& context,
     for (auto&& entry : this->config->getArrayOption(CFG_IMPORT_RESOURCES_ORDER)) {
         auto ch = MetadataHandler::remapContentHandler(entry);
         if (ch > -1) {
-            orderedHandler.push_back(ch);
+            orderedHandler.push_back(std::move(ch));
```
KarlStraussberger, line 53: That's an integer. Is move really changing anything?

neheb: Nope. Will fix.

neheb: Fixed.

PR #1794, opened 2021-08-15
use more auto in taglib

neheb:
Helps with readability. Also provides a size reduction because of
auto&&.


whyman:
All the types are there because it helped me figure out wtf was going on :smile:

neheb:
Removed some to keep more types.

PR #1806, opened 2021-08-20
remove pointless close

neheb:
The destructor calls this already.

### DIFF
--- a/src/metadata/matroska_handler.cc
+++ b/src/metadata/matroska_handler.cc
@@ -159,8 +159,6 @@ void MatroskaHandler::parseMKV(const std::shared_ptr<CdsItem>& item, std::unique
 
         el_l0 = ebml_stream.FindNextElement(LIBMATROSKA_NAMESPACE::KaxSegment_Context, i_upper_level, ~0, true);
     } // while elementLevel0
-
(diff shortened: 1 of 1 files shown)


src/metadata/matroska_handler.cc
```diff
@@ -159,8 +159,6 @@ void MatroskaHandler::parseMKV(const std::shared_ptr<CdsItem>& item, std::unique
 
         el_l0 = ebml_stream.FindNextElement(LIBMATROSKA_NAMESPACE::KaxSegment_Context, i_upper_level, ~0, true);
     } // while elementLevel0
-
-    ebml_file.close();
 }
```
KarlStraussberger, line 162: That file is created in line 141 and must be closed here

KarlStraussberger (changes requested):
The destructor close does not address this object

src/metadata/matroska_handler.cc
neheb: Why would the destructor not be called?

KarlStraussberger: The destructor is called by the ebml_file object cannot be closed, because its scope is the method parseMKV

PR #1808, opened 2021-08-20
debian::unstable uses libduktape206

KarlStraussberger:
fixes build issue on master


whyman:
I think its codename is "sid"

KarlStraussberger:
Found that out too, don't have debian running anywhere close by

PR #1810, opened 2021-08-20
improve debian script

KarlStraussberger:
News Dome printout to fix unstable build


KarlStraussberger:
Seems to be an issue in the image. It still reports `bullseye` although it should be `sid`.

PR #1816, opened 2021-08-23
std::tuple conversion

neheb:


src/file_request_handler.cc
```diff
@@ -73,7 +75,7 @@ static bool checkFileAndSubtitle(fs::path& path, const std::shared_ptr<CdsObject
         }
         throw_std_runtime_error("Failed to open {}: {}", path.c_str(), std::strerror(errno));
     }
-    return is_srt;
+    return { is_srt, mimeType, statbuf };
```
KarlStraussberger, line 78: That copies the whole struct which results in a similar issue as https://github.com/gerbera/gerbera/security/code-scanning/398?query=ref%3Arefs%2Fheads%2Fmaster

neheb: dead link for me.

KarlStraussberger: grafik

neheb: Out of curiosity, does that check trigger elsewhere?

KarlStraussberger: > Out of curiosity, does that check trigger elsewhere?

CodeQL triggered at the new struct which is in the screenshot. Nowhere else, right now.

neheb: Strange that this PR didn't touch that code.

PR #1818, opened 2021-08-23
Fix linkage error in on aarch64 with g++-10

### LINKED ISSUE #1674: Gerbera fails to link with GCC 10+ when build with -Os
Just going to link this here: https://github.com/openwrt/packages/issues/16080

KarlStraussberger:
fixes #1674


whyman:
Its been a long time since ive seen the compiler mess up like that.

KarlStraussberger:
> 
> 
> Its been a long time since ive seen the compiler mess up like that.

It optimized a template out, so that assignment was not possible. Hope it's gone.

PR #1819, opened 2021-08-23
Split CI validation jobs

KarlStraussberger:
makes it easier to rerun js if stuck


whyman:
Nice! I already had similar in my branch

PR #1824, opened 2021-08-24
add missing std::vector

neheb:
small size reduction. fmt seems to like vector.

This also avoids using std::initializer_list, which should not ve used.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -1171,12 +1171,12 @@ int SQLDatabase::createContainer(int parentID, std::string name, const std::stri
     log_debug("Created object row, id: {}", newId);
 
     if (!itemMetadata.empty()) {
+        auto mfields = std::vector {
+            identifier("item_id"),
+            identifier("property_name"),
+            identifier("property_value"),
+        };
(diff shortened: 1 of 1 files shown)


src/database/sql_database.cc
```diff
@@ -2515,18 +2515,18 @@ void SQLDatabase::migrateMetadata(int objectId, const std::string& metadataStr)
...
+                identifier("property_name"),
+                identifier("property_value"),
+            };
             auto values = std::vector {
                 fmt::format("{}", objectId),
                 fmt::format("{}", key),
                 fmt::format("{}", val),
             };
-            insert(METADATA_TABLE, fields, values);
+            insert(METADATA_TABLE, std::move(fields), std::move(values));
```
KarlStraussberger, line 2529: Move does not make sense to me if the target is a const ref. Recreating the vector in each loop seems expensive, too.

KarlStraussberger, line 2529: Move does not make sense to me if the target is a const ref. Recreating the vector in each loop seems expensive, too.

neheb:
How about now?

KarlStraussberger:
Great!

PR #1834, opened 2021-08-26
Simplify or remove unnecessary fmt::format calls

matthiasstraka:
There are several places where there is an unnecessary integer `quote()` call or a simple `fmt::format("{}", value)` which can be simplified to `fmt::to_string(value)`, or removed entirely if value is a `std::string` already.
Sometimes, I added a {:d} format type specifier so the runtime should throw immediately in case someone messes with the #defines.


matthiasstraka:
I reviewed the all modifications for possible lost quotes on varchar columns and could not find any. All dropped `quote()` calls are on int columns. Also, I'm not planning to add more changes to this PR.

neheb (approved):

PR #1839, opened 2021-08-27
remove Debian 10 compatibility wrappers

neheb:
Debian 11 is already out, which fixes this issue.

### DIFF
--- a/src/cds_objects.cc
+++ b/src/cds_objects.cc
@@ -161,7 +161,7 @@ void CdsItem::validate() const
         return;
 
     std::error_code ec;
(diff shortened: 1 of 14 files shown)


neheb:
can't those folks run docker?

KarlStraussberger:
No, because they ran it on the system for ages. Most people don't upgrade within a week.

neheb:
Wonder if for Debian 10 it would be a good idea to recompile all C++ dependencies with LLVM libc++ and link them statically.

whyman:
Lets revisit this in a year or so...

PR #1854, opened 2021-08-28
CamelCase changes

neheb:


whyman:
Please ensure you also apply to all the ifdef variants

neheb:
Not sure what you mean. This is mostly the web stuff.

whyman:
Why did web not get updated?

For Cases where part is an acronym e.g. IO it should be in upper case. Eg FileIOHandler

neheb:
Web is in the second commit.

neheb:
Fixed FileIOHandler.

PR #1855, opened 2021-08-29
use updated FindTagLib.cmake

neheb:
Avoids using horrible taglib-config, which doesn't work right on Debian
10 without dpkg-architecture.

Taken from upstream:

https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/find-modules/FindTaglib.cmake


neheb:
Interesting...

At least fedora and debian have these modules as packages. Wonder if a dependency should be added instead.

whyman:
And it uses a target too. Great.

PR #1858, opened 2021-08-29
use updated FindLibExiv2

neheb:
Taken from upstream:

https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/find-modules/FindLibExiv2.cmake


neheb:
Another benefit is that the version is shown correctly instead of ""

whyman:
Thanks

PR #1859, opened 2021-08-29
remove optional check

neheb:
Gerbera requires GCC8 nowadays.


whyman:
I don't think a minimum compiler is set, other than language support cpp17?

neheb:
GCC7 supports c++17 but cannot compile Gerbera because of std::filesystem. That's handled by a different check.

PR #1860, opened 2021-08-29
use normal libcurl search

neheb:
CMake includes a FindCURL.cmake


whyman:
Thanks

PR #1862, opened 2021-08-29
Batch inserts to metadata table

matthiasstraka:
When creating a container, often multiple inserts to the metadata table are executed in a for loop. It is faster to batch these inserts into a single SQL statement.


src/database/sql_database.cc
```diff
@@ -1176,14 +1176,13 @@ int SQLDatabase::createContainer(int parentID, std::string name, const std::stri
...
+        valuesets.reserve(itemMetadata.size());
+        const std::string newIdStr = fmt::to_string(newId);
         for (auto&& [key, val] : itemMetadata) {
-            auto mvalues = std::vector {
-                fmt::to_string(newId),
-                quote(key),
-                quote(val),
-            };
-            insert(METADATA_TABLE, mfields, mvalues);
+            valuesets.push_back({ newIdStr, quote(key), quote(val) });
```
neheb, line 1183: Sounds like emplace_back should be used.

src/database/sql_database.cc
```diff
@@ -1193,10 +1192,27 @@ int SQLDatabase::createContainer(int parentID, std::string name, const std::stri
 
 int SQLDatabase::insert(const char* tableName, const std::vector<SQLIdentifier>& fields, const std::vector<std::string>& values, bool getLastInsertId)
 {
+    assert(fields.size() == values.size());
     auto sql = fmt::format("INSERT INTO {} ({}) VALUES ({})", identifier(tableName), fmt::join(fields, ","), fmt::join(values, ","));
     return exec(sql, getLastInsertId);
 }
 
+void SQLDatabase::insertMultipleRows(const char* tableName, const std::vector<SQLIdentifier>& fields, const std::vector<std::vector<std::string>>& valuesets)
```
neheb, line 1200: These const ref parameters look like they can be replaced by &&

whyman:
At least for sqlite, the important thing is when you commkt your transaction.

src/database/sql_database.cc
matthiasstraka: Only if `std::vector<std::string>` had a three-argument constructor. In fact, push_back is the only way this works: https://godbolt.org/z/nEdWoWoEY

neheb: Hmmm yeah. Only way to make it work is to use std::tuple or something. No point.

matthiasstraka: Ideally, we'd have a compiler enforced linking between the field names and values, too. For now, at least I added an assert that assures they have the same number of elements. Since there are only two methods that even use batched insert, the effort is not worth it.

src/database/sql_database.cc
matthiasstraka: I tried to keep it similar to insert(), and since there is no transfer of ownership intended, I see no point in doing so.

matthiasstraka:
> At least for sqlite, the important thing is when you commit your transaction.

What is your concern? This change should improve the non-transaction version since I expect to be the batched insert to be atomic in comparison to multiple inserts. Isn't it?
The only "issue" I found is that multiple inserts are only supported since [SQLite 3.7.11](http://www.sqlite.org/releaselog/3_7_11.html) and Gerbera states it requires [3.7.0](https://docs.gerbera.io/en/stable/compile.html) But this release is now almost 10 years old, so I guess we are safe. I will nevertheless update the CMakeLists.txt

src/database/sql_database.cc
neheb: The point is to avoid copies.

neheb: Just tested. Small size reduction by doing so.

matthiasstraka: What do you mean exactly? The fields/valuesets references?

neheb: Change const std::vector& to std::vector&& and add std::move. Done here: https://github.com/neheb/gerbera/commit/b7a721698a5545379f9c6160dab7695ca8c913b4

matthiasstraka: If you really insist on that change, I can do it. But I'm not a friend of this way. There should not be any copy when passing a reference and when adding std::vector&&, there is a high risk that in future code, someone forgets the std::move and then there will be a copy. Also, I should change the parameters of insert(), too - to be consistent.
What (code?) size reduction are we talking about and is it significant?

whyman:
This should give us a nice speed boost on import.

src/database/sql_database.cc
neheb: How? && forces std::move.

Size reduction is there but not significant.

matthiasstraka: > && forces std::move

True, this will force a move.

Feel free to change the interface and submit a PR. But similar to insert(), if I wanted to call the function twice with the same vector and for example reuse fields, I must assume that insertMultipleRows moves away the data (even though it will not move it anywhere). I'm OK with a small increase in size (what feature doesn't) and preserve code readabiliy and ease of understanding without the need for extra documentation.

PR #1868, opened 2021-08-30
WIP: add meson

neheb:

Gonna just post this here. I've tried to make it as similar as the CMakeLists.txt file.

Benefit is smaller size when built with minsize. It also works with coverity's build tool. The CMake one does not for some reason.

Some bugs are fixed compared to the cmake one: libatomic does not get linked unconditionally, the exiv2 auto_ptr hack is "fixed". The actual define needs to match the libc++, not the compiler.

FILE_OFFSET_BITS are also set to 64. Not configurable.

Various hacks still need to be fixed. Notably the git and compile info stuff.


whyman:
I would rather have fixes for the CMake.

neheb:
Added fixes from Debian 10. Turns out, even clang 7 can compile Gerbera.

KarlStraussberger:
Please, add a section to `doc/compile` to explain usage.

neheb:
Added doc section. Also updated to install files in /usr/share as well.

What's left with the meson build is:

testing support
proper support of compile info. It's currently hardcoded.

The big change is that it's actually usable now. GNOME Builder runs it just fine. Profiling also works.

meson.build
```diff
@@ -0,0 +1,467 @@
...
+endif
+
+fmt_dep = dependency('fmt', version : '>=6.1.2')
+pugixml_dep = dependency('pugixml')
+spdlog_dep = dependency('spdlog', version : '>=1.5.0')
+sqlite3_dep = dependency('sqlite3', version : '>=3.7.11')
+
+iconv_dep = cpp.find_library('iconv', required : false)
+if not iconv_dep.found()
+  if cpp.has_function('iconv')
```
eli-schwartz, line 32: In meson 0.60, you will be able to do `dependency('iconv')` as I've added a custom lookup in https://github.com/mesonbuild/meson/pull/9205 which will first check if iconv_open is available in libc without any dependencies, and then look for `cpp.find_library('iconv')`.

Unfortunately that does require a version so new it is not even released right now. But it will eventually let you simplify these 8 lines down to 1.

meson.build
```diff
@@ -0,0 +1,467 @@
...
+if cpp.links('''
+  #include <stdint.h>
+  int main(void) {
+    uint32_t x32 = 0;
+    uint64_t x64 = 0;
+    __atomic_load_n(&x32, __ATOMIC_SEQ_CST);
+    __atomic_load_n(&x64, __ATOMIC_SEQ_CST);
+    return 0;
+    }''', name : 'built-in atomics')
+  libatomic_dep = declare_dependency()
```
eli-schwartz, line 48: You can generally do this to get a fake dependency, which is slightly better because you can use dependency() in more places than declare_dependency()...

```
dependency('', required: false)
```

meson.build
```diff
@@ -0,0 +1,467 @@
...
+# Link to the socket library if it exists. This is something you need on Solaris/OmniOS/Joyent
+socket_dep = dependency('socket', required : false)
+
+if host_machine.system() == 'solaris'
+  libgerbera_args += '-DSOLARIS'
+elif host_machine.system() != 'linux'
+  libgerbera_args += '-DBSD'
+endif
+
+if get_option('buildtype') != 'debug'
```
eli-schwartz, line 83:
```
get_option('debug')
```

buildtype is a combo of optimization+debug.

meson.build
```diff
@@ -0,0 +1,467 @@
...
+    libgerbera_args += '-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_INFO'
+else
+    libgerbera_args += '-DTOMBDEBUG'
+    libgerbera_args += '-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_DEBUG'
+endif
+
+if get_option('avcodec')
+  avcodec_dep = dependency('libavcodec')
```
eli-schwartz, line 91: In general, for all these dependency lookups I suggest you stop using boolean options and begin using https://mesonbuild.com/Build-options.html#features, which you can pass as 

```
avcodec_dep = dependency('libavcodec', required: get_option('avcodec'))
```

If -Davcodec=auto, it will try to find the dependency and gracefully fail, if =disabled, it will log a message saying "Dependency libavcodec skipped: feature avcodec disabled", if =enabled then it will be required.

You can then reduce your if/elses, and just check whether the dependency(ies) got found, safe and secure in the knowledge that the feature option handles determining whether it should be an error or whether to return not-found-dependency for disabled features.

meson.build
```diff
@@ -0,0 +1,467 @@
...
+
+if get_option('lastfm')
+  lastfm_dep = dependency('lastfm')
+  libgerbera_args += '-DHAVE_LASTFMLIB'
+else
+  lastfm_dep = declare_dependency()
+endif
+
+if get_option('magic')
+  magic_dep = cpp.find_library('magic')
```
eli-schwartz, line 163: The standard file utility and its accompanying libmagic, provides libmagic.pc so dependency() should work here.

meson.build
```diff
@@ -0,0 +1,467 @@
...
+  socket_dep,
+  spdlog_dep,
+  sqlite3_dep,
+  taglib_dep,
+  upnp_dep,
+  uuid_dep,
+]
+
+# mangix hacks
+add_global_arguments('-DCOMPILE_INFO="mangix"', language : 'cpp')
```
eli-schwartz, line 229: You probably want add_project_arguments instead, which does exactly the same thing unless subprojects are involved, in which case add_global_arguments accidentally infects all subprojects even though no one actually wants that -- hence why add_project_arguments got added. :D

eli-schwartz (commented):
Hi, nice to see you considering meson. Some suggestions for potential optimizations:

meson.build
neheb: I'll keep this in mind.

meson.build
neheb: hence the hacks comment :).

They'll need to be implemented properly at some point to match CMake.

meson.build
neheb: AFAIK that's only for newer versions. Older versions don't have it.

meson.build
neheb: That is indeed cleaner. Noted.

meson.build
neheb: Could you elaborate on this?

My initial attempt at adding WITH_DEBUG failed as debug seems to be a reserved option.

meson.build
eli-schwartz: Oh hmm, darn it. Indeed: https://github.com/file/file/commit/2f8ae1e99e89cde6ef6b6d2b5b01480637809aba

My system does have it... :)

In that case you can do a fallback 

```
if get_option('magic') 
    magic_dep = dependency('libmagic', required: false)
    if not magic_dep.found()
        magic_dep = cpp.find_library('magic')
    endif
endif
```

But using feature options is a bit more boilerplate.

meson.build
eli-schwartz: https://mesonbuild.com/Builtin-options.html#core-options

"buildtype" is a core option, "optimization" and "debug" are slightly newer core options which bidirectionally map to buildtype. Instead of checking whether the *buildtype* option is "debug" (or debugoptimized! which you do not handle), you can check whether the "debug" option is boolean true.

See https://mesonbuild.com/Release-notes-for-0-48-0.html#toggles-for-build-type-optimization-and-vcrt-type

You cannot add a "debug" option to meson_options.txt if that is what you were trying.

meson.build
eli-schwartz:
```
magic_dep = dependency('', required: false)
if not get_option('magic').disabled() 
    magic_dep = dependency('libmagic', required: false)
    if not magic_dep.found()
        magic_dep = cpp.find_library('magic', required: get_option('magic'))
    endif
endif
```

Actually hmm, not so much boilerplate after all...

neheb: OTOH this project tracks very recent library versions. It's probably easier to make version 39 the minimum.

neheb:
I think I addressed most stuff.

Some stuff I kept as booleans. Specifically the stuff that needs two or more dependencies.

neheb:
Also note this project uses cxxopts as a single header library. Would it make sense to use https://github.com/mesonbuild/wrapdb/blob/master/subprojects/cxxopts.wrap ?

eli-schwartz:
It would, but on the other hand I don't know if you can actually remove the cmake vendored copy just yet...

Either way, meson would be able to take advantage of encapsulation, accountability, and clean upgradeability of the .wrap file interface.

neheb:
https://github.com/gerbera/gerbera/tree/master/scripts contains a bunch of scripts to install newer versions of packages. I assume these "wraps" build and statically link these libraries? If so, that would be a great alternative.

eli-schwartz:
Indeed they do, with `-Default_library=static` or per-wrap `-D<wrapname>:default_library=static` meson enforces that those wraps are linked statically.

And they don't even need root in order to do it :p they are just downloaded and checksummed and private copies built. (Those scripts do not do checksumming, BTW.)

neheb:
I'll test later on Debian 10. fmtlib and spdlog are the issues there.

neheb:
Hmm doesn't seem to work with meson 0.49.1 (version that comes with Debian 10)

eli-schwartz:
For versions of meson earlier than 0.55, automatic wrap fallback via the `[provide]` section was not yet implemented and that section of the .wrap file will be ignored.

If you need to support versions older than that, the dependency() function takes a kwarg `fallback: [ 'subproject_name', 'foo_dep' ]` which will fallback to configuring the subproject and looking for the variable foo_dep as the right-hand side of the mapping in the newer style [provide] section.

meson.build
```diff
@@ -0,0 +1,449 @@
...
+  libgerbera_args += '-DHAVE_NL_LANGINFO'
+endif
+
+if (cpp.has_function('setlocale'))
+  libgerbera_args += '-DHAVE_SETLOCALE'
+endif
+
+uuid_dep = dependency('uuid', required : false)
+if not uuid_dep.found()
+  if host_machine.system() == 'linux'
```
eli-schwartz, line 65: I'd use
```
uuid_dep = dependency('uuid', required : host_machine.system() == 'linux')
```

neheb: I think this is what you meant.

eli-schwartz: Yup.

neheb:
Added compile_info. Not totally happy with how it turned out...

meson.build
```diff
@@ -0,0 +1,463 @@
...
+  pugixml_dep,
+  socket_dep,
+  spdlog_dep,
+  sqlite3_dep,
+  taglib_dep,
+  upnp_dep,
+  uuid_dep,
+]
+
+compile_info = [ '-DCOMPILE_INFO="' +
```
eli-schwartz, line 206: This looks painful and complicated indeed. But this seems to be an array with one element, and later inserted as `'-DCOMPILE_INFO='.join(compile_info)`

I think maybe you meant to not include all those `+ ' ' +`. Just have:

```

  'WITH_MAGIC=' + magic_dep.found().to_string().to_upper(),
  'WITH_MYSQL=' + mysql_dep.found().to_string().to_upper(),
  'WITH_CURL=' + curl_dep.found().to_string().to_upper(),
```

etc.

maybe...

```meson
foreach info: [
    ['MAGIC', magic_dep],
    ['MYSQL', mysql_dep],
    ...
    ]
    found = info[1].found().to_string().to_upper()
    compile_info += 'WITH_@0@=@1@'.format(info[0], found)
endforeach
```

I'm very curious why the current cmake build includes COMPILE_INFO about whether or not the tests were enabled. :/ Likewise about whether the systemd unit was installed...

Given most of these are nearly identical to existing `libgerbera_args += '-DHAVE_*'` it feels like this should be handled in the .cpp source code using preprocessor macros to extend a string. But clearly that is not the current state of the cmake...

neheb: I tried the first version. I got output like -DCOMPILE_INFO="-DXXX=TRUE-DYY=FALSE-DZZZ=TRUE

neheb: Ah no

```
-DCOMPILE_INFO=\"WITH_MAGIC=TRUE-DCOMPILE_INFO=WITH_MYSQL=TRUE-DCOMPILE_INFO=WITH_CURL=TRUE-DCOMPILE_INFO=WITH_INOTIFY=TRUE-DCOMPILE_INFO=WITH_JS=TRUE-DCOMPILE_INFO=WITH_TAGLIB=TRUE-DCOMPILE_INFO=WITH_AVCODEC=TRUE-DCOMPILE_INFO=WITH_FFMPEGTHUMBNAILER=TRUE-DCOMPILE_INFO=WITH_EXIF=TRUE-DCOMPILE_I    NFO=WITH_EXIV2=TRUE-DCOMPILE_INFO=WITH_SYSTEMD=TRUE-DCOMPILE_INFO=WITH_LASTFM=TRUE-DCOMPILE_INFO=WITH_DEBUG=TRUE-DCOMPILE_INFO=WITH_TESTS=FALSE\"'
```

eli-schwartz: well, '-DCOMPILE_INFO='.join(compile_info) will get you:
- just compile_info, if there is only one element
- otherwise: compile_info[0] + '-DCOMPILE_INFO=' + compile_info[1] + '-DCOMPILE_INFO=' + compile_info[2] .........


But you have a second copy of -DCOMPILE_INFO= at the beginning of your compile_info string building. So you might as well do compile_info = '-DCOMPILE_INFO=' + ........ followed by add_project_arguments(compile_info, language: 'cpp')

Or, you could do:

```meson
add_project_arguments('-DCOMPILE_INFO=' + ' '.join(compile_info), language : 'cpp')
```

Because the .join() should apply to the single-space separator token.

neheb: I tried this: 

```
--- a/meson.build
+++ b/meson.build
@@ -203,24 +203,31 @@ libgerbera_deps = [
   uuid_dep,
 ]

-compile_info = [ '-DCOMPILE_INFO="' +
-  'WITH_MAGIC=' + magic_dep.found().to_string().to_upper() + ' ' +
-  'WITH_MYSQL=' + mysql_dep.found().to_string().to_upper() + ' ' +
-  'WITH_CURL=' + curl_dep.found().to_string().to_upper() + ' ' +
-  'WITH_INOTIFY=' + get_option('inotify').to_string().to_upper() + ' ' +
-  'WITH_JS=' + duktape_dep.found().to_string().to_upper() + ' ' +
-  'WITH_TAGLIB=' + taglib_dep.found().to_string().to_upper() + ' ' +
-  'WITH_AVCODEC=' + avcodec_dep.found().to_string().to_upper() + ' ' +
-  'WITH_FFMPEGTHUMBNAILER=' + ffmpegthumbnailer_dep.found().to_string().to_upper() + ' ' +
-  'WITH_EXIF=' + exif_dep.found().to_string().to_upper() +  ' ' +
-  'WITH_EXIV2=' + exiv2_dep.found().to_string().to_upper() + ' ' +
-  'WITH_SYSTEMD=' + get_option('systemd').to_string().to_upper() + ' ' +
-  'WITH_LASTFM=' + lastfm_dep.found().to_string().to_upper() + ' ' +
-  'WITH_DEBUG=' + get_option('debug').to_string().to_upper() + ' ' +
-  'WITH_TESTS=' + get_option('tests').to_string().to_upper() + '"'
-]
-
-add_project_arguments('-DCOMPILE_INFO='.join(compile_info), language : 'cpp')
+compile_info = '-DCOMPILE_INFO="'
+foreach info: [
+  ['WITH_MAGIC', magic_dep.found()],
+  ['WITH_MYSQL', mysql_dep.found()],
+  ['WITH_CURL', curl_dep.found()],
+  ['WITH_INOTIFY', get_option('inotify')],
+  ['WITH_JS', duktape_dep.found()],
+  ['WITH_TAGLIB', taglib_dep.found()],
+  ['WITH_AVCODEC', avcodec_dep.found()],
+  ['WITH_FFMPEGTHUMBNAILER', ffmpegthumbnailer_dep.found()],
+  ['WITH_EXIF', exif_dep.found()],
+  ['WITH_EXIV2', exiv2_dep.found()],
+  ['WITH_SYSTEMD', get_option('systemd')],
+  ['WITH_LASTFM', lastfm_dep.found()],
+  ['WITH_DEBUG', get_option('debug')],
+  ['WITH_TESTS', get_option('tests')],
+  ]
+
+  found = info[1].to_string().to_upper()
+  compile_info += 'WITH_@0@=@1@ '.format(info[0], found)
+endforeach
+
+compile_info += '"'
+
+add_project_arguments(compile_info, language : 'cpp')

 # mangix hacks
 add_project_arguments('-DGIT_COMMIT_HASH=""', language : 'cpp')
```

works. Except the last entry ends in a space.

eli-schwartz: compile_info should be declared as a [] and += adds new array elements.

> Or, you could do:
> ```meson
> add_project_arguments('-DCOMPILE_INFO=' + ' '.join(compile_info), language : 'cpp')
> ```

neheb:
Updated.

Added a subproject for lastfmlib to make it easier to build. The CMake stuff doesn't work right with it. Rather, upstream's pkgconfig file is broken and points to the wrong path (lib vs lib64).

Converted more stuff to use feature instead of boolean so that it gets enabled when present unless explicitly disabled.

Stuff left:

- Fix database test. Stock CMake test copies the databases to a new directory. Need to do the same here.
- git info missing
- compile info is...suboptimal. Stock CMake uses a hack to present each entry on its own line. Doesn't work here.

Future:
- Maybe add libupnp subproject to avoid having to use the install scripts. That would take a bit of work...
- taglib too. I lost some of that work during an OS reinstall...

neheb:
turns out cmake projects can be handled as subprojects. Added upnp fallback. This compiles on stock Debian 8 now.

neheb:
Updated with subprojects for libupnp, libebml, and libmatroska. They have been great in finding memory defects in gerbera with memory sanitizer.

The libupnp one is incomplete but works well enough for gerbera's purposes.

neheb:
@r4start @whorfin Any chance of testing this under SunOS and friends?

neheb:
Added npupnp subproject. Simpler and less problem free than the pupnp one. I'll probably end up removing the latter.

Adjusted inotify to support haiku and BSDs. No idea if it can be improved.

meson.build
```diff
@@ -0,0 +1,362 @@
...
+  libgerbera_args += '-DHAVE_FFMPEGTHUMBNAILER'
+endif
+
+inotify_dep = dependency('', required : false)
+if not get_option('inotify').disabled()
+  if host_machine.system() != 'linux' and host_machine.system() != 'haiku'
+    inotify_dep = dependency('inotify')
+  endif
+  libgerbera_args += '-DHAVE_INOTIFY'
+endif
```
eli-schwartz, line 129: When using feature options, you want the auto case to not be exactly equal to the enabled case. That means passing `required: get_option('inotify')` to the non-linux dependency lookup.

Since on Linux it won't be a dependency and is just "there", you will want to set a boolean inotify_found, e.g.

```

if host_machine.system() == 'linux' or host_machine.system() == 'haiku'
    inotify_dep = dependency('', required : false)
    inotify_found = not get_option('inotify').disabled()
else
    inotify_dep = dependency('inotify', required: get_option('inotify'))
    inotify_found = inotify_dep.found()
endif

if inotify_found
    libgerbera_args += '-DHAVE_INOTIFY'
endif
```

neheb:
updated to upstream meson wraps. libupnp is the major one left.

neheb:
Update: I was able to get this to compile and link under Debian 10 with clang + libc++ to avoid the std::filesystem problem. I was able to test briefly but a more thorough look is needed I think.

The two dependencies that do not work with such a setup are ffmpeg_thumbnailer and libexiv2. Those two need wraps to be made.

neheb:
```
error: Exception caught: filesystem error: in file_size: Value too large for defined data type [xxx.mp4]
```

so even libc++ 7.0 has this problem?  Debian 10 has LLVM 8 and 11 as part of backports. I'll try 8 next.

neheb:
libc++ 8 and 11 show the same issue on Debian 10. I think it's time to abandon this effort.

neheb:
Removing draft from this. I believe I fixed all remaining issues. gerbera --compile-info now works properly. ninja test also works correctly.

test/database/meson.build
```diff
@@ -0,0 +1,19 @@
+run_command('cp', meson.source_root() + '/src/database/sqlite3/sqlite3.sql', meson.current_build_dir(), check : true)
+run_command('cp', meson.source_root() + '/src/database/sqlite3/sqlite3-upgrade.xml', meson.current_build_dir(), check : true)
+run_command('cp', meson.source_root() + '/src/database/mysql/mysql.sql', meson.current_build_dir(), check : true)
+run_command('cp', meson.source_root() + '/src/database/mysql/mysql-upgrade.xml', meson.current_build_dir(), check : true)
```
eli-schwartz, line 4: Probably makes sense to use `configure_file(..., copy: true)` here, as that runs in-process, is cross-platform, and gets updated when the original files are modified.

eli-schwartz: Ideally it would work with something like the previous revision of this PR, by just setting the `workdir: ` of the test so it picks up the files in the source tree...

neheb: Fixed.

neheb:
Added missing tests (newly added since this PR).

Added minimum meson version. 0.56 is widely available.

Fixed up various meson warnings after addding that.

neheb:
@eli-schwartz what do you think about

```
git grep "\"@0@"
meson.build:add_project_arguments('-DGERBERA_VERSION="@0@"'.format(meson.project_version()), language : 'cpp')
meson.build:add_project_arguments('-DPACKAGE_DATADIR="@0@/@1@/@2@"'.format(get_option('prefix'), get_option('datadir'), meson.project_name()), language : 'cpp')
meson.build:add_project_arguments('-DCOMPILE_INFO="@0@"'.format('\\n'.join(compile_info)), language : 'cpp')
meson.build:  add_project_arguments('-DGIT_COMMIT_HASH="@0@"'.format(gchash.stdout().strip()), language : 'cpp')
meson.build:  add_project_arguments('-DGIT_BRANCH="@0@"'.format(gbranch.stdout().strip()), language : 'cpp')
test/config/meson.build:libgerbera_args += '-DCMAKE_BINARY_DIR="@0@"'.format(meson.project_source_root())
test/core/meson.build:libgerbera_args += '-DCMAKE_BINARY_DIR="@0@"'.format(meson.project_build_root())
test/scripting/meson.build:libgerbera_args += '-DSCRIPTS_DIR="@0@/scripts"'.format(meson.project_source_root())
```

?

eli-schwartz:
You must use either `/` or `join_paths()` for at least -DPACKAGE_DATADIR since it needs to correctly handle e.g. NixOS where datadir is not a subdirectory of the prefix. Which of the two you use is probably not super important.

It's also probably somewhat clearer to read, if you create the final path outside of the string, and then format it in.

e.g.

```
'-DSCRIPTS_DIR="@0@/scripts"'.format(...)
```
vs
```
'-DSCRIPTS_DIR="@0@"'.format(...)
```


it's easier to spot the semantic structure of the define.

neheb:
Hmm? How would datadir not being a subdirectory be handled?

neheb:
Switched to join_paths.

eli-schwartz:
> Hmm? How would datadir not being a subdirectory be handled?

NixOS paths distinguish between certain types of artifacts. Think of it as split packages, e.g. the debian distinction between -doc, -dev, and -bin packages. So for example, you can end up with the following `meson setup` args:
```
--prefix=/nix/store/hqvf17nn5fmwrlrgq9qyplz1id923jri-glib-2.72.1
--libdir=/nix/store/hqvf17nn5fmwrlrgq9qyplz1id923jri-glib-2.72.1/lib
--bindir=/nix/store/jj5pjm2dgwi8w9rgnaz75mnbkds0pazx-glib-2.72.1-bin/bin
--includedir=/nix/store/6zvnlni8ayc1hvdy51fppvssngqhc5g8-glib-2.72.1-dev/include
```

Meson will then return get_option() as follows:
```
prefix -> /nix/store/hqvf17nn5fmwrlrgq9qyplz1id923jri-glib-2.72.1
libdir -> lib # relative to prefix
bindir -> /nix/store/jj5pjm2dgwi8w9rgnaz75mnbkds0pazx-glib-2.72.1-bin/bin # cannot be made relative and therefore is absolute
includedir -> /nix/store/6zvnlni8ayc1hvdy51fppvssngqhc5g8-glib-2.72.1-dev/include # cannot be made relative and therefore is absolute
```

Nix does this IIRC because not all parts of any given package are relevant for all possible contexts e.g. runtime vs.dev (and their context isolation is more thorough than Debian's simple package splitting which requires installing or removing packages to manage the system context).

Anyway, the point is that get_option('datadir') may return a relative or an absolute path. IF it is a relative path, doing string formatting with `@0@/@1@` will work by happenstance. If it is absolute, you need join_paths() to sensibly handle this, which is like string formatting except that it will prune leading components that precede another absolute path.

neheb:
updated version and ran all the meson.build files through muon fmt.

neheb:
Just found out exif.wrap is broken. Ugh,

https://github.com/mesonbuild/wrapdb/pull/423

neheb:
Let's try fixed exif...

neheb:
Agan...

https://github.com/mesonbuild/wrapdb/pull/424

PR #1869, opened 2021-08-30
update FindFilesystem.cmake

neheb:
From https://github.com/vector-of-bool/CMakeCM

Fixes cross compilation problems.


neheb:
I'll note it's by the same author.

PR #1871, opened 2021-08-30
Revert "use targets for libmatroska and libebml"

whyman:
Reverts gerbera/gerbera#1861

Breaks build with older distos (e.g. Ubuntu 16.04 & 18.04) as no CMake config is provided in their versions of the package.


whyman:
We should still convert these to targets, but our find modules should also create them.

Issue #1863, opened 2021-08-29
Search returns duplicate results for albums due to virtual containers


mill1000:
Gerbera v1.9.1

Searching for an album returns duplicate results as shown in the screenshot. 

![Screenshot_20210829-160526](https://user-images.githubusercontent.com/6924622/131267161-298aec02-d71a-4fa7-8d46-083a14dc838d.png)

The duplicate results are from each virtual container that has `upnp_class` set to `object.container.album.musicAlbum`. 
![image](https://user-images.githubusercontent.com/6924622/131267213-a0a985fa-25f2-4a90-a589-4c04d9224d62.png)

This database was imported using the default JS import script. My custom import script has even more duplicates due to additional virtual containers. 

Removing upnp_class from the additional containers eliminates them from the search results, but browsing behavior is negatively impacted as the control point no longer knows to display the container as a music album.

Music track's don't have this issue. Perhaps the same approach could be used here. For each album, a "root" item is added to the database with `upnp_class = object.container.album.musicAlbum`. Each virtual container references this "root" album via `ref_id`.

KarlStraussberger:
The result is correct from the database point of view. The search matches for each virtual entry and the query does not know which is the _correct_ one (PC Directory may be hidden). In my opinion we should prefix the search result with the full virtual path or add a flag to the import which item is searchable.

mill1000:
A flag to the import would be akin to the idea of a "root" item I suppose.

I don't think there is any situation where I would want all 3 results, full paths or not.

KarlStraussberger:
Actually, everything needed was already there but not stored to database for containers. Now you can enable it in the config.

PR #1872, opened 2021-08-31
Use metadata title in search results if present

### LINKED ISSUE #1867: Title displayed in search results does not match title displayed during browse
Last one for a bit I promise.

The title displayed for an item in search does not match the title that is displayed when browsing. The title for search appears to be the raw dc_title field from that database. I think we should populate the title from metadata instead.

Population from metadata is specifically blocked here:
https://github.com/gerbera/gerbera/blob/9aaddea1306f16de227ffa441dac96ac15dde189/src/upnp_xml.cc#L161-L163

which was likely to prevent modification of the title which may have
- Been truncated due to size
- Modified to include a "Played" indicator

Instead, we could update the title when building the CdsObject like so:
```patch
diff --git a/src/database/sql_database.cc b/src/database/sql_database.cc
index 7d8913f5..cce34b94 100644
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -1389,7 +1389,10 @@ std::shared_ptr<CdsObject> SQLDatabase::createObjectFromSearchRow(const std::uni
 
     auto meta = retrieveMetadataForObject(obj->getID());
     if (!meta.empty())
+    {
         obj->setMetadata(meta);
+        obj->setTitle(meta[MetadataHandler::getMetaFieldName(M_TITLE)]);
+    }
 
     auto resources = retrieveResourcesForObject(obj->getID());
     if (!resources.empty()) {
```

If the devs feel like this is something worth doing I can packaged it up in a proper PR

mill1000:
Use metadata title in search results (if present) instead of the raw `dc_title` field. Closes #1867

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -1388,8 +1388,13 @@ std::shared_ptr<CdsObject> SQLDatabase::createObjectFromSearchRow(const std::uni
     obj->setClass(getCol(row, SearchCol::upnp_class));
 
     auto meta = retrieveMetadataForObject(obj->getID());
-    if (!meta.empty())
+    if (!meta.empty()) {
         obj->setMetadata(meta);
+
+        // Update tile from metadata if present
+        if (meta.count(MetadataHandler::getMetaFieldName(M_TITLE)))
+            obj->setTitle(meta.at(MetadataHandler::getMetaFieldName(M_TITLE)));
+    }
 
     auto resources = retrieveResourcesForObject(obj->getID());
     if (!resources.empty()) {
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -1392,7 +1392,7 @@ std::shared_ptr<CdsObject> SQLDatabase::createObjectFromSearchRow(const std::uni
         obj->setMetadata(meta);
 
         // Update tile from metadata if present
-        if (meta.count(MetadataHandler::getMetaFieldName(M_TITLE)))
+        if (meta.find(MetadataHandler::getMetaFieldName(M_TITLE)) != meta.end())
             obj->setTitle(meta.at(MetadataHandler::getMetaFieldName(M_TITLE)));
     }
 


src/database/sql_database.cc
```diff
@@ -1388,9 +1388,14 @@ std::shared_ptr<CdsObject> SQLDatabase::createObjectFromSearchRow(const std::uni
     obj->setClass(getCol(row, SearchCol::upnp_class));
 
     auto meta = retrieveMetadataForObject(obj->getID());
-    if (!meta.empty())
+    if (!meta.empty()) {
         obj->setMetadata(meta);
 
+        // Update tile from metadata if present
+        if (meta.count(MetadataHandler::getMetaFieldName(M_TITLE)))
```
neheb, line 1395: find is a bool, no?

neheb: to clarify, find() is faster than count() as it exits early.

mill1000: find() returns an iterator, so the proper form would be `meta.find() != meta.end()` IIRC. In the case of a std::map since all keys are unique I believe the performance would be equivalent.

mill1000: Sure no problem. I'll switch to find() later today

neheb (approved):

KarlStraussberger:
`title` is handled exactly the same way for search and browse (https://github.com/gerbera/gerbera/blob/24e5be820a1daaded4fcac1403787a452971cbc4/src/upnp_cds.cc#L189). The text depends on the item in the virtual layout. If you want to change the output, this is the better place.

KarlStraussberger:
I double checked the code in sql_database and it returns the same column. I may happen that your endpoint uses metadata in case of browsing while the text displayed after search is the original set by the layout script on import. Please, try another client, like foobar, on your mobile.

mill1000:
@KarlStraussberger I've tried BubbleUpnp on mobile, and foobar2000 on desktop. Both have the same behavior of showing the metadata title in browse, and the dc_title in search.

**foobar2000**
foobar search
foobar browse

**BubbleUpnp**
Screenshot_20210903-072623

mill1000:
I also saw no ability to perform a search against Gerbera on foobar2000 mobile.

KarlStraussberger:
I still don't see where it happens, did you configure title-properties? (https://docs.gerbera.io/en/latest/config-server.html#upnp)

mill1000:
I did not configure any additional properties. (Plus the documentation says title-properties can't be changed?)

It looks like the browse title is coming directly from the database. When I re-run the browse query on the DB directly I get the desired dc:title. This must be because browse is looking at the virtual items, while search (for tracks) is looking directly at the non-virtual items marked with `object.item.audioItem.musicTrack`.

So perhaps this PR was just a poorly thought out band-aid for a bigger issue. Along with the duplicate album issue in #1863. Perhaps this database structure just isn't ideal for search.

KarlStraussberger:
The real reason for the behaviour is that the title for the source items is determined from the file name.

I took your idea and implemented a configuration option for the title.

PR #1878, opened 2021-09-01
some code cleanups

neheb:
Uses remove() instead of erase(). Not too sure this works as intended.


neheb:
OK soooo, it turns out that changing std::list to std::deque or std::vector is enough to fix. Interestingly enough, this is the only usage of std::list in the codebase. I'm not particularly sure why.

whyman:
Can you write some tests for this? I dont have confidence to merge it as it stands.

neheb:
Since I ended up littering code cleanups here, I'll put the list removal in a different PR.

neheb:
Going to split these.

PR #1880, opened 2021-09-01
don't assign nullptr to std::string

neheb:
Basic godbolt test shows that assigning to nullptr calls
std::__throw_logic_error


whyman:
Why not do the nullptr check and return?

neheb:
Done.

PR #1884, opened 2021-09-01
remove last std::list

neheb:
Removes compiler warnings on ARM regarding changed ABI.

### DIFF
--- a/src/util/timer.cc
+++ b/src/util/timer.cc
@@ -143,7 +143,7 @@ void Timer::notify()
     auto lock = threadRunner->uniqueLock();
     assert(lock.owns_lock());
 
-    std::list<TimerSubscriberElement> toNotify;
+    std::deque<TimerSubscriberElement> toNotify;
 
     if (!subscribers.empty()) {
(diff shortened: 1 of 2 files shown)


whyman:
Do we care about those warnings? They are just warnings after all.

There is a reason to use list here. Can you tell me what it is?

neheb:
Warnings are annoying :).

I don't know why list is used here.

Reading https://baptiste-wicht.com/posts/2012/12/cpp-benchmark-vector-list-deque.html indicates vector and deaue are better in most cases. Although these benchmarks use quite large values to show differences...

whyman:
Did you test this?

neheb:
I've done a basic test, yes. More accurately, my raw branch that includes this and other commits.

Adding and listening to music works just fine.

whyman:
Does timer autoscan work? Multiple timer autoscans with different directories?

neheb:
Tested, Works fine.

whyman:
Just this patch, on its own

neheb:
Same result.

std::list to std::deque just sacrifices removal (erase()) performance while everything else gets faster.

PR #1886, opened 2021-09-01
Refactor database schema to remove unused indices

matthiasstraka:
I found some sub-optimal database schema settings I want to fix in this PR.

- The index `grb_config_value_item` is redundant as it references only the primary key of `grb_config_value`. It can be removed without side-effects
```SQL
CREATE TABLE "grb_config_value" (
            "item" varchar(255) primary key,
            "key" varchar(255) NOT NULL,
            "item_value" varchar(255) NOT NULL,
            "status" varchar(20) NOT NULL)
CREATE INDEX grb_config_value_item ON grb_config_value(item)
```

### DIFF
--- a/src/database/mysql/mysql-upgrade.xml
+++ b/src/database/mysql/mysql-upgrade.xml
@@ -76,4 +76,7 @@
         <script>CREATE INDEX `grb_cds_resource_id` ON `grb_cds_resource`(`item_id`,`res_id`)</script>
         <script migration="resources">ALTER TABLE `mt_cds_object` DROP COLUMN `resources`</script>
     </version>
+    <version number="14" remark="drop redundant index">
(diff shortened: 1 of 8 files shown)


matthiasstraka:
I'm not changing any functionality in the code here. The change simply removes an index on a primary key which only uses up storage but will not be used by the database (accessing the primary key directly is faster).

KarlStraussberger:
Sorry, misunderstood your comment. Removing just the index is fine.

matthiasstraka:
I did an upgrade test 13->14 with both mysql and sqlite, fixed an issue with MySql and I think this is ready to merge. I won't do more index updates in this PR.
There are a few too many indices on mt_cds_object and I suspect some are never used, but this requires more analysis at a later time.

Issue #1894, opened 2021-09-03
some memory leak


neheb:
This is a result of the whole SQL rework. ping @matthiasstraka 

```
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:699:1: warning: Potential leak of memory pointed to by field '_M_pi' [clang-analyzer-cplusplus.NewDeleteLeaks]
}
^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:543:9: note: Assuming 'playlistRef' is false
    if (playlistRef) {
        ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:543:5: note: Taking false branch
    if (playlistRef) {
    ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:551:16: note: Assuming the condition is false
    } else if (obj->isVirtual() && obj->isPureItem()) {
               ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:551:33: note: Left side of '&&' is false
    } else if (obj->isVirtual() && obj->isPureItem()) {
                                ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:556:16: note: Assuming the condition is false
    } else if (obj->getRefID() > 0) {
               ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:556:12: note: Taking false branch
    } else if (obj->getRefID() > 0) {
           ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:572:9: note: 'hasReference' is false
    if (hasReference || playlistRef)
        ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:572:9: note: Left side of '||' is false
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:572:25: note: 'playlistRef' is false
    if (hasReference || playlistRef)
                        ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:572:5: note: Taking false branch
    if (hasReference || playlistRef)
    ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:574:14: note: Assuming 'op' is not equal to Update
    else if (op == Operation::Update)
             ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:574:10: note: Taking false branch
    else if (op == Operation::Update)
         ^
/home/mangix/devstuff/gerbera/src/database/sql_database.cc:577:10: note: 'hasReference' is false
[...]

PR #1895, opened 2021-09-03
fix some weird memory leak

neheb:
clang-analyzer reports that memory is leaked here. I have no idea why.

Fixes: https://github.com/gerbera/gerbera/issues/1894

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -687,9 +687,8 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
 
     cdsObjectSql["parent_id"] = fmt::to_string(obj->getParentID());
 
-    auto returnVal = std::vector {
-        std::make_shared<AddUpdateTable>(CDS_OBJECT_TABLE, std::move(cdsObjectSql), op),
-    };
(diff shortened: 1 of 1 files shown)


neheb:
I just grepped the codebase. This pattern is used in initialization of complexOptions and nowhere else. All other places use push_back.

whyman:
Can you show us the "leak"?

neheb:
The linked issue has clang-analyzer output.

whyman:
@neheb Do you get the warning libc++ too?

neheb:
I do not. Guess it's a libstdc++ issue.

The fix is simple enough.

My theory on why this is leaking is because {} in C++ really loves to resolve to std::initializer_list, which is itself a temporary. So it gets created and subsequently gets destroyed after initialization, making the vector hold a pointer to a nonexistent value.

Or maybe it's something else... Who knows.

neheb:
Yeah... I think that's exactly what's happening.

https://godbolt.org/z/s6E79Pe3W

The shared_ptr is getting copied into the vector while the original is destroyed.

whyman:
I dont think there is a leak here. But the patch seems simple enough.

neheb:
With the original code, two shared_ptrs are created. Now it's just one.

PR #1899, opened 2021-09-04
Use lambda in UpnpXMLBuilder::renderObject to trim strings

mill1000:
Just a minor cleanup opportunity I noticed when digging into the search title issue.

- Use a lambda to perform string truncation when over `upnp-string-limit`. This unifies the behavior as one was checking for string::npos,and  the other was only looking for a negative value. According to the documentation "A negative value will disable this feature..." so I took this implementation as correct.
- Apply string truncation to all metadata values, instead of just title and description.
- Removed unnecessary brackets around album art

### DIFF
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -153,7 +153,7 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t 
 
         for (auto&& [key, val] : meta) {
             // Trim metadata value as needed
-            const char* str = trimString(val).c_str();
+            auto str = trimString(val).c_str();
 
             if (key == MetadataHandler::getMetaFieldName(M_DESCRIPTION))
                 result.append_child(key.c_str()).append_child(pugi::node_pcdata).set_value(str);
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -126,14 +126,20 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t 
     result.append_attribute("parentID") = obj->getParentID();
     result.append_attribute("restricted") = obj->isRestricted() ? "1" : "0";
 
-    std::string tmp = obj->getTitle();
-    if ((stringLimit != std::string::npos) && (tmp.length() > stringLimit)) {
-        tmp = tmp.substr(0, getValidUTF8CutPosition(tmp, stringLimit - 3));
-        tmp = tmp + "...";
-    }
+    auto trimString = [stringLimit](const std::string& s) {
+        // Do nothing if limit is negative, or string is already short enough
+        if (stringLimit < 0 || s.length() <= stringLimit)
+            return s;
 
-    result.append_child("dc:title").append_child(pugi::node_pcdata).set_value(tmp.c_str());
-    result.append_child("upnp:class").append_child(pugi::node_pcdata).set_value(obj->getClass().c_str());
+        ssize_t cutPosition = getValidUTF8CutPosition(s, stringLimit - strlen("..."));
+        return s.substr(0, cutPosition) + "...";
+    };
+
+    const std::string title = obj->getTitle();
+    const std::string upnp_class = obj->getClass();
+
+    result.append_child("dc:title").append_child(pugi::node_pcdata).set_value(trimString(title).c_str());
+    result.append_child("upnp:class").append_child(pugi::node_pcdata).set_value(upnp_class.c_str());
 
     auto auxData = obj->getAuxData();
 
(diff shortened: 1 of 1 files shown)


src/upnp_xml.cc
```diff
@@ -126,14 +126,20 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
...
     result.append_attribute("restricted") = obj->isRestricted() ? "1" : "0";
 
-    std::string tmp = obj->getTitle();
-    if ((stringLimit != std::string::npos) && (tmp.length() > stringLimit)) {
-        tmp = tmp.substr(0, getValidUTF8CutPosition(tmp, stringLimit - 3));
-        tmp = tmp + "...";
-    }
+    auto trimString = [stringLimit](const std::string& s) {
+        // Do nothing if limit is negative, or string is already short enough
+        if (stringLimit < 0 || s.length() <= stringLimit)
```
neheb, line 131: size_t cannot be negative.

src/upnp_xml.cc
```diff
@@ -144,24 +150,19 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
...
-                result.append_child(key.c_str()).append_child(pugi::node_pcdata).set_value(tmp.c_str());
-            } else if (key == MetadataHandler::getMetaFieldName(M_TRACKNUMBER)) {
-                if (upnp_class == UPNP_CLASS_MUSIC_TRACK) {
-                    result.append_child(key.c_str()).append_child(pugi::node_pcdata).set_value(val.c_str());
-                }
-            } else if (key != MetadataHandler::getMetaFieldName(M_TITLE)) {
-                addField(result, key, val);
-            }
+            // Trim metadata value as needed
+            const char* str = trimString(val).c_str();
```
neheb, line 156: auto can be used here. c_str is well understood to be const char*.

src/upnp_xml.cc
mill1000: I just matched what `getValidUTF8CutPosition` returns.
https://github.com/gerbera/gerbera/blob/04ce6d5ddebcb6a899b26d154ac792e6ee64be54/src/util/tools.cc#L985

src/upnp_xml.cc
mill1000: Done.

src/upnp_xml.cc
mill1000: Sorry I misunderstood what you were asking for. I see now you mean `stringLimit`. You are correct. I'm surprised the compiler isn't throwing a warning for this.

neheb: cppcheck does. But yes, I think this is an extremely common pattern such that compilers pretty much have given up.

mill1000: Well I just turned on `-Wall -Wextra -pedantic` to see...

About 370 something warnings. Lots and lots of unused parameters to prune.

neheb: clang's -Weverything is pretty insane :)

Those unused parameters are questionable. They mostly deal with virtual functions where unused parameters are expected.

mill1000: Yes that makes sense. I wonder if they could be silenced by adding default values. Oh well.

KarlStraussberger:
Thank you, looks really nice. I'm not sure about truncating every value, but I don't assume anything can get as long as description.

src/upnp_xml.cc
```diff
@@ -144,30 +150,24 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, size_t
...
-                result.append_child(key.c_str()).append_child(pugi::node_pcdata).set_value(tmp.c_str());
-            } else if (key == MetadataHandler::getMetaFieldName(M_TRACKNUMBER)) {
-                if (upnp_class == UPNP_CLASS_MUSIC_TRACK) {
-                    result.append_child(key.c_str()).append_child(pugi::node_pcdata).set_value(val.c_str());
-                }
-            } else if (key != MetadataHandler::getMetaFieldName(M_TITLE)) {
-                addField(result, key, val);
-            }
+            // Trim metadata value as needed
+            auto str = trimString(val).c_str();
```
KarlStraussberger, line 156: The result of `trimString` is temporary, therefore `str` points to _nowhere_. `str` needs to be a std:string to store the result and `c_str` can only be called when using it in `set_value`.

KarlStraussberger (changes requested):

src/upnp_xml.cc
mill1000: Good catch. I should probably review some older code of mine now because I'm fairly certain I've seen that pattern before.

mill1000:
I agree. Unless stringLimit was set very low (like 20 characters), I wouldn't expect many of the other items to be truncated.

PR #1900, opened 2021-09-04
replace pointers with references

neheb:
Simpler. It's never null either.


KarlStraussberger:
I never understood why pointers were used!

PR #1901, opened 2021-09-04
Add Composer Tag to Metacontent Handler

mayanez:
Added the `%composer%` pattern for resources. Should be useful for others looking to add composer artwork to their browsing experience.


KarlStraussberger:
Great, thank you

PR #1906, opened 2021-09-04
use string::npos

neheb:
It's more accurate here than numeric limits.

Also fix port type.

### DIFF
--- a/src/util/string_converter.cc
+++ b/src/util/string_converter.cc
@@ -92,7 +92,7 @@ std::string StringConverter::_convert(const std::string& str, bool validate,
 
     auto input = str.c_str();
     size_t length = str.length();
-    if (length < (std::numeric_limits<size_t>::max() / 4)) {
+    if (length < (std::string::npos / 4)) {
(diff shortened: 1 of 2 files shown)


src/util/string_converter.cc
```diff
@@ -92,7 +92,7 @@ std::string StringConverter::_convert(const std::string& str, bool validate,
 
     auto input = str.c_str();
     size_t length = str.length();
-    if (length < (std::numeric_limits<size_t>::max() / 4)) {
+    if (length < (std::string::npos / 4)) {
```
KarlStraussberger, line 95: From the definition shown in https://en.cppreference.com/w/cpp/string/basic_string/npos this may cause overflows between signed and unsigned

neheb: How so? npos is size_type which is the same as size_t. The -1 evaluates to max()

KarlStraussberger: Then I'm a bit too much afraid of integer overflow. :smile:

PR #1907, opened 2021-09-04
remove sort of redundant std::move

neheb:
If the return type matches the function, copy elision applies. In this
case, since the types don't match, the variables get copied. But because
unique_ptr has copy operations deleted, These types get moved.

This reintroduces the clang warning that this breaks under C++11 but
fixes GCC's -Wredundant-move one. The latter is probably more
applicable.

### DIFF
--- a/src/content/onlineservice/atrailers_content_handler.cc
+++ b/src/content/onlineservice/atrailers_content_handler.cc
@@ -215,7 +215,7 @@ std::shared_ptr<CdsObject> ATrailersContentHandler::getObject(const pugi::xml_no
     item->setFlag(OBJECT_FLAG_ONLINE_SERVICE);
     try {
         item->validate();
-        return std::move(item);
+        return item;
     } catch (const std::runtime_error& ex) {
         log_warning("Failed to validate newly created Trailer item: {}",
             ex.what());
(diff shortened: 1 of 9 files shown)


neheb:
ping @matthiasstraka

KarlStraussberger:
What about calling `unique_ptr` to avoid implicit calls which could be misinterpreted?

neheb:
Note sure what you mean. If you're talking about casting the make_unique call: see https://github.com/gerbera/gerbera/pull/1876#discussion_r700572729

KarlStraussberger:
I meant something like `return std::unique_ptr<IOHandler>(io_handler)` ?

_My best guess down here in Italy_

neheb:
There's no compile size difference. It's fundamentally the same anyway. Casting a derived class to a base one.

KarlStraussberger:
The benefit would be if one of these options did not cause warnings

neheb:
Do note that the clang warning is only for older compilers. It's also not default. -Wall -Wextra -pedantic does not throw it. -Weverything does.

GCC8 does not warn on these but GCC11 says these are redundant moves.

KarlStraussberger:
The older compilers will be gone in the nearer future. Thanks for the insights.

PR #1909, opened 2021-09-04
rename several shadowed variables

neheb:

### DIFF
--- a/src/metadata/exiv2_handler.cc
+++ b/src/metadata/exiv2_handler.cc
@@ -137,29 +137,27 @@ void Exiv2Handler::fillMetadata(const std::shared_ptr<CdsObject>& item)
         // if there are any auxilary tags that the user wants - add them
         const auto aux = config->getArrayOption(CFG_IMPORT_LIBOPTS_EXIV2_AUXDATA_TAGS_LIST);
         if (!aux.empty()) {
-            std::string value;
-            std::string auxtag;
+            std::string auxval;
 
-            for (auto&& j : aux) {
-                value.clear();
-                auxtag = j;
+            for (auto&& auxtag : aux) {
+                auxval.clear();
                 log_debug("auxtag: {} ", auxtag.c_str());
                 if (auxtag.substr(0, 4) == "Exif") {
-                    const auto md = exifData.findKey(Exiv2::ExifKey(auxtag));
+                    md = exifData.findKey(Exiv2::ExifKey(auxtag));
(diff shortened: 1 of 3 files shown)


src/metadata/exiv2_handler.cc
```diff
@@ -137,29 +137,29 @@ void Exiv2Handler::fillMetadata(const std::shared_ptr<CdsObject>& item)
...
                 log_debug("auxtag: {} ", auxtag.c_str());
                 if (auxtag.substr(0, 4) == "Exif") {
-                    const auto md = exifData.findKey(Exiv2::ExifKey(auxtag));
+                    md = exifData.findKey(Exiv2::ExifKey(auxtag));
                     if (md != exifData.end())
-                        value = md->toString();
+                        val = md->toString();
                 } else if (auxtag.substr(0, 3) == "Xmp") {
-                    const auto md = xmpData.findKey(Exiv2::XmpKey(auxtag));
+                    auto md = xmpData.findKey(Exiv2::XmpKey(auxtag));
```
KarlStraussberger, line 152: You don't need `auto` here either.

src/metadata/exiv2_handler.cc
```diff
@@ -137,29 +137,29 @@ void Exiv2Handler::fillMetadata(const std::shared_ptr<CdsObject>& item)
         // if there are any auxilary tags that the user wants - add them
         const auto aux = config->getArrayOption(CFG_IMPORT_LIBOPTS_EXIV2_AUXDATA_TAGS_LIST);
         if (!aux.empty()) {
-            std::string value;
+            std::string val;
```
KarlStraussberger, line 140: I'd prefer `auxval`

src/metadata/exiv2_handler.cc
```diff
@@ -137,29 +137,29 @@ void Exiv2Handler::fillMetadata(const std::shared_ptr<CdsObject>& item)
...
             std::string auxtag;
 
             for (auto&& j : aux) {
```
KarlStraussberger, line 143: `j` can become `auxtag` to drop the outer loop declaration.

src/metadata/exiv2_handler.cc
neheb: Yeah you do. xmpData is a difference type than ExifData.

KarlStraussberger: Then choose another name like `xmpMd`.

PR #1914, opened 2021-09-05
remove dead assignments

neheb:
There's an immediate return.


src/content/autoscan_inotify.cc
```diff
@@ -448,14 +448,12 @@ void AutoscanInotify::removeNonexistingMonitor(int wd, const std::shared_ptr<Wd>
                     //watches.remove(wd);
```
KarlStraussberger, line 448: Please change the comment to 
```
// removeWatch triggers an IN_IGNORED event so watches.erase(wd) is called in threadProc
```

neheb: Done.

KarlStraussberger: the second line can go!

PR #1919, opened 2021-09-07
Update build scripts to run on Raspbian

KarlStraussberger:
Kind of backing up some low priority changes.

My system at home is broken down with a filesystem error and Italy is running on EDGE mobile. If you think things are bad, there's always room for a bit more.


KarlStraussberger:
Also changed `install-pupnp` to work like the other scripts and made things work similarly (use directory `build` and run below current dir)

Includes update of upnp version which is not critical but available.

scripts/debian/build-deb.sh
```diff
@@ -30,7 +32,7 @@ function install-cmake() {
 function install-fmt {
   echo "::group::Installing fmt"
   #if [[ "$lsb_codename" == "bionic" || "$lsb_codename" == "buster" || "$lsb_codename" == "hirsute" ]]; then
-    sudo bash scripts/install-fmt.sh static
+    sudo bash ${ROOT_DIR}scripts/install-fmt.sh static
```
whyman, line 35: Missing slash here, see the broken deb CI jobs.

PR #1921, opened 2021-09-08
Optimize CdsObject creation/updates

matthiasstraka:
There is some potential to speed up how objects are created or updated

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -571,25 +571,25 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
     }
 
     std::map<std::string, std::string> cdsObjectSql;
-    cdsObjectSql["object_type"] = quote(obj->getObjectType());
+    cdsObjectSql.emplace("object_type", quote(obj->getObjectType()));
 
     if (hasReference || playlistRef)
-        cdsObjectSql["ref_id"] = quote(refObj->getID());
+        cdsObjectSql.emplace("ref_id", quote(refObj->getID()));
     else if (op == Operation::Update)
-        cdsObjectSql["ref_id"] = SQL_NULL;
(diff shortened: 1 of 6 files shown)


neheb (approved):

neheb:
Nice optimization.

operator[] is equivalent to try_emplace according to cppref. The try_part is not needed as this is emplacing into an empty container.

src/database/sql_database.cc
```diff
@@ -624,69 +622,64 @@ std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdat
...
-        cdsObjectSql["mime_type"] = quote(item->getMimeType().substr(0, 40));
+        cdsObjectSql.emplace("mime_type", quote(item->getMimeType().substr(0, 40)));
     }
 
+    if (obj->getParentID() == INVALID_OBJECT_ID) {
+        throw_std_runtime_error("Tried to create or update an object {} with an illegal parent id {}", obj->getLocation().c_str(), obj->getParentID());
+    }
+
+    cdsObjectSql.emplace("parent_id", quote(obj->getParentID()));
+
```
KarlStraussberger, line 668: Moving this up, may trigger the exception for an entry which could be skipped by the following check (`return {}`).

matthiasstraka: I had this in mind, too. But since we filter for the same `parent_id`, finding such an item would mean that an illegal object was inserted before and the database is invalid. If this case should be ignored (by evaluating the invalid id after duplicated checking), I can revert the code-move. Otherwise I prefer a simple check before the more expensive SQL query.

KarlStraussberger: First of all: there should be none of this. If duplicate entries are thrown in to be added, they may have broken `parent_id`.

Errata: If `parent_id` is _invalid_ there should not be a duplicate entry for this, because it could never have made its way into the database, because of the same check.

PR #1922, opened 2021-09-11
replace vector with forward_list

neheb:
Instead of doing erase(remove()), allows using only remove.


neheb:
Note that push_back has been replaced with push_front. The former does not exist and its replacement (insert_after) is inefficient with forward_list. As far as I can tell, the ordering is not important.

PR #1924, opened 2021-09-13
simplify it loop slightly

neheb:
Add a continue to avoid some code duplication.


src/content/autoscan_inotify.cc
```diff
@@ -648,13 +648,13 @@ void AutoscanInotify::removeWatchMoves(int wd)
                     auto watchMv = std::static_pointer_cast<WatchMove>(watch);
                     if (watchMv->getRemoveWd() == wd) {
                         log_debug("removing watch move");
-                        if (wdWatches->size() == 1) {
-                            inotify->removeWatch(checkWd);
-                            ++it;
-                        } else
+                        if (wdWatches->size() != 1) {
```
KarlStraussberger, line 651: For readability it should be `> 1`

PR #1925, opened 2021-09-13
use reference for std::string

neheb:
This only works by accident as the function is called once here.

### DIFF
--- a/src/database/database.h
+++ b/src/database/database.h
@@ -178,7 +178,7 @@ public:
     /// the object ID of the container that is last in the path. The
     /// updateID will hold the objectID of the container that was changed,
     /// in case new containers were created during the operation.
-    virtual void addContainerChain(std::string path, const std::string& lastClass, int lastRefID, int* containerID,
+    virtual void addContainerChain(std::string& path, const std::string& lastClass, int lastRefID, int* containerID,
         std::deque<int>& updateID, const std::map<std::string, std::string>& lastMetadata)
         = 0;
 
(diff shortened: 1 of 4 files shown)


KarlStraussberger:
If it does not work as `const &`, it's not the kind of code we should maintain!

neheb:
It does not:

```
../src/database/sql_database.cc: In member function ‘virtual void SQLDatabase::addContainerChain(const string&, const string&, int, int*, std::deque<int>&, const std::map<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >&)’:
../src/database/sql_database.cc:1240:18: error: binding reference of type ‘std::string&’ {aka ‘std::__cxx11::basic_string<char>&’} to ‘const string’ {aka ‘const std::__cxx11::basic_string<char>’} discards qualifiers
 1240 |     reduceString(virtualPath, VIRTUAL_CONTAINER_SEPARATOR);
      |                  ^~~~~~~~~~~
In file included from ../src/metadata/metadata_handler.h:41,
                 from ../src/cds_resource.h:40,
                 from ../src/cds_objects.h:42,
                 from ../src/database/sql_database.cc:41:
../src/util/tools.h:76:32: note:   initializing argument 1 of ‘void reduceString(std::string&, char)’
   76 | void reduceString(std::string& str, char ch);
      |                   ~~~~~~~~~~~~~^~~
ninja: build stopped: subcommand failed.
```

KarlStraussberger:
So `virtualPath` was copied by intention

PR #1926, opened 2021-09-13
fix wrong value being assigned

neheb:
The variable name was changed but this entry was not.


KarlStraussberger:
OOps

PR #1927, opened 2021-09-13
simplify init

neheb:
std::map's operator[] calls try_emplace by default. There's no need to
check if it exists.


KarlStraussberger:
Init is called more than once. If you remove the search you drop content that was already added.

PR #1934, opened 2021-09-13
Simplify AddUpdateTable vectors without shared_ptr

matthiasstraka:
There is no benefit in using shared_ptr for AddUpdateTable objects. This PR simplifies the use of this helper which reduces the code size and removes access-indirections. Dynamically growing the vector of objects is ensured to only use std::move operations.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -539,7 +539,7 @@ std::shared_ptr<CdsObject> SQLDatabase::checkRefID(const std::shared_ptr<CdsObje
     return findObjectByPath(location);
 }
 
-std::vector<std::shared_ptr<SQLDatabase::AddUpdateTable>> SQLDatabase::_addUpdateObject(const std::shared_ptr<CdsObject>& obj, Operation op, int* changedContainer)
+std::vector<SQLDatabase::AddUpdateTable> SQLDatabase::_addUpdateObject(const std::shared_ptr<CdsObject>& obj, Operation op, int* changedContainer)
 {
     std::shared_ptr<CdsObject> refObj;
     bool hasReference = false;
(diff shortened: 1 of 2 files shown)


src/database/sql_database.h
```diff
@@ -236,6 +236,8 @@ class SQLDatabase : public Database {
             , operation(operation)
         {
         }
+        AddUpdateTable(const AddUpdateTable&) = delete;
+        AddUpdateTable(AddUpdateTable&&) = default;
```
neheb, line 240: The C++ standard says if move operations are defined, copy operations get implicitly deleted.

I would also add an operator=(&&) here.

src/database/sql_database.cc
```diff
@@ -718,25 +718,25 @@ void SQLDatabase::addObject(const std::shared_ptr<CdsObject>& obj, int* changedC
...
+        data.emplace_back(CDS_OBJECT_TABLE, std::move(cdsObjectSql), Operation::Update);
     } else {
         if (IS_FORBIDDEN_CDS_ID(obj->getID()))
             throw_std_runtime_error("Tried to update an object with a forbidden ID ({})", obj->getID());
         data = _addUpdateObject(obj, Operation::Update, changedContainer);
     }
 
     beginTransaction("updateObject");
-    for (auto&& addUpdateTable : data) {
+    for (const auto& addUpdateTable : data) {
```
neheb, line 737: all for loops use auto&& in the codebase. I would not change that. auto&& here evaluates to const auto&.

src/database/sql_database.h
neheb: https://en.cppreference.com/w/cpp/language/copy_constructor

KarlStraussberger:
LGTM

PR #1935, opened 2021-09-13
capture several lambda refs by value

neheb:
value is the recommended way to do lambdas. Small size decrease.

### DIFF
--- a/src/config/dynamic_content.cc
+++ b/src/config/dynamic_content.cc
@@ -56,7 +56,7 @@ std::size_t DynamicContentList::getEditSize() const
     if (indexMap.empty()) {
         return 0;
     }
-    return (*std::max_element(indexMap.begin(), indexMap.end(), [&](auto a, auto b) { return (a.first < b.first); })).first + 1;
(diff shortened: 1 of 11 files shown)


src/config/dynamic_content.cc
```diff
@@ -83,7 +83,7 @@ std::shared_ptr<DynamicContent> DynamicContentList::get(std::size_t id, bool edi
 std::shared_ptr<DynamicContent> DynamicContentList::get(const fs::path& location)
 {
     AutoLock lock(mutex);
-    auto entry = std::find_if(list.begin(), list.end(), [&](auto&& c) { return c->getLocation() == location; });
+    auto entry = std::find_if(list.begin(), list.end(), [=](auto&& c) { return c->getLocation() == location; });
```
KarlStraussberger, line 86: Doesn't this copy `location` or is there some special treatment for const&?

neheb: There's a size reduction. I don't know why. All the lambdas that resulted in size increases I left alone.

KarlStraussberger: Size does not mean that it's better at runtime. I assume the the copy constructor for path is called here, which is linked from stdc++ instead of passing the reference.

PR #1944, opened 2021-09-15
Downgrade libspdlog dependency to 1.8.1

matthiasstraka:
There is only one line of code that prevents SpdLog 1.8.1 to be used. Downgrading to 1.8.1 makes it easier to build on Ubuntu [Hirsute](https://packages.ubuntu.com/hirsute/libspdlog-dev) with standard packages.
The `#if SPDLOG_VERSION >= 10802` is up for discussion, I just left it in there so we can later easily identify old code when Ubuntu updates spdlog to 1.8.2 or newer.


KarlStraussberger:
Thank you for working this out. We still build using our install-scripts, so I'm fine with it

matthiasstraka:
Thanks. Hopefully, we can one day have a self-sufficient build on Ubuntu and install a current gerbera using `apt install gerbera` to increase the number of users.

neheb:
Gerbera right now compiles with much lower versions.

In any case, the meson PR solves this differently.

KarlStraussberger:
We'll have to wait another year or two for that. During the last 18 months gerbera pushed forward a lot faster than any distro will be able to follow.

PR #1951, opened 2021-09-16
Replace stdio with fmt calls

matthiasstraka:
Simply replace std::cout with corresponding fmt::print calls to be consistent with other formatting code


KarlStraussberger:
Great

PR #1952, opened 2021-09-16
Unify simple SQL queries

matthiasstraka:
Simple SQL queries such as INSERT and DELETE often follow the same pattern. Provide simple methods that automatically handle quoting of identifiers and values and remove code redundancy.

Code for UPDATE methods will be added soon.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -1186,14 +1186,14 @@ int SQLDatabase::createContainer(int parentID, const std::string& name, const st
     return newId;
 }
 
-int SQLDatabase::insert(const char* tableName, const std::vector<SQLIdentifier>& fields, const std::vector<std::string>& values, bool getLastInsertId)
+int SQLDatabase::insert(const std::string_view& tableName, const std::vector<SQLIdentifier>& fields, const std::vector<std::string>& values, bool getLastInsertId)
 {
     assert(fields.size() == values.size());
     auto sql = fmt::format("INSERT INTO {} ({}) VALUES ({})", identifier(tableName), fmt::join(fields, ","), fmt::join(values, ","));
     return exec(sql, getLastInsertId);
 }
 
-void SQLDatabase::insertMultipleRows(const char* tableName, const std::vector<SQLIdentifier>& fields, const std::vector<std::vector<std::string>>& valuesets)
+void SQLDatabase::insertMultipleRows(const std::string_view& tableName, const std::vector<SQLIdentifier>& fields, const std::vector<std::vector<std::string>>& valuesets)
 {
     if (valuesets.size() == 1) {
         insert(tableName, fields, valuesets.front());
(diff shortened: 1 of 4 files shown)


KarlStraussberger:
Sad thing is, that each of your great functions is only called once!

matthiasstraka:
> Sad thing is, that each of your great functions is only called once!
There is still one in `SQLDatabase::_removeObjects` where we delete integer ids (but currently it is a string list, which needs to changed)
I didn't look through all UPDATEs yet, but there will be a higher degree of code-resue I suppose.

KarlStraussberger:
Looks like you should rebase your branch.

src/database/sql_database.cc
```diff
@@ -1741,11 +1756,7 @@ std::unique_ptr<Database::ChangedContainers> SQLDatabase::_recursiveRemove(
 
 std::unique_ptr<Database::ChangedContainers> SQLDatabase::_purgeEmptyContainers(const std::unique_ptr<ChangedContainers>& maybeEmpty)
 {
-#if SPDLOG_VERSION >= 10802
-    log_debug("start upnp: {}; ui: {}", fmt::join(maybeEmpty->upnp, ","), fmt::join(maybeEmpty->ui, ","));
-#else
-    log_debug("start upnp: {}; ui: {}", fmt::format("{}", fmt::join(maybeEmpty->upnp, ",")), fmt::format("{}", fmt::join(maybeEmpty->ui, ",")));
-#endif
+    log_debug("start upnp: {}; ui: {}", fmt::to_string(fmt::join(maybeEmpty->upnp, ",")), fmt::to_string(fmt::join(maybeEmpty->ui, ",")));
```
KarlStraussberger, line 1759: I'm quite sure, you don't intend to revert that change?

KarlStraussberger:
For the test to pass, the implementation of the template has to be in the header.

src/database/sql_database.cc
matthiasstraka: I came across the same pattern in the last commit. I think using fmt::to_string is readable enough to leave it like this. So I removed the #if

matthiasstraka:
> For the test to pass, the implementation of the template has to be in the header.

Now that you mention it, this makes sense. I wonder why it works on my machine. I'll look at it later when I'm at a computer again.

KarlStraussberger:
> 
> 
> > For the test to pass, the implementation of the template has to be in the header.
> 
> Now that you mention it, this makes sense. I wonder why it works on my machine. I'll look at it later when I'm at a computer again.

_Minimal_ does only include part of the stuff and maybe does not need the `delete<int>` version and therefor it's not in the lib...

matthiasstraka:
I had a similar code transformation planned for the UPDATE statements but I'm not sure when and how long I will find the time to write and stabilize those. Let's better wrap this PR up so we there is still time to test before 1.9.2 is released.

PR #1956, opened 2021-09-21
Update command generator method

matthiasstraka:
Provide an easy to use updateRow method that generates UPDATE statements with a single WHERE filter
More complex filter capabilities planned for a future PR.


KarlStraussberger:
Great, I'd prefer it without the `identifier` call, because it's shorter, but that would need another call.

PR #1957, opened 2021-09-21
Rewrite split string

matthiasstraka:
The splitString function looks overly complex. Rewrite using simple element-wise iteration over std::string_view. I added a UT to verify that the implementation behaves identical to the old code.

Discussion: The old behavior of the empty parameter is inconsistent since empty strings a the start and end of str are not reported in the result. The rewrite copies that behavior so we don't break anything. Should we leave it at that?


KarlStraussberger:
> Discussion: The old behavior of the empty parameter is inconsistent since empty strings a the start and end of str are not reported in the result. The rewrite copies that behavior so we don't break anything. Should we leave it at that?

As far as I can see there's only one usage of `splitString` with `empty=true` in `cds_resource.cc`. What about copying the strange code to cds_resource and simplifying `splitString`.

whyman:
And some tests, nice

matthiasstraka:
Since there is only one existing call of splitString where empty=true, I analyzed the code. Basically, it is used in this context:
```cpp
std::vector<std::string> parts = splitString(serial, RESOURCE_PART_SEP, true);
if (parts.size() < 2 || parts.size() > 4) // fail
dictDecode(parts[1..3], &output_map);
```
Since dictDecode is a no-op for empty input strings, I went along an broke the behavior of the old splitString to return an empty string wherever you might expect one (i.e. splitString("1,2,,") -> {"1", "2", "", ""})
I added an additional unit test for the currently only usecase (migrating the old `mt_cds_object.resource` column) and used an actual resource-string of and old DB-version 12. Basically, these strings are a concatenation produced by CdsResource::encode() which contain 4 parts. Since dictDecode does nothing with an empty part, we can safely pass is an empty part (the old splitString would sometimes only return 2 parts when the last two were empty).
Upgrading from DB 12 to the current version appears to work without errors.

neheb:
Is is possible to make dictDecode return by value instead of void?

matthiasstraka:
> Is is possible to make dictDecode return by value instead of void?

Too late for this PR, but it is on my radar. I just did for its sister method and I will definitely stumble across dictDecode again

PR #1959, opened 2021-09-23
Modernize parts of the request handling with C++17 and add UTs

matthiasstraka:
I'm going over probably old code that is central to request and URL handling. These methods can strongly benefit from `std::string_view` optimizations to avoid copying string all over the place when dissecting them. While I'm at it, I add unit tests wherever suitable.

### DIFF
--- a/src/request_handler.cc
+++ b/src/request_handler.cc
@@ -31,12 +31,11 @@
 
 #include "request_handler.h" // API
 
-#include <numeric>
-#include <vector>
-
 #include "content/content_manager.h"
 #include "database/database.h"
 #include "util/tools.h"
+
+#include <fmt/core.h>
 
 RequestHandler::RequestHandler(std::shared_ptr<ContentManager> content)
     : content(std::move(content))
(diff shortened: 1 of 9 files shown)


src/request_handler.cc
```diff
@@ -48,22 +48,24 @@ RequestHandler::RequestHandler(std::shared_ptr<ContentManager> content)
 {
 }
 
-void RequestHandler::splitUrl(const std::string& url, char separator, std::string& path, std::string& parameters)
+std::tuple<std::string_view, std::string_view> RequestHandler::splitUrl(std::string_view url, char separator)
```
neheb, line 51: pair makes more sense here. tuple is used nowhere in the code.

src/request_handler.cc
```diff
@@ -48,22 +48,24 @@ RequestHandler::RequestHandler(std::shared_ptr<ContentManager> content)
...
+    }
 
     if (i1 == std::string::npos) {
-        path = url;
-        parameters.clear();
+        return { url, std::string_view() };
     } else {
-        parameters = url.substr(i1 + 1);
-        path = url.substr(0, i1);
+        return { url.substr(0, i1), url.substr(i1 + 1) };
```
KarlStraussberger, line 68: Be careful with `string_view`. It copies the pointer. If `url` is destroyed too soon it points to nowhere!

src/request_handler.cc
matthiasstraka: I'm OK with either version.

src/request_handler.cc
matthiasstraka: In all current use cases, the url is a `const char*` input to a method and all `string_view` are exclusively used inside that method. So we are safe.

src/request_handler.cc
```diff
@@ -69,30 +68,28 @@ std::tuple<std::string_view, std::string_view> RequestHandler::splitUrl(std::str
...
 {
-    std::map<std::string, std::string> params;
-
-    const auto parameters = std::string(filename + std::strlen(baseLink));
-    dictDecodeSimple(parameters, params);
-    log_debug("filename: {} -> parameters: {}", filename, parameters.c_str());
-
-    return params;
+    const auto parameters = filename.substr(baseLink.size());
+    //log_debug("filename: {} -> parameters: {}", filename, parameters);
```
KarlStraussberger, line 83: Please remove commented out log messages

src/request_handler.cc
```diff
@@ -69,30 +68,28 @@ std::tuple<std::string_view, std::string_view> RequestHandler::splitUrl(std::str
...
+    return dictDecodeSimple(parameters);
 }
 
-std::shared_ptr<CdsObject> RequestHandler::getObjectById(std::map<std::string, std::string> params) const
+std::shared_ptr<CdsObject> RequestHandler::getObjectById(const std::map<std::string, std::string>& params) const
 {
     auto it = params.find("object_id");
     if (it == params.end()) {
         //log_error("object_id not found in url");
```
KarlStraussberger, line 91: same here

src/request_handler.cc
```diff
@@ -69,30 +68,28 @@ std::tuple<std::string_view, std::string_view> RequestHandler::splitUrl(std::str
...
 
-std::string RequestHandler::joinUrl(const std::vector<std::string>& components, bool addToEnd, const std::string& separator)
+std::string RequestHandler::joinUrl(const std::vector<std::string>& components, bool addToEnd, std::string_view separator)
 {
-    return !components.empty()
-        ? (std::accumulate(next(components.begin()), components.end(), separator + components[0], [=](auto&& a, auto&& b) { return a + separator + b; }) + (addToEnd ? separator : ""))
-        : "/";
+    if (components.empty()) {
+        return std::string(separator);
+    } else {
```
neheb, line 75: no else after return please.

whyman:
Nice, thanks!

PR #1961, opened 2021-09-24
static_cast conversions

neheb:
No need for reinterpret_cast


neheb:
Interesting discovery I made...

PR #1963, opened 2021-09-24
Make title of search result configurable + set searchable container flag

### LINKED ISSUE #1863: Search returns duplicate results for albums due to virtual containers

### LINKED ISSUE #1867: Title displayed in search results does not match title displayed during browse

KarlStraussberger:
supports all metadata values

fixes #1863 
fixes #1867

replaces #1872


src/upnp_cds.cc
```diff
@@ -185,15 +185,29 @@ void ContentDirectoryService::doSearch(const std::unique_ptr<ActionRequest>& req
...
-            else
-                title.append(config->getOption(CFG_SERVER_EXTOPTS_MARK_PLAYED_ITEMS_STRING));
-
+            if (!titleSegs.empty()) {
+                auto values = std::vector<std::string>();
+                for (auto&& segment : titleSegs) {
+                    auto mtField = MetadataHandler::remapMetaDataField(segment);
+                    auto value = (mtField != M_MAX) ? cdsObject->getMetaData(mtField) : cdsObject->getMetaData(segment);
+                    if (!value.empty())
+                        values.emplace_back(value);
```
neheb, line 200: Why emplace and not push?

KarlStraussberger: Friday? :smile:

PR #1964, opened 2021-09-24
Tools optimizations

matthiasstraka:
Added more unit tests to tool functions, improved some string handling (avoid copying, avoid stringstream for known output size)

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -1341,8 +1341,7 @@ std::shared_ptr<CdsObject> SQLDatabase::createObjectFromRow(const std::unique_pt
     }
 
     std::string auxdataStr = fallbackString(getCol(row, BrowseCol::auxdata), getCol(row, BrowseCol::ref_auxdata));
-    std::map<std::string, std::string> aux;
-    dictDecode(auxdataStr, aux);
+    std::map<std::string, std::string> aux = dictDecode(auxdataStr);
     obj->setAuxData(aux);
 
     auto resources = retrieveResourcesForObject(obj->getID());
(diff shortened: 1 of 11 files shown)


src/database/sql_database.cc
```diff
@@ -2539,8 +2538,7 @@ bool SQLDatabase::doMetadataMigration()
 
 void SQLDatabase::migrateMetadata(int objectId, const std::string& metadataStr)
 {
-    std::map<std::string, std::string> dict;
-    dictDecode(metadataStr, dict);
+    std::map<std::string, std::string> dict = dictDecode(metadataStr);
```
neheb, line 2541: Wonder if auto is useful here.

src/util/tools.cc
```diff
@@ -123,11 +123,10 @@ bool startswith(std::string_view str, std::string_view check)
 #endif
 }
 
-std::string toLower(std::string_view str)
+std::string toLower(std::string str)
```
neheb, line 126: This one's somewhat interesting. So you're modifying the function parameter and then copying it. Wouldn't it make more sense to change it to

```
void toLower(std::string& str)
```

matthiasstraka: I added a `toLowerInPlace`, but currently I could only identify a single line where it can be used. But It's in there for the future now.

src/database/sql_database.cc
matthiasstraka: Yes, we could. But since it was there before, I'd like to keep it because it increases readability of the code. This way, we look at it and know its a map of string->string.

src/util/tools.cc
neheb: no I mean the toLower makes no sense now. The parameter is std::string, not const ref. So there's a copy.

src/util/tools.cc
```diff
@@ -426,28 +431,29 @@ std::string dictEncodeSimple(const std::map<std::string, std::string>& dict)
     return dictEncode(dict, '/', '/');
 }
 
-void dictDecode(std::string_view url, std::map<std::string, std::string>& dict, bool unEscape)
+std::map<std::string, std::string> dictDecode(std::string_view url, bool unEscape)
 {
-    auto data = url.data();
-    auto dataEnd = data + url.length();
+    std::map<std::string, std::string> dict;
```
KarlStraussberger, line 436: The point that worries me is that the original version just added to the dict while now we create a **new** and reassign it in the calling class.

The is no problem for temporary objects, but changes logic for a second call with the same instance of a class like`WebRequestHandler`

matthiasstraka: I am also a little unsure about this point. I have so far not really tested it - this is among other things the reasing this PR is still a draft. I will definitely look into it more closely when I find the time - and before setting this PR ready to merge.

matthiasstraka: I rewrote all cases that would have changed due to the new `dictDecode` function. `CdsResource` now decodes directly in the constructor, so we can never overwrite existing values. And for `WebRequestHandler`, I added a `dictMerge` function that merges the existing `params` with new ones (even though I have not observed any non-empty `params` so far). `dictMerge` does `emplace`, which is what `dictDecode` did before.

matthiasstraka:
For now, I consider this PR complete. Let me know of any further review concerns. If there are none, I will squash my commits together before undrafting this PR.

neheb:
This is still listed as a draft.

PR #1967, opened 2021-09-25
fix lastfm compilation

neheb:
It seems there was a function rename.


KarlStraussberger:
Thanks, haven't thought of grepping the codebase.

neheb:
CI doesn't check lastfmlib :)

whyman:
Metadata is one word, so it should be `getMetadata()`

https://en.wikipedia.org/wiki/Metadata

KarlStraussberger:
Good to learn :smile:. Well, while adding multi-value support I had to distinguish between the old `.*Metadata` with `map` and the new `.*MetaData` with `vector<pair>` in order to avoid issues with unaddressed usages. That's why the spelling is "wrong" now.

PR #1968, opened 2021-09-25
use addMetaData

neheb:
It's the proper function.


src/upnp_xml.cc
```diff
@@ -188,7 +188,7 @@ void UpnpXMLBuilder::renderObject(const std::shared_ptr<CdsObject>& obj, std::si
         auto meta = obj->getMetaData();
         auto [url, artAdded] = renderItemImage(virtualURL, item);
         if (artAdded) {
-            meta.emplace_back(MetadataHandler::getMetaFieldName(M_ALBUMARTURI), url);
+            obj->addMetaData(M_ALBUMARTURI, url);
```
KarlStraussberger, line 191: `addPropertyList` only reads the local copy, that's why it's added to `meta`. No goal to store albumart in metadata after searching it in the resources.

KarlStraussberger (changes requested):

PR #1973, opened 2021-09-28
Remove .c_str() for fmt-string arguments

matthiasstraka:
There are many calls to fmt::format (or spdlog calls) that call .c_str() on std::string. This makes the code less readable and throws away the string-length information (requires fmt to call strlen). I did a manual search-replace to find such calls and remove the .c_str() call whereever suitable.

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -100,7 +100,7 @@ void ActionRequest::update()
         std::ostringstream buf;
         response->print(buf, "", 0);
         std::string xml = buf.str();
-        log_debug("ActionRequest::update(): {}", xml.c_str());
+        log_debug("ActionRequest::update(): {}", xml);
 
 #if defined(USING_NPUPNP)
         UpnpActionRequest_set_xmlResponse(upnp_request, xml);
@@ -127,6 +127,6 @@ void ActionRequest::update()
             UpnpActionRequest_set_ErrCode(upnp_request, UPNP_E_ACTION_FAILED);
         }
 
-        log_error("ActionRequest::update(): response is nullptr, code {} for {}", errCode, actionName.c_str());
+        log_error("ActionRequest::update(): response is nullptr, code {} for {}", errCode, actionName);
     }
 }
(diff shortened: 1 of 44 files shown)


neheb:
IIRC, having c_str results in a smaller compiled size. Maybe I'm wrong though.

A better solution would probably be to pass a view, but that doesn't sound like a god idea always.

matthiasstraka:
> IIRC, having c_str results in a smaller compiled size. Maybe I'm wrong though.

Probably, the size is smaller with a single pointer instead of ptr+size. But execution speed is generally faster. 

> A better solution would probably be to pass a view, but that doesn't sound like a god idea always.

I did a short godbolt comparison before I went through the code. If I understand the result correctly, fmt uses a view by default. Otherwise it would hardly be able to brag about its speed.

neheb:
I tested this. Size does indeed increase.

Maybe it could be useful to do some string_view conversions along the way. For example: https://github.com/gerbera/gerbera/blob/master/src/config/client_config.cc#L168 can be converted to string_view. That reduces size.

matthiasstraka:
Well, if increasing size is such a problem, we should stop adding new features ;)

neheb:
It's not that. My point is that there's a reason c_str() is used here. Whether or not that reason is valid is up for argument.

matthiasstraka:
I did some comparison in [Compiler Explorer](https://gcc.godbolt.org/z/xEqWbe64v) to show different ways to pass a string to fmt. The added code with string is a few moves (and those are possibly more than in gerbera since in my short example string needs to be loaded from parameters into registers first). I am more than willing to add those few bytes to the binary to make the code better readable.
If we really want to decrese size and increase executions speed, we should go through log_debug messages and remove the ones that are not really needed. There are plenty of them that will not contribute to locating errors.

whyman (approved):

KarlStraussberger:
Rebase is required.
To be merged right after 1.9.2.

PR #1977, opened 2021-09-29
Use better primary key for grb_cds_resource 

matthiasstraka:
In database table `grb_cds_resource`, there is a primary key `id` which is not used anywhere in the code. What is actually used is the tuple `(item_id, res_id)` which is perfectly suited as a primary key. `item_id` is a foreign key and `res_id` is always created such that there are only unique `(item_id, res_id)` pairs.
In addition, with the new primary key, we can drop the index on `(item_id, res_id)`.

This change should not change the behavior of the current code but reduce the database size (fewer columns, fewer indices) as well as increase the speed of creating a database (no additional index update).

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -120,8 +120,7 @@ enum class MetadataCol {
 
 /// \brief resource column ids
 enum class ResourceCol {
-    Id = 0,
-    ItemId,
+    ItemId = 0,
     ResId,
     HandlerType,
     Options,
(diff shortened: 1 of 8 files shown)


src/database/sql_database.cc
```diff
@@ -2358,7 +2355,8 @@ std::vector<std::shared_ptr<CdsResource>> SQLDatabase::retrieveResourcesForObjec
     while ((row = res->nextRow())) {
         auto resource = std::make_shared<CdsResource>(std::stoi(getCol(row, ResourceCol::handlerType)));
         resource->decode(getCol(row, ResourceCol::options), getCol(row, ResourceCol::parameters));
-        resource->setResId(resources.size());
+        int res_id = row->col_int(to_underlying(ResourceCol::res_id), 0);
+        resource->setResId(res_id);
```
KarlStraussberger, line 2359: The idea was to ensure that there is no gap in ids when loading - the database may have been changed.

matthiasstraka: In that case, I can revert that change

PR #1978, opened 2021-09-30
Batch-execute multiple getChildCount queries

matthiasstraka:
In `SQLDatabase::browse(BrowseParam& param)`, we repeatedly call `getChildCount` which emits a new SQL query when browsing even though only the `parent_id` changes from query to query. This PR converts the multiple queries to a single query with GROUP BY. This will make browsing faster.


whyman:
I really like this change, but would love a test to verify that the behaviour is unchanged

matthiasstraka:
> I really like this change, but would love a test to verify that the behaviour is unchanged

It won't be that easy due to DB access, but I might be able to write a test that tests the Database interface and simulates the results like a database would. There is some implicit 0-value generation in `child_counts[cont->getID()]` when a parent_id returns match. I'll get to this later.

src/database/sql_database.cc
```diff
@@ -914,17 +914,25 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::browse(BrowseParam& param)
     commit("browse");
 
     std::vector<std::shared_ptr<CdsObject>> result;
+    std::vector<int> contIds;
```
KarlStraussberger, line 917: If you use a map<int, CdsContainer> you can avoid the second check and run through it directly.

matthiasstraka: Before, I had the collection of contIds in a separate loop but merged them together. As an alternative, we could make a second CdsObject-vector of containers and use this to iterate in the second loop. I will definitely put some more work into this PR but currently my spare time is limited. There's still time for 1.9.3 :)

matthiasstraka: I added a new changeset which addesses this. We could move the extraction of ID values into `SQLDatabase::getChildCounts` but on the other hand, I prefer to keep `SQLDatabase::getChildCount` and `SQLDatabase::getChildCounts` very similar.

PR #1991, opened 2021-10-02
clang-tidy: renames to WebKit style

neheb:


neheb:
This thing uncovered an annoying issue. Because of the way macros work with compilers, clang-tidy is unable to follow all paths. I think these macros should be replaced with if constexpr statements.

src/util/tools.cc
```diff
@@ -242,9 +242,9 @@ fs::path findInPath(const fs::path& exec)
     if (!p)
         return {};
 
-    std::string PATH = p;
+    std::string path = p;
     std::error_code ec;
```
KarlStraussberger, line 246: You need to select another name here, e.g. `pathString`. path is the loop variable already.

src/metadata/matroska_handler.cc
```diff
@@ -52,17 +52,17 @@
 // file managment
 class FileIOCallback : public IOCallback {
 private:
-    std::FILE* file;
+    std::FILE* m_file;
```
KarlStraussberger, line 55: We don't use that style for attributes. Maybe call it `mediaFile`

neheb:
Fixed.

neheb:
Rebased again.

KarlStraussberger:
Thank you. Really great job.

PR #1992, opened 2021-10-03
Block XML Declaration for IRadio based devices

### LINKED ISSUE #1987: Client chokes on <?xml..?> header in browse response and shows empty tree
I have one client (a Dual internet radio) that reports an empty content tree for my gerbera server.
After checking the differences between the gerbera browse response and the response of servers that work with the client, I found none of the other servers sends a `<?xml..?>` header in the result part of the browse response.

And indeed, after commenting out the two places in upn_cds.cc where pugi::node_declaration is prepended to the DIDL-Lite response the client works as expected.

Even if this is most probably a client bug:
As other UPNP servers do not add this declaration, I'd like to propose to either drop it completely or make it configurable.

KarlStraussberger:
fixes #1987

### DIFF
--- a/src/util/upnp_quirks.cc
+++ b/src/util/upnp_quirks.cc
@@ -175,3 +175,8 @@ void Quirks::saveSamsungBookMarkedPosition(const std::unique_ptr<ActionRequest>&
     auto response = UpnpXMLBuilder::createResponse(request->getActionName(), UPNP_DESC_CDS_SERVICE_TYPE);
     request->setResponse(std::move(response));
 }
+
+bool Quirks::blockXmlDeclaration() const
+{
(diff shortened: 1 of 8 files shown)


src/util/upnp_quirks.cc
```diff
@@ -175,3 +175,11 @@ void Quirks::saveSamsungBookMarkedPosition(const std::unique_ptr<ActionRequest>&
     auto response = UpnpXMLBuilder::createResponse(request->getActionName(), UPNP_DESC_CDS_SERVICE_TYPE);
     request->setResponse(std::move(response));
 }
+
+bool Quirks::blockXmlDeclaration() const
+{
+    if (!pClientInfo || (pClientInfo->flags & QUIRK_FLAG_IRADIO) == 0)
```
neheb, line 181: This looks like it can be simplified to a single return statement.

neheb (approved):

Issue #1993, opened 2021-10-03
Media are not playable on ARM after upgrade to 1.9.2


kyak:
I've upgraded from 1.8.2 to 1.9.2 on Arch Linux ARM. I've removed old database and diligently moved my old config to the new config (based on the automatically generated config).

I'm seeing the following error messages in log. The TV just shows "Check your network connection" when trying to play a file.

> error: Exception during seek: [/home/user/.cache/pikaur/build/gerbera/src/gerbera-1.9.2/src/iohandler/file_io_handler.cc:88] seek Error: fseek failed

Transcoding has also stopped working, but the error message is different from transcoded files:

> error: Exception: [/home/user/.cache/pikaur/build/gerbera/src/gerbera-1.9.2/src/file_request_handler.cc:244] open Error: Invalid resource ID given

What could be the issue?

kyak:
Frankly speaking, it broke in 1.9.1. But I was too busy to report/hoping the issue would fix itself in 1.9.2, but it didn't.

KarlStraussberger:
Thank you for reporting, any time is fine.

Could you please provide a stacktrace of the fiest crash, the file_io_handler code there was not changed for ages.

Did you rebuild your database after updating? Resource handling did get some overhaul which may also affect transcoding.

kyak:
> Could you please provide a stacktrace of the fiest crash, the file_io_handler code there was not changed for ages.

But gerbera doesn't crash at this point, there is only this error message in logs.

> Did you rebuild your database after updating? Resource handling did get some overhaul which may also affect transcoding.

I've completely removed the old database and it is recreated by gerbera.

ectotropic:
Very quickly adding my own experience to this - exactly the same issue with transcoding on 1.9.2, but I'm on x84/ubuntu.

That is, opening _any_ file for transcoding results in nothing being served and one or more
> error: Exception: [/__w/gerbera/gerbera/src/file_request_handler.cc:244] open Error: Invalid resource ID given

entries in the log file.

Definitely used to work relatively recently, but couldn't say exactly which version was the last I used that did work (something in the last few months though). FWIW a delete and complete rebuild of the database didn't help.

If I get time I'll run with `--debug` and post my logs.

whyman:
Thanks for confirming @ectotropic

stinovlas:
[...]

PR #1998, opened 2021-10-04
Path to map rework, handle empty values

whyman:
Added some tests to make sure things mostly work as I would expect.

Fixes: https://github.com/gerbera/gerbera/issues/1993


KarlStraussberger:
Thanks for covering up _my_ issue. Didn't intend that it shows up in url at all.

whyman:
Yes I suspected as much.

I think it's probably best to remove it as some clients have issues with double slashes in paths

PR #1999, opened 2021-10-05
convert some defines with constexpr

neheb:

### DIFF
--- a/src/content/scripting/playlist_parser_script.cc
+++ b/src/content/scripting/playlist_parser_script.cc
@@ -157,11 +157,7 @@ void PlaylistParserScript::processPlaylistObject(const std::shared_ptr<CdsObject
     currentLine = new char[ONE_TEXTLINE_BYTES];
     currentLine[0] = '\0';
 
-#ifdef __linux__
-    currentHandle = std::fopen(obj->getLocation().c_str(), "re");
-#else
-    currentHandle = std::fopen(obj->getLocation().c_str(), "r");
-#endif
+    currentHandle = std::fopen(obj->getLocation().c_str(), apple ? "r" : "re");
     if (!currentHandle) {
         currentObjectID = INVALID_OBJECT_ID;
         currentTask = nullptr;
--- a/src/iohandler/process_io_handler.cc
+++ b/src/iohandler/process_io_handler.cc
@@ -139,17 +139,9 @@ void ProcessIOHandler::open(enum UpnpOpenFileMode mode)
     }
 
     if (mode == UPNP_READ)
-#ifdef __linux__
         fd = ::open(filename.c_str(), O_RDONLY | O_NONBLOCK | O_CLOEXEC);
-#else
-        fd = ::open(filename.c_str(), O_RDONLY | O_NONBLOCK);
-#endif
     else if (mode == UPNP_WRITE)
-#ifdef __linux__
         fd = ::open(filename.c_str(), O_WRONLY | O_NONBLOCK | O_CLOEXEC);
-#else
-        fd = ::open(filename.c_str(), O_WRONLY | O_NONBLOCK);
-#endif
     else
         fd = -1;
 
--- a/src/util/mt_inotify.cc
+++ b/src/util/mt_inotify.cc
@@ -56,11 +56,7 @@
 #define MAX_STRLEN 4096
 
 Inotify::Inotify()
-#ifdef __linux__
     : inotify_fd(inotify_init1(IN_CLOEXEC))
-#else
-    : inotify_fd(inotify_init())
-#endif
 {
     if (inotify_fd < 0)
         throw_std_runtime_error("Unable to initialize inotify");
@@ -84,11 +80,7 @@ Inotify::~Inotify()
 
 bool Inotify::supported()
 {
-#ifdef __linux__
     int testFd = inotify_init1(IN_CLOEXEC);
-#else
-    int testFd = inotify_init();
-#endif
     if (testFd < 0)
         return false;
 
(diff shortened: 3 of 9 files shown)


src/util/mt_inotify.cc
```diff
@@ -55,12 +55,14 @@
...
 #ifdef __linux__
-    : inotify_fd(inotify_init1(IN_CLOEXEC))
+static constexpr bool linux = true;
 #else
-    : inotify_fd(inotify_init())
+static constexpr bool linux = false;
 #endif
+
+Inotify::Inotify()
+    : inotify_fd(linux ? inotify_init1(IN_CLOEXEC) : inotify_init())
```
KarlStraussberger, line 65: I'm not sure whether this works on all systems. The symbol may be undefined.

src/content/scripting/playlist_parser_script.cc
```diff
@@ -42,6 +42,12 @@
 
 #define ONE_TEXTLINE_BYTES 1024
 
+#ifdef __linux__
+static constexpr bool linux = true;
+#else
+static constexpr bool linux = false;
+#endif
+
```
KarlStraussberger, line 50: What about moving this to a header like `compat.h`. Avoids having the same code in many files

src/util/process.cc
```diff
@@ -48,21 +48,22 @@
...
     char tempIn[] = "mt_in_XXXXXX";
     char tempOut[] = "mt_out_XXXXXX";
 
     std::string inputFile = tempName(cfg->getOption(CFG_SERVER_TMPDIR), tempIn);
-#ifdef __linux__
-    fd = open(inputFile.c_str(), O_RDWR | O_CREAT | O_EXCL | O_CLOEXEC, S_IRUSR | S_IWUSR);
-#else
-    fd = open(inputFile.c_str(), O_RDWR | O_CREAT | O_EXCL, S_IRUSR | S_IWUSR);
-#endif
+    fd = open(inputFile.c_str(), linux ? (O_RDWR | O_CREAT | O_EXCL | O_CLOEXEC) : (O_RDWR | O_CREAT | O_EXCL), S_IRUSR | S_IWUSR);
```
KarlStraussberger, line 66: You can merge declaration and definition now.

KarlStraussberger (commented):
I really like the reduced code.

src/util/mt_inotify.cc
neheb: IN_CLOEXEC?

src/content/scripting/playlist_parser_script.cc
```diff
@@ -157,11 +163,7 @@ void PlaylistParserScript::processPlaylistObject(const std::shared_ptr<CdsObject
     currentLine = new char[ONE_TEXTLINE_BYTES];
     currentLine[0] = '\0';
 
-#ifdef __linux__
-    currentHandle = std::fopen(obj->getLocation().c_str(), "re");
-#else
-    currentHandle = std::fopen(obj->getLocation().c_str(), "r");
-#endif
+    currentHandle = std::fopen(obj->getLocation().c_str(), linux ? "re" : "r");
```
matthiasstraka, line 166: Wouldn't it make more sense to define/constexpr some `openMode = "re"` or `openMode = "r"` and use that in fopen instead of `linux ?:`? This way we can reuse if for other fopens

neheb: not really. there aren't too many similar ones.

src/content/scripting/playlist_parser_script.cc
neheb: Fixed.

src/util/mt_inotify.cc
KarlStraussberger: `inotify_init` or `inotify_init1`

src/iohandler/process_io_handler.cc
```diff
@@ -139,17 +139,9 @@ void ProcessIOHandler::open(enum UpnpOpenFileMode mode)
     }
 
     if (mode == UPNP_READ)
-#ifdef __linux__
-        fd = ::open(filename.c_str(), O_RDONLY | O_NONBLOCK | O_CLOEXEC);
-#else
-        fd = ::open(filename.c_str(), O_RDONLY | O_NONBLOCK);
-#endif
+        fd = ::open(filename.c_str(), linux ? (O_RDONLY | O_NONBLOCK | O_CLOEXEC) : (O_RDONLY | O_NONBLOCK));
```
KarlStraussberger, line 142: Is `O_CLOEXEC` define on non-linux system?

neheb: I don't think so.

src/util/mt_inotify.cc
neheb: CI seems to have no issue.

src/iohandler/process_io_handler.cc
KarlStraussberger: Yes, because it's only running on Linux-like systems. Some users still compile for Sun etc.

neheb: shouldn't be an issue. all of this should be done at compile time.

neheb:
Anyone with Haiku or Solaris able to test this? @extrowerk

neheb:
My question is really if Solaris supports O_CLOEXEC and friends.

neheb:
https://www.mail-archive.com/bug-gnulib@gnu.org/msg38493.html says that everything except macOS does. Workaround is to use open and fdopen.

neheb:
Reworked to check for macOS instead of linux. This stuff is supported everywhere relevant. That is, every place with at least GCC8 or clang7.

extrowerk:
Tested, builds fine on Haiku.

neheb:
Let's ignore this for now.

PR #2002, opened 2021-10-05
Rework creation of containers

KarlStraussberger:
Use parentId instead of full path.

Removes compatibility of js function addCdsObject (after half a year)


src/content/content_manager.cc
```diff
@@ -925,8 +925,9 @@ void ContentManager::finishScan(const std::shared_ptr<AutoscanDirectory>& adir,
         if (parent && lmt > std::chrono::seconds::zero()) {
             parent->setMTime(lmt);
             int changedContainer;
+            int count = firstObject ? std::distance(firstObject->getLocation().begin(), firstObject->getLocation().end()) - std::distance(location.begin(), location.end()) : 0;
```
neheb, line 928: Uhh what? Why not use size() ?

src/content/layout/builtin_layout.cc
```diff
@@ -51,6 +51,74 @@
...
 #endif //ONLINE_SERVICES
 
+BuiltinLayout::BuiltinLayout(std::shared_ptr<ContentManager> content)
+    : Layout(std::move(content))
+    , config(this->content->getContext()->getConfig())
+    , genreMap(this->config->getDictionaryOption(CFG_IMPORT_SCRIPTING_IMPORT_GENRE_MAP))
+{
+    container["Video"] = std::make_shared<CdsContainer>("Video");
+    container["All Video"] = std::make_shared<CdsContainer>("All Video");
+    chain["/Video/All Video"] = this->content->addContainerTree(std::vector<std::shared_ptr<CdsObject>> { container["Video"], container["All Video"] });
```
neheb, line 61: the vector param can be removed. {} is enough.

I am curious what happens here though. std::initializer_list is a view into an array. not an array itself. So these shared pointers get created and then copied.

I've hit this issue before in this codebase where std::vector {std::make_shared x} warned about a memory leak.

Maybe the proper solution is to make the function parameter an initializer_list. I don't know the ins and outs of initializer_list that well.

neheb: Yeah...

https://godbolt.org/z/7x4Tjc4fP

The commented part does not compile.

src/content/layout/builtin_layout.cc
```diff
@@ -51,6 +51,74 @@
...
+    container["Online Services"] = std::make_shared<CdsContainer>("Online Services");
+#ifdef SOPCAST
+    container["SopCast"] = std::make_shared<CdsContainer>("SopCast");
+    container["SopCast/All Channels"] = std::make_shared<CdsContainer>("All Channels");
+    container["SopCast/Groups"] = std::make_shared<CdsContainer>("Groups");
+    chain["/Online Services/SopCast/All Channels"] = this->content->addContainerTree(std::vector<std::shared_ptr<CdsObject>> { container["Online Services"], container["SopCast"], container["SopCast/All Channels"] });
+    chain["/Online Services/SopCast/Groups"] = this->content->addContainerTree(std::vector<std::shared_ptr<CdsObject>> { container["Online Services"], container["SopCast"], container["SopCast/Groups"] });
+#endif
+
+#ifdef ATRAILERS
```
neheb, line 108: Would it be possible to convert these to if constexpr?

src/content/layout/builtin_layout.cc
KarlStraussberger: Your example uses `unique_ptr`. On the other hand, I'm not creating the `shared_ptr` in line, it's already in the container map.

neheb: It uses unique_ptr to see when copies happen. Well anyway, maybe fanalyzer or clang-analyzer can detect if there's an issue.

src/content/layout/builtin_layout.cc
KarlStraussberger: They are spread all over the codebase.

neheb: Different PR then.

src/database/sql_database.cc
```diff
@@ -1264,14 +1264,17 @@ fs::path SQLDatabase::buildContainerPath(int parentID, const std::string& title)
     return path;
 }
 
-void SQLDatabase::addContainerChain(std::string virtualPath, const std::string& lastClass, int flags, int lastRefID, int* containerID, std::deque<int>& updateID, const std::vector<std::pair<std::string, std::string>>& lastMetadata)
+bool SQLDatabase::addContainer(int parentContainerId, std::string virtualPath, const std::shared_ptr<CdsContainer>& cont, int* containerID)
 {
-    log_debug("Adding container Chain for path: {}, lastRefId: {}, containerId: {}", virtualPath.c_str(), lastRefID, *containerID);
+    log_debug("Adding container Chain for path: {}, lastRefId: {}, containerId: {}", virtualPath.c_str(), cont->getRefID(), *containerID);
+
+    if (parentContainerId == INVALID_OBJECT_ID)
```
whyman, line 1271: I personally think this should bail out instead, you have to explicitly set a root attachment.

src/database/database.h
```diff
@@ -181,9 +181,7 @@ class Database {
     /// the object ID of the container that is last in the path. The
     /// updateID will hold the objectID of the container that was changed,
     /// in case new containers were created during the operation.
-    virtual void addContainerChain(std::string path, const std::string& lastClass, int flags, int lastRefID, int* containerID,
-        std::deque<int>& updateID, const std::vector<std::pair<std::string, std::string>>& lastMetadata)
-        = 0;
+    virtual bool addContainer(int parentContainerId, std::string virtualPath, const std::shared_ptr<CdsContainer>& cont, int* containerID) = 0;
```
whyman, line 184: Is virtual path multi-level? or do we just support a single "name" now? If so, I think renaming path is sensible.

KarlStraussberger: The name is stored in `cont->title`. `virtualPath` is needed to detect existing entries and therefor multilevel.

src/database/sql_database.cc
KarlStraussberger: The only call comes from `ContentManager`, so this can be changed.

src/upnp_xml.cc
```diff
@@ -621,7 +621,8 @@ void UpnpXMLBuilder::addResources(const std::shared_ptr<CdsItem>& item, pugi::xm
 
             tRes->mergeAttributes(tp->getAttributes());
 
-            tRes->addParameter("dlnaProfile", tp->dlnaProfile());
+            if (!tp->dlnaProfile().empty())
+                tRes->addParameter("dlnaProfile", tp->dlnaProfile());
```
KarlStraussberger, line 626: Just dropped in: avoids double / in url

PR #2005, opened 2021-10-06
replace several protected to private

neheb:
Most of the time member variables should be private


whyman:
Sure, but does it make any difference other than making it harder to extend these classes

neheb:
I found an unused variable as a result of this before. Otherwise no.

PR #2009, opened 2021-10-07
remove nullptr check from reference

neheb:
a reference cannot be NULL.


KarlStraussberger:
Sometime thing look different than they are...

PR #2013, opened 2021-10-11
prefix member variable with m_

neheb:
WebKit style says member variables should be prefixed with m_.


whyman:
I guess we should write a code style document, is there some kind of standard for that?

Issue #325, opened 2018-07-03
Gerbera doesn't parse .opus metadata


emk2203:
I have a collection of music files in .opus format (or MIME-Type: audio/ogg, audio/opus). They are properly tagged with Metadata, including album art. Gerbera shows them under "Audio/Full Name", but doesn't parse the files for metadata.

Since they are basically the offspring of the more or less dead .ogg format, parsing the metadata shouldn't be too difficult. Could this enhancement be implemented?

`ffprobe` can show all metadata of the files.

emk2203:
<map from="opus" to="audio/opus"/>

is the mapping I made for `<extension/mimetype ignore-unknown="no">`, and 

     <treat mimetype="audio/opus" as="opus"/>

is the mapping for `<mimetype-contenttype>`.

These are the only changes from the default `/etc/gerbera/config.xml`.

Is there anything I have overlooked?

v00d00:
Where is a good place to get an OPUS album?

emk2203:
The examples lack somewhat - it's important that things like album art and other metadata work as well. I try to give you a link to an album later.

KarlStraussberger:
Thanks, for reporting. It would interesting to reproduce this issue with the current HEAD, because we reworked the SQL code.

Toby-Haynes:
I'm happy to take a Git HEAD branch - is there a "clean" way to keep my current packaged version separate from a debug dev build? Or is it better to backup the database and config.xml and remove the packages?

KarlStraussberger:
If the goal is to reproduce the issue, you may run a docker image which definitely does not leave any traces on your system. In that case it's quite complicated to keep your database and config.

If you build gerbera yourself, it for sure is not reversible in all aspects (additional dev packages etc.). A database backup is always a good idea before an update!

Toby-Haynes:
The docker container helps.

Gerbera does remain functional after attempting to add an opus file. It reports version Gerbera UPnP Server v1.9.2-57-g47ab772b 

```
2021-10-14 16:04:52   info: Gerbera UPnP Server v1.9.2-57-g47ab772b - http://gerbera.io/
2021-10-14 16:04:52   info: Copyright 2016-2021 Gerbera Contributors.
2021-10-14 16:04:52   info: Gerbera is free software, covered by the GNU General Public License version 2
2021-10-14 16:04:52   info: Loading configuration from: /var/run/gerbera/config.xml
2021-10-14 16:04:52   info: Checking configuration...
[...]

PR #2025, opened 2021-10-13
Taglib: Handle OGG containing Opus, Speex or FLAC

whyman:
Previously we assumed that .ogg would only contain vorbis, but this is
often not the case nowadays with the usage of Opus expanding.

So instead we now try to find a supported type and use that.

Fixes: https://github.com/gerbera/gerbera/issues/325

### DIFF
--- a/scripts/debian/build-deb.sh
+++ b/scripts/debian/build-deb.sh
@@ -56,6 +56,12 @@ function install-pupnp() {
   echo "::endgroup::"
 }
 
+function install-taglib() {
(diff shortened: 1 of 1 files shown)


whyman:
Once #2027 and #2028 are merged this will be green.

Not sure what to do here, the 1.12 bump adds this useful functionality. From what I can see no eqivalent exists in 1.11.1

The problem is that you would no longer to be able to use Ubuntu 20.04's libtag package.

PR #2031, opened 2021-10-13
Tidy unused scripts, update docs

whyman:


KarlStraussberger:
I'm fine with dropping `pugixml` but I had to install duktape from source because openSuSE comes with libduktape201 (which is 2.2.1 source)

whyman:
I think pretty much any recentish version of duktape works

PR #2039, opened 2021-10-14
remove time_t stuff

neheb:
Just include fmt/chrono.


KarlStraussberger:
I'm always a bit nervous with time casts, because of former issues we introduced.

neheb:
you mean the std::chrono stuff?

KarlStraussberger:
Yes, "real" time versus "Unix" time. In the case here it does not not affect any logic, so I'm relaxed.

Issue #256, opened 2018-02-24
gerbera does not reap transcoding zombies


cweiske:
Today I saw that a ffmpeg transcoding process started by gerbera was hanging in the zombie state, and gerbera did not reap it:

```
$ ps auxf
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
...
gerbera  14550  0.0  1.9 1597880 149748 ?      Ssl  Jan13  18:32 /usr/local/bin/gerbera -c /etc/gerbera/config.xml
gerbera  24014  0.0  0.0      0     0 ?        Z    07:38   0:00  \_ [ffmpeg] <defunct>
```

Gerbera needs to reap zombies itself, Linux does not do it automatically.

elmodaddyb:
Maybe a new handler for **SIGCHLD**

http://www.microhowto.info/howto/reap_zombie_processes_using_a_sigchld_handler.html

cweiske:
I had restart gerbera today because of this probolem:
```
gerbera   173858  0.0  0.7 1347120 57544 ?       Ssl  Okt16   0:33 /usr/local/bin/gerbera -c /etc/gerbera/config.xml
gerbera   183614  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   183629  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   184254  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   184297  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   184860  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   185473  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   186031  0.0  0.0      0     0 ?        Z    Okt16   0:00  \_ [ffmpeg] <defunct>
gerbera   187059  0.0  0.0      0     0 ?        Z    00:00   0:00  \_ [ffmpeg] <defunct>
gerbera   187627  0.0  0.0      0     0 ?        Z    00:05   0:00  \_ [ffmpeg] <defunct>
[...]

PR #2048, opened 2021-10-18
Allow setting environment variables in transcoding sub processes

KarlStraussberger:
Did solve transcoding issue for me

See also #256


KarlStraussberger:
Yes, that's the problem. If run by systemd they get system defaults which default to `de` on my suse and vlc/ffmpeg can't parse numbers like 1.0 then.

PR #2049, opened 2021-10-19
Replace double-pointer out-param with return value

matthiasstraka:
Looks a lot better with simple return values.


neheb:
Need to fix the test stuff.

PR #2052, opened 2021-10-20
avoid string copying

neheb:

### DIFF
--- a/src/cds_objects.h
+++ b/src/cds_objects.h
@@ -276,7 +276,7 @@ public:
     }
 
     /// \brief Add a single metadata value.
-    void addMetaData(const metadata_fields_t key, const std::string& value)
+    void addMetaData(const metadata_fields_t key, std::string&& value)
     {
         metaData.emplace_back(MetadataHandler::getMetaFieldName(key), value);
     }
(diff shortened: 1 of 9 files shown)


neheb:
a const ref "can" be a copy.

```
addMetaData(0, "test");
```

is a copy.

matthiasstraka:
Why not use 

```cpp
void addMetaData(const metadata_fields_t key, std::string value)
{
    metaData.emplace_back(MetadataHandler::getMetaFieldName(key), std::move(value));
}
```

Then, the caller can still decide if he wants his string to be moved away or not.

KarlStraussberger:
In my experience, restructuring/ refactoring is the more reliable method to gain performance than addressing single parameters. 

As a user of a method I don't want to think about copying or moving. That's why, in most cases I use `const &` because the object in the calling enclosure still exists (you never know) and the called operation can decide on creating a copy or just using the reference. It may cost a bit of time for the extra object, but I don't have to think. I only change this strategy if there are visible performance issues or if the objects grow bigger. In that case I consider refactoring the code and creating some sort of helper class.

Too much optimization renders unreadable and critical code, like using an object that was just move away.

PR #2079, opened 2021-10-30
remove custom merge function

neheb:
std::map::merge() is equivalent.


neheb:
note that there's a test case for this that passes.

whyman:
Good to know the test passes, but the patch should remove the test too, as that method no longer exists.

neheb:
Removed.

whyman:
Thanks!

PR #2080, opened 2021-10-30
Simplify makeFifo, unify with curl and tidy unused

whyman:


neheb:
run_simple_process should be removed from process.h as well.

KarlStraussberger:
You forgot to drop `makeTemp` from `tools.h`

whyman:
@neheb was already done
@KarlStraussberger thanks, done!

PR #2081, opened 2021-10-30
Drop SopCast "support"

whyman:
Not sure this code has worked since MediaTomb days. The SopCast service
is long discontinued. RIP me mateys.


neheb:
Wonder if lastfm works too.

whyman:
Lastfm works

PR #2089, opened 2021-10-31
remove several unused functions

neheb:

### DIFF
--- a/src/action_request.h
+++ b/src/action_request.h
@@ -116,8 +116,6 @@ public:
     /// This function writes all changed values into the upnp_request structure, and
     /// must be called at the very end before giving *upnp_request back to the SDK.
     void update();
-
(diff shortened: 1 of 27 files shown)


src/upnp_cm.cc
```diff
@@ -138,32 +138,3 @@ void ConnectionManagerService::processSubscriptionRequest(const std::unique_ptr<
     ixmlDocument_free(event);
 #endif
 }
-
-void ConnectionManagerService::sendSubscriptionUpdate(const std::string& sourceProtocolCsv)
```
KarlStraussberger, line 142: I'd love to keep that. It looks like there is some UPnP request we do not implement yet

neheb: Kept.

PR #2096, opened 2021-10-31
JS: add array check to getArrayProperty

whyman:
Not sure if this is the correct fix (should there be a string here?) but
it makes the import successful.


KarlStraussberger (approved):
duktape code is alway trial and error for me - if it works it's fine

Issue #2091, opened 2021-10-31
JS fails with autoscan added from UI


whyman:
From master with master JS import, no changes.

```
2021-10-31 11:46:26.935  debug: [sql_database.cc:719] addObject(): Generated insert: INSERT INTO "grb_cds_resource" ("item_id", "res_id", "handlerType", "size", "duration", "bitrate", "sampleFrequency", "nrAudioChannels", "protocolInfo", "sec:acodec") VALUES (3055, 0, 0, '8867675', '0:03:35.954', '41062', '44100', '2', 'http-get:*:audio/mpeg:*', 'mp3')
2021-10-31 11:46:26.935  debug: [sqlite_database.cc:247] exec(): Adding query to Queue: INSERT INTO "grb_cds_resource" ("item_id", "res_id", "handlerType", "size", "duration", "bitrate", "sampleFrequency", "nrAudioChannels", "protocolInfo", "sec:acodec") VALUES (3055, 0, 0, '8867675', '0:03:35.954', '41062', '44100', '2', 'http-get:*:audio/mpeg:*', 'mp3')
2021-10-31 11:46:26.935  debug: [thread_runner.h:135] lockGuard(): ThreadRunner: Guard for SQLiteThread - addTask ExecTask
2021-10-31 11:46:26.935  debug: [thread_runner.h:125] notify(): ThreadRunner: Notifying SQLiteThread
2021-10-31 11:46:26.935  debug: [sqlite_database.cc:482] run(): Running: INSERT INTO "grb_cds_resource" ("item_id", "res_id", "handlerType", "size", "duration", "bitrate", "sampleFrequency", "nrAudioChannels", "protocolInfo", "sec:acodec") VALUES (3055, 0, 0, '8867675', '0:03:35.954', '41062', '44100', '2', 'http-get:*:audio/mpeg:*', 'mp3')
2021-10-31 11:46:26.935  debug: [thread_runner.h:95] wait(): ThreadRunner: Waiting for SQLiteThread
2021-10-31 11:46:26.935  debug: [sql_database.cc:719] addObject(): Generated insert: INSERT INTO "grb_cds_resource" ("item_id", "res_id", "handlerType", "parameters", "resolution", "protocolInfo", "sec:vcodec") VALUES (3055, 1, 2, 'rct=aa', '1000x1000', 'http-get:*:image/jpeg:*', 'mjpeg')
[...]

PR #2097, opened 2021-10-31
fix definition of M_CONTENT_CLASS

KarlStraussberger:
also add missing values for BuildinLayout

this should be the root cause for #2091


whyman:
Ah, that makes sense

Issue #2044, opened 2021-10-15
Default "server alive" configuration


ngosang:
In the documentation is clear the default value for server alive is 180 s. I don't know how but in my configuration file the value is 1800 and I don't remember changing it.
I see there are several files in this repository with alive=1800. Could you review it and make sure default values are consistent with documentation?

* https://github.com/gerbera/gerbera/blob/5cfc26cf421383ff06cf5c2815b5a45badd21bc5/src/config/config_definition.cc#L56
* https://github.com/gerbera/gerbera/search?q=alive+1800

PR #2098, opened 2021-11-01
Send alive messages every 60s by default

whyman:
Previous default was 1800 which resulted in messages every 14.5
minutes.

Add a note on the libupnp behavior to the docs.

Thanks to @ngosang for reporting

Fixes: https://github.com/gerbera/gerbera/issues/2044


KarlStraussberger:
I suppose the second `0` was a long lasting typo.

PR #2106, opened 2021-11-02
restore log level to debug

dyseg:
Causes a lot of noise in the non-debug log.


KarlStraussberger:
I'm sure it dropped in by accident

PR #2107, opened 2021-11-02
Bleeding edge

KarlStraussberger:
Update build scripts to build fairly recent releases of the libs. Add this to the CI checks so we can ensure that gerbera still builds if one or the other distro bumps their version.

Load versions from file `versions.sh`. Which also support an environment variable to build _older_ libs. 

Counter proposal for #2031


whyman:
I think taglib 1.12 is required now for the opus stuff, I like having the versions centralised.

KarlStraussberger:
Oops, I took the version from the old script without thinking

KarlStraussberger:
Will have to squash the PR

PR #2111, opened 2021-11-02
replace all lock_guard with scoped_lock

neheb:
SonarLint reports the latter to be better.

### DIFF
--- a/src/config/client_config.cc
+++ b/src/config/client_config.cc
@@ -58,7 +58,7 @@ ClientConfig::ClientConfig(int flags, std::string_view ip, std::string_view user
 
 void ClientConfigList::add(const std::shared_ptr<ClientConfig>& client, std::size_t index)
 {
-    AutoLock lock(mutex);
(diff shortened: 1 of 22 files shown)


whyman:
Do we not need some changes to pass in mutexes in some places? You can construct a mutexless scoped_lock right? But not a mutexless lock_guard

neheb:
I don't think so.

Anyway, added CTAD commit. Simplifies things I think.

neheb:
Added comment from SonarLint in the first commit.

whyman:
@KarlStraussberger one to keep an eye on and revert if there is trouble.

KarlStraussberger:
I was planning for a smoke test during the weekend anyway.

neheb:
Second commit is faulty. No idea why.

whyman:
So needs a revert?

neheb:
Or a fix. I can't figure out what's wrong with it. Maybe CTAD's not working right.

PR #2113, opened 2021-11-03
several duktake type fixes

neheb:
Added some structured bindings and fixed wrong index type.

### DIFF
--- a/src/content/scripting/js_functions.cc
+++ b/src/content/scripting/js_functions.cc
@@ -71,7 +71,7 @@ duk_ret_t js_addContainerTree(duk_context* ctx)
     std::vector<std::shared_ptr<CdsObject>> result;
     auto length = duk_get_length(ctx, -1);
 
-    for (duk_size_t i = 0; i < length; i++) {
+    for (duk_uarridx_t i = 0; i < length; i++) {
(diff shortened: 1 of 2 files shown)


src/content/scripting/js_functions.cc
```diff
@@ -192,14 +192,14 @@ duk_ret_t js_addCdsObject(duk_context* ctx)
             return 0;
         }
 
-        auto parentId = std::pair(stoiString(containerId), false);
+        auto [id, str] = std::pair(stoiString(containerId), false);
```
KarlStraussberger, line 195: please use `parentId` instead of id and parentStatus `instead` of `str`

src/content/scripting/js_functions.cc
```diff
@@ -91,10 +91,10 @@ duk_ret_t js_addContainerTree(duk_context* ctx)
 
     if (!result.empty()) {
         auto cm = self->getContent();
-        auto containerId = cm->addContainerTree(result);
-        if (containerId.first != INVALID_OBJECT_ID) {
+        auto [id, str] = cm->addContainerTree(result);
```
KarlStraussberger, line 94: please use `containertId` instead of id and `containerStatus` instead of `str`

neheb:
Fixed.

PR #2114, opened 2021-11-03
silence exiv2 warnings

neheb:
These are not important for normal operation.

### DIFF
--- a/src/metadata/exiv2_handler.h
+++ b/src/metadata/exiv2_handler.h
@@ -36,12 +36,16 @@
 
 #include "metadata_handler.h"
 
+#include <exiv2/error.hpp>
+
 /// \brief This class is responsible for reading exif header metadata
 class Exiv2Handler : public MetadataHandler {
 public:
     explicit Exiv2Handler(const std::shared_ptr<Context>& context)
         : MetadataHandler(context)
     {
+        // silence exiv2 messages without debug
+        Exiv2::LogMsg::defaultHandler([](auto, auto s) { log_debug("Exiv2: {}", s); });
     }
     void fillMetadata(const std::shared_ptr<CdsObject>& item) override;
     std::unique_ptr<IOHandler> serveContent(const std::shared_ptr<CdsObject>& item, int resNum) override;


neheb:
On the issue of not silencing this in debug, exiv2 writes everything to stderr by default. A custom log handler can be registered. I chose to do this as there's a similar thing in the ffmpeg code where an empty log handler is registered.

neheb:
Maybe this is better.

KarlStraussberger:
Great change!

count00zero:
In my case that leads to an error when building with -DWITH_EXIV2=1

```
gerbera/src/metadata/exiv2_handler.h:48:38: error: no matching function for call to ‘Exiv2::LogMsg::defaultHandler(Exiv2Handler::Exiv2Handler(const std::shared_ptr<Context>&)::<lambda(auto:3, auto:4)>)’
   48 |         Exiv2::LogMsg::defaultHandler([](auto, auto s) { log_debug("Exiv2: {}", s); });

```

KarlStraussberger:
> 
> 
> In my case that leads to an error when building with -DWITH_EXIV2=1
> 
> ```
> gerbera/src/metadata/exiv2_handler.h:48:38: error: no matching function for call to ‘Exiv2::LogMsg::defaultHandler(Exiv2Handler::Exiv2Handler(const std::shared_ptr<Context>&)::<lambda(auto:3, auto:4)>)’
>    48 |         Exiv2::LogMsg::defaultHandler([](auto, auto s) { log_debug("Exiv2: {}", s); });
> ```

Can you please find out your libexiv2 version?

count00zero:
seems to be:
Found LibExiv2: /usr/lib/x86_64-linux-gnu/libexiv2.so (found version "0.27.3")

neheb:
Function is setHandler, not defaultHandler. Wonder why it didn't fail locally.

Anyway, I'll fix it up when I can. The constructor needs to be defined out of line as well.

KarlStraussberger:
@whyman : Do we have a CI check with EXIV2 enabled?

whyman:
> @whyman : Do we have a CI check with EXIV2 enabled?

Seems like everything is using libexif rather than exiv2

count00zero:
https://github.com/gerbera/gerbera/pull/2121 => that fixed my build error, thx!

PR #2117, opened 2021-11-03
remove const from several unique_ptrs

neheb:
SonarLint reports:

If you use std::unique_ptr<T> const & for a function parameter type, it
means that the function will not be able to alter the ownership of the
pointed-to object by the unique_ptr:

It cannot acquire ownership of the pointed-to object (this would require
a parameter of type std::unique_ptr<T>)
It cannot transfer the object ownership to someone else (this would
require a std::unique_ptr<T> &).
That means the function can only observe the pointed-to object, and in
this case passing a T* (if the unique_ptr can be null) or a T& (if it
cannot) provides the same features, while also allowing the function to
work with objects that are not handled by a unique_ptr (E.G. objects on
the stack, in a vector, or in another kind of smart pointer), thus making
the function more general-purpose.

### DIFF
--- a/src/config/client_config.h
+++ b/src/config/client_config.h
@@ -71,7 +71,7 @@ protected:
 /// \brief Provides information about one manual client.
 class ClientConfig {
 public:
-    ClientConfig();
+    ClientConfig() = default;
 
     /// \brief Creates a new ClientConfig object.
     /// \param flags quirks flags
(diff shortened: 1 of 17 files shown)


whyman:
This will require careful review.

If something is owned by a unique_ptr in some scope it doesnt mean we want to transfer ownership just because a funciton needs to access that object. Passing a reference allows _access_ but without _ownership_.

neheb:
Right. That’s why in database I removed one unique_ptr that didn’t need to be there.

neheb:
Reworked this commit to remove unique_ptrs instead.

So the interesting part here is that several functions were passing const ref wrongly. Once unique_ptr was removed, this became an error. This was mostly the case with functions that call setResponse, which should not be const of course. The amazing part is that this confused both clang-tidy and sonarlint. Wonder if an actual bug gets fixed here...

src/config/client_config.h
```diff
@@ -71,31 +71,31 @@ class ClientConfigList {
...
+    ClientConfig() = default;
 
     /// \brief Creates a new ClientConfig object.
     /// \param flags quirks flags
     /// \param ip ip address
     /// \param userAgent user agent
     ClientConfig(int flags, std::string_view ip, std::string_view userAgent);
 
-    const std::unique_ptr<ClientInfo>& getClientInfo() const { return clientInfo; }
+    const ClientInfo& getClientInfo() const { return clientInfo; }
```
KarlStraussberger, line 82: This can become critical if `ClientConfig` instance is destroyed before the referenced object.

src/database/sql_database.h
```diff
@@ -270,11 +270,11 @@ class SQLDatabase : public Database {
     /* helper for removeObject(s) */
     void _removeObjects(const std::vector<std::int32_t>& objectIDs);
 
-    std::unique_ptr<ChangedContainers> _recursiveRemove(
+    ChangedContainers _recursiveRemove(
         const std::vector<std::int32_t>& items,
         const std::vector<std::int32_t>& containers, bool all);
 
-    virtual std::unique_ptr<ChangedContainers> _purgeEmptyContainers(const std::unique_ptr<ChangedContainers>& maybeEmpty);
+    virtual std::unique_ptr<ChangedContainers> _purgeEmptyContainers(ChangedContainers maybeEmpty);
```
KarlStraussberger, line 277: Looks like a copy is created.

src/database/sql_database.h
```diff
@@ -270,11 +270,11 @@ class SQLDatabase : public Database {
     /* helper for removeObject(s) */
     void _removeObjects(const std::vector<std::int32_t>& objectIDs);
 
-    std::unique_ptr<ChangedContainers> _recursiveRemove(
+    ChangedContainers _recursiveRemove(
```
KarlStraussberger, line 273: If not used exactly in the correcty `move`s, it creates extra copies

src/database/sql_database.h
neheb: Not possible. Copy constructor is implicitly deleted because move constructors are defined (yes I tested this). See: https://en.cppreference.com/w/cpp/language/copy_constructor <-- last part under implicitly deleted.

src/database/sql_database.h
neheb: See above. If you try to copy it will fail to compile.

src/config/client_config.h
neheb: How to test that?

neheb:
Rebased.

neheb:
Got rid of most of these.

PR #2128, opened 2021-11-03
replace constructors with using declarations

neheb:
Simpler.


neheb:
Funny. Doesn't work for ConfigBoolSetup as the last two parameters are swapped.

neheb:
Rebased.

whyman:
Would be nice to fix up ConfigBoolSetup too

PR #2130, opened 2021-11-04
Update man page and help text to reflect the current situation

### LINKED ISSUE #2122: Inconsistent help text and manpage with regard to gerbera config which is searched in home dir.
`gerbera -h` states:

> ```
> [...]
>   -m, --home arg       Search this directory for a .gerbera folder containing
>                        a config file
>   -f, --cfgdir arg     Override name of config folder (.config/gerbera) by
>                        default. -h must also be set.
> [...]
> ```

There are several inconsistencies:

1. - The explanation to the `-m` argument says that `gerbera` is looking for a folder `.gerbera` under the directory which is specified by the `-m` argument,
   - whereas the explanation for the `-f` argument says that it is by default looking dor a folder `.config/gerbera` under the directory which is specified by the `-m` argument (which in fact is what happens in reality).
2. The explanation for the `-f` argument mentions that `-h` must be set, but `-h` prints the help text. I *suppose* that `-m` is meant here instead of `-h`.

The manpage also shows the inconsostent information:

`man 1 gerbera`:

> ```
> [...]
>        -c, --config
>                 By default Gerbera will search for a file named "config.xml" in the ~/.config/gerbera directory. [...]
> 
>        -m, --home
>               [...] then it will look for a .gerbera
>               directory in users home. [...] the .Gerbera
>               directory [...]
> [...]
> ```

Here there is even an inconsistency within the same explanation: Once `.gerbera`, once `.Gerbera` (note that directory names are case sensitive on Unix machines).

### LINKED ISSUE #2123: Manpage and builtin help show a different set of command line options.
The manpage (`man 1 gerbera`) mentions only the following command line options:

`-i -e -p -c -m -f -a -l -D --compile-info --version -h`,

While `gerbera -h` mentions the following (which are more):

`-D -d -u -P -e -p -i -c -m -f -l --compile-info -v -h --create-config --add-file`.

### LINKED ISSUE #2126: pidfile is created after dropping privileges: Cannot create pidfile in standard directory, and misleading wrong error message.
When I start `gerbera --config /etc/gerbera/gerbera.xml --user gerbera --pidfile /var/run/gerbera.pid`, gerbera misleadingly complains *`Pidfile /var/run/gerbera.pid exists.`*:

```
2021-11-03 18:31:12   info: Gerbera UPnP Server v1.9.2-123-g2d03e26c - http://gerbera.io/
2021-11-03 18:31:12   info: Copyright 2016-2021 Gerbera Contributors.
2021-11-03 18:31:12   info: Gerbera is free software, covered by the GNU General Public License version 2
2021-11-03 18:31:12   info: Dropped to User: gerbera
2021-11-03 18:31:12  error: Pidfile /var/run/gerbera.pid exists. It may be that gerbera is already
2021-11-03 18:31:12  error: running or the file is a leftover from an unclean shutdown.
2021-11-03 18:31:12  error: In that case, remove the file before starting gerbera.
```

But it does not exist!, the real case seems to be that user `gerbera` cannot create files in `/var/run`. This breaks when gerbera is to be used as a system-wide daemon following the standard ways how things are done on most systems.

It does _not_ complain if I either let gerbera run as root or if I specify a directory where the `gerbera` user can create files.

* Please, change it that way that the pidfile is created _before_ dropping privileges! (i.e. do a `fork()` and let the parent process retaining root rights, create the pidfile with the PID of the child process and quit itself while the child process dropps privileges), and
* please make the error message reflecting the real case and not beeing misleading.

KarlStraussberger:
closes #2122
closes #2123 
closes #2126 

Hope I got things right

### DIFF
--- a/doc/gerbera.1
+++ b/doc/gerbera.1
@@ -5,7 +5,7 @@
 \\$2 \(la\\$1\(ra\\$3
 ..
 .if \n(.g .mso www.tmac
-.TH gerbera 1 2017-09-05 "" ""
+.TH gerbera 1 2021-11-04 "" ""
 .SH NAME
 gerbera \- UPnP MediaServer
 .SH SYNOPSIS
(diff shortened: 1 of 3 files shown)


whyman:
Code scanning says: File created without restricting permissions High
A file may be created here with mode 0666, which would make it world-writable.

KarlStraussberger:
> 
> 
> Code scanning says: File created without restricting permissions High A file may be created here with mode 0666, which would make it world-writable.

We've got a couple of these already (all `fopen`). The message seems to upgraded to _High_. We should consider some `tools` function to ensapsulate the solution.

whyman:
Yeah we can follow that up with another PR.

Issue #2132, opened 2021-11-06
Documentation not updated on latest


KarlStraussberger:
https://docs.gerbera.io/en/latest

Last successful update was `ebca4487` (oct 25).

whyman:
I'll add you to readthedocs

whyman:
Build log

```
Running Sphinx v1.8.5
loading translations [en]... done
making output directory...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 20 source files that are out of date
updating environment: 20 added, 0 changed, 0 removed
reading sources... [  5%] compile
reading sources... [ 10%] conan
reading sources... [ 15%] config-clients
reading sources... [ 20%] config-extended
/home/docs/checkouts/readthedocs.org/user_builds/gerbera/checkouts/latest/doc/compile.rst:16: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/docs/checkouts/readthedocs.org/user_builds/gerbera/checkouts/latest/doc/compile.rst:16: WARNING: Unknown target name: "readme.md".
/home/docs/checkouts/readthedocs.org/user_builds/gerbera/checkouts/latest/doc/config-clients.rst:9: WARNING: Title underline too short.

``clients``
~~~~~~~~~~
/home/docs/checkouts/readthedocs.org/user_builds/gerbera/checkouts/latest/doc/config-clients.rst:61: WARNING: Inline literal start-string without end-string.

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/gerbera/envs/latest/lib/python3.7/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/gerbera/envs/latest/lib/python3.7/site-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/home/docs/checkouts/readthedocs.org/user_builds/gerbera/envs/latest/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 347, in build_update
    len(to_build))
  File "/home/docs/checkouts/readthedocs.org/user_builds/gerbera/envs/latest/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 360, in build
    updated_docnames = set(self.read())
[...]

PR #2134, opened 2021-11-06
Docs: sphinx 4.2, pin versions, fix warnings

whyman:
Fixes: https://github.com/gerbera/gerbera/issues/2132


whyman:
Also enabled the PR builds so we can see if they start failing, and you get a preview of docs now too! https://gerbera--2134.org.readthedocs.build/en/2134/

PR #2136, opened 2021-11-07
replace fatalHandler with lambda

neheb:


whyman:
Sensible, but I think we should keep the same behaviour as before.

neheb:
k back to the original. throw_std_runtime_error doesn't work right in lambdas anyway (__FUNCTION__ is not correct).

whyman:
Thanks!

PR #2140, opened 2021-11-07
Improve matroska parsing speed

### LINKED ISSUE #2045: Large MKV imports take a long time
I have been using Gerbera for some time and It's working very well for me. Thank you.

The issue I'm experiencing from the first day is the huge amount of time it takes to index the collection.
My collection is not that big ~4.8 TB and ~7000 media files. 
**It takes 7 hours to import and that happens after every restart!**

image
image

Since Gerbera CPU usage is low and the disk read speed is low too (much slower than the disk speed) I don't know where the bottle neck is.
* I run Gerbera 1.9.2 in Docker in a Raspberry Pi 4. The storage is in a external RAID 1 connected with USB 3.0
* I have thumbnails disabled and inotify enabled.
* I'm using sqlite database, current size is just 13 MB for the complete collection. This database is in the RaspberryPi SD card which is slow. ~~I tried to enable transactions but the import speed is the same. **I'm pretty sure the issue is here, I see it takes some time to write the data in de DB if I enable debug logs.**~~

I have some questions:
* It's normal to take 7 hours in importing 7k files?
* Why is required to re-import after restart? NOTE: I'm not sure about this, I just see it sometimes. Maybe the database is corrupted due to bad shutdown.
* Can we speed up importing of already imported files just hashing the file or checking the modification date / size?
* ~~It will be faster if I use MySQL database instead of SQLite?~~
* Do you have any configuration or recommendation to improve my Gerbera?

**UPDATE**: After testing with MYSQL storage I'm not sure about sqlite anymore. It takes almost the same time. The issue I see is the following. The music is imported pretty fast, several songs per second. But, I have a folder with MKV movies (each movie is around 10 GB). It takes 20-30s to import each movie and 5 -10s for TV shows. If you do the math it takes 7 hours indeed. **Are you reading the whole file or something like that?**

Thank you!

KarlStraussberger:
closes #2045 

It's been more complicated than expected because we don't know in which header segment the searched entry is stored

### DIFF
--- a/src/metadata/matroska_handler.cc
+++ b/src/metadata/matroska_handler.cc
@@ -35,6 +35,7 @@
 
 #include <matroska/KaxAttachments.h>
 #include <matroska/KaxContexts.h>
+#include <matroska/KaxSeekHead.h>
 #include <matroska/KaxSegment.h>
 
 #include "cds_objects.h"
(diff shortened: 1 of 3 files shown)


src/metadata/matroska_handler.cc
```diff
@@ -130,10 +130,15 @@ std::unique_ptr<IOHandler> MatroskaHandler::serveContent(const std::shared_ptr<C
     return ioHandler;
 }
 
-void MatroskaHandler::parseMKV(const std::shared_ptr<CdsItem>& item, std::unique_ptr<MemIOHandler>* pIoHandler) const
+#define MATROSKA_TITLE 0x01
```
whyman, line 133: Shouldnt these be defined in LibMatroska somewhere?

whyman:
Per the comment in that issue https://github.com/gerbera/gerbera/issues/2045#issuecomment-946782700

We still have to iterate over the "Cluster" (e.g. gigabytes of movie) to get to the attachments if we dont seek directly using the SeekHead offsets, so I think the problem will remain :(

KarlStraussberger:
> 
> 
> Per the comment in that issue [#2045 (comment)](https://github.com/gerbera/gerbera/issues/2045#issuecomment-946782700)
> 
> We still have to iterate over the "Cluster" (e.g. gigabytes of movie) to get to the attachments if we dont seek directly using the SeekHead offsets, so I think the problem will remain :(

Need to dig into libMatroska docs, which do not really exist.

whyman:
> Need to dig into libMatroska docs, which do not really exist.

Its incredibly bad. It looks like most projects dont actually use those libraries, and you can see why!

KarlStraussberger:
Had to dig really deep but finally it should be close to perfect 🥲

whyman:
Fantastic effort @KarlStraussberger!

So much complexity to deal with, and an ugly API.

KarlStraussberger:
> So much complexity to deal with, and an ugly API.

It's the file structure and the fact that most stuff is optional

PR #2142, opened 2021-11-08
Fix segfaulting on ARM due to (maybe) unintended pointer  

Smartsmurf:
I cross-compiled Gerbera to run on my ARM-powered NAS. However everytime when opening the Config-Section using the web interface the Gerbera process died with a SEGFAULT.
Gdb showed it happens inside Web::ConfigLoad::process() at this routine:

```
//update entries with datebase values
for (auto&& entry : dbEntries) {
        auto exItem = allItems.find(entry.item);
        if (exItem != allItems.end()) {
            auto item = exItem->second;
            setValue(item, entry.value);

```
exItem->second did not point to a valid \<pugi::xml_node\> item structure (just garbage).

For debug reasons I dumped the "allItems" map just before the loop and found out that the map contains distinct values for \<std::string\>, but blockwise a lot of identical \<pugi::xml_node*\> values - all pointing to the locally declared variables "auto item =..." above. But those variables already lost their context at this point and may contain garbage data when entering the lookup database loop.

I fixed the issue (working for me) by changing the pointer type.
Maybe I am mistaken, but I would be happy if someone can look into this issue.

### DIFF
--- a/src/web/config_load.cc
+++ b/src/web/config_load.cc
@@ -66,7 +66,7 @@ void Web::ConfigLoad::addTypeMeta(pugi::xml_node& meta, const std::shared_ptr<Co
 
 void Web::ConfigLoad::createItem(pugi::xml_node& item, const std::string& name, config_option_t id, config_option_t aid, const std::shared_ptr<ConfigSetup>& cs)
 {
-    allItems[name] = &item;
+    allItems[name] = item;
     item.append_attribute("item") = name.c_str();
     item.append_attribute("id") = fmt::format("{:03d}", id).c_str();
     item.append_attribute("aid") = fmt::format("{:03d}", aid).c_str();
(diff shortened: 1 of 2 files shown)


KarlStraussberger:
I'm not sure that this really works. Your creating a copy of the node which is not linked to the overall xml document.

Seems that the original node was created in a scope that is already distroyed. 

Need to check the places where `createItem` is called.

KarlStraussberger:
@Smartsmurf can you please try #2143. It's intended to fix your segfault and keep functionality.

Smartsmurf:
> 
> 
> @Smartsmurf can you please try #2143. It's intended to fix your segfault and keep functionality.

I can confirm, that #2143 fixes the segfault issue mentioned above. Thanks.

KarlStraussberger:
Thank you

PR #2145, opened 2021-11-09
content manager changes

neheb:
Mostly const stuff.


KarlStraussberger:
I don't like the idea of const method like `addObject` because in the end they modify the contents of the database. From the C++ view it does not change the object directly but it changes it's state.

neheb:
Removed a bunch of them.

PR #2146, opened 2021-11-09
remove pointless reference

neheb:
It should actually be const. But no need for it.


KarlStraussberger:
Nice

PR #2147, opened 2021-11-09
const

neheb:


whyman:
Why should these methods be const?

whyman:
https://github.com/gerbera/gerbera/pull/2145#issuecomment-964164559 👍

PR #2150, opened 2021-11-10
use operator < instead of startswith

neheb:
Avoids unnecessary conversions from fs::path, to std::string, to
std::string_view.


whyman:
I think this shows that we should have a `isParent()` or similar method

KarlStraussberger:
> I think this shows that we should have a `isParent()` or similar method
> 
> 

I think, we need to learn the _new_ operators and hope they are more efficient than a function call. :smile:

PR #2153, opened 2021-11-11
remove pointless destructor and virtual member

neheb:
Not really used.


whyman:
Removing the virtual makes sense.

The destructor is no harm and compiled out without debug anyway.

The deleted copy constuctor and operator should stay to make sure that never happens.

neheb:
The reason I added those constructors was because a non-default virtual destructor is present. Without them the class follows the rule of 0.

whyman:
Could we just not make them non-virtual instead?

PR #2155, opened 2021-11-12
matroska_handler: fix wrong field type

neheb:
The ffmpeg handler reads the exact same matroska section but writes it
to M_CREATION_DATE. Match it here.


KarlStraussberger:
Ffmpeg reads both. Additionally M_DATE is used in UPNP responses. So either set both or keep it as it is.

neheb:
hmm? I checked the code. ffmpeg's creation_time routine and libmatroska's KaxDateUTC read the same section of the file. I don't see the ffmpeg code writing creation_time to M_DATE.

KarlStraussberger:
In the absence of ffmpeg we have to set m_date as well.

neheb:
Maybe edit https://github.com/gerbera/gerbera/blob/master/src/metadata/ffmpeg_handler.cc#L153 instead?

KarlStraussberger:
> 
> 
> Maybe edit https://github.com/gerbera/gerbera/blob/master/src/metadata/ffmpeg_handler.cc#L153 instead?

ffmpeg tag this field with creation_date so it's wise to keep that code and add it to matroksa (I already did in #2140)

PR #2158, opened 2021-11-12
rework for loop

neheb:
SonarLint reports that this is a bad loop. Rework it.

for loops are very flexible in C and C++. Because of that, they can carry
a complexity that can make the code error-prone, difficult to understand,
and hard to maintain.

Many for loops can be written in a way that clearly separates the
iteration process from the content of the iteration. This rule makes sure
that all the code relevant to the iteration is placed in the for-loop
header. The compliant code is then easier to reason about.

A for loop is composed of 4 sub-parts:
  for([initialization]; [condition]; [update])    [body]  

We classify the variables used to control them in three categories:

A loop-counter is a variable modified in the update. It should not be
modified in the body.

A loop-constant is an auxiliary variable declared in the initialization.
It’s very often used to precompute some data about the end condition or
the stride.

A pseudo-counter shares some properties with a loop counter, but its
update conditions are more complex. It will therefore only be updated in
the body, and cannot be used in the update. Using a pseudo-counter makes
the loop more complex to reason about, and therefore is not permitted.
They are very often declared in the initialization, for instance, to limit
their scope, but in some cases reuse existing variables.

Additionally, the loop condition should refer to at least one loop-counter,
and should not modify anything.

This rule is only checking for loops with a condition and an update.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -218,19 +218,21 @@ int Server::startupInterface(const std::string& iface, in_port_t inPort)
     const char* ifName = iface.empty() ? nullptr : iface.c_str();
 
     int ret = UPNP_E_INIT_FAILED;
-    for (std::size_t attempt = 0; ret != UPNP_E_SUCCESS; attempt++) {
+    for (std::size_t attempt = 0; attempt < 4; attempt++) {
         ret = UpnpInit2(ifName, inPort);
-        if (ret != UPNP_E_SUCCESS) {
-            if (attempt > 3) {
-                throw UpnpException(ret, fmt::format("run: UpnpInit failed with {} {}", ifName, inPort));
-            }
+        if (ret == UPNP_E_SUCCESS)
+            break;
+
 #ifdef UPNP_HAVE_TOOLS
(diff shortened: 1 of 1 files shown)


src/server.cc
```diff
@@ -217,20 +217,23 @@ int Server::startupInterface(const std::string& iface, in_port_t inPort)
...
 #else
-            log_warning("UPnP Init {}:{} failed: ({}). Retrying in {} seconds...", ifName, inPort, ret, attempt + 1);
+        log_warning("UPnP Init {}:{} failed: ({}). Retrying in {} seconds...", ifName, inPort, ret, attempt + 1);
 #endif
-            std::this_thread::sleep_for(std::chrono::seconds(attempt + 1));
-        }
+        std::this_thread::sleep_for(std::chrono::seconds(attempt + 1));
+    }
+
+    if (attempt > 3) {
```
KarlStraussberger, line 235: It's clearer if you check `ret != sucess` here.

neheb: I think this is what you mean.

PR #2162, opened 2021-11-13
use std::byte

neheb:

### DIFF
--- a/src/util/jpeg_resolution.cc
+++ b/src/util/jpeg_resolution.cc
@@ -31,6 +31,8 @@
 
 #include "tools.h" // API
 
+#include <cstddef>
+
 #include "iohandler/io_handler.h"
 
 using uchar = unsigned char;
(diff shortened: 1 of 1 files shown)


KarlStraussberger:
Code scan throws

This pointer might have type `inotify_event` (size 16), but the pointer arithmetic here is done with type `byte *` (size 1).

neheb:
Right. This PR changes nothing about that. I don't really know what the original code is doing.

src/util/jpeg_resolution.cc
```diff
@@ -59,14 +61,14 @@ using uchar = unsigned char;
...
+static int get16m(std::byte* shrt)
 {
-    return (static_cast<uchar*>(const_cast<void*>(shrt))[0] << 8) | static_cast<uchar*>(const_cast<void*>(shrt))[1];
+    return std::to_integer<int>((shrt[0] << 8) | shrt[1]);
 }
 
 static int iohFgetc(const std::unique_ptr<IOHandler>& ioh)
 {
-    uchar c[1] = { 0 };
+    std::byte c[1] {};
```
KarlStraussberger, line 71: Does this init `c[0]` to 0 (see return statement)?

neheb: Should be identical. int() is equivalent to std::to_integer<int>

neheb:
Removed first commit as that's being handled elsewhere.

PR #2167, opened 2021-11-13
use fs::path for path

neheb:


whyman:
Based on the logic below this can also be the value "--all" which is not a valid path.

PR #2169, opened 2021-11-13
remove some unused macros

neheb:

### DIFF
--- a/src/util/jpeg_resolution.cc
+++ b/src/util/jpeg_resolution.cc
@@ -35,11 +35,6 @@
 
 using uchar = unsigned char;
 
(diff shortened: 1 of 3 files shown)


src/util/jpeg_resolution.cc
```diff
@@ -37,7 +37,6 @@ using uchar = unsigned char;
 
 #ifndef TRUE
 #define TRUE 1
-#define FALSE 0
```
whyman, line 40: All of those can go surely?

neheb: I'll look into it.

PR #2172, opened 2021-11-13
use lambda and fs::path

neheb:
The former allows to avoid reassigning a value. The latter gets rid of
various conversions.

### DIFF
--- a/src/web/web_autoscan.cc
+++ b/src/web/web_autoscan.cc
@@ -44,17 +44,17 @@ void Web::Autoscan::process()
         throw_std_runtime_error("web:autoscan called with illegal action");
 
     bool fromFs = boolParam("from_fs");
-    std::string path;
     std::string objID = param("object_id");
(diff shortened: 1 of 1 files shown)


src/web/config_save.cc
```diff
@@ -163,10 +163,10 @@ void Web::ConfigSave::process()
             if (objectID > 0 && autoscan) {
                 content->rescanDirectory(autoscan, objectID, target);
                 auto taskEl = root.append_child("task");
-                taskEl.append_attribute("text") = fmt::format("Rescanning directory {}", target).c_str();
-                log_info("Rescanning directory {}", target);
+                taskEl.append_attribute("text") = fmt::format("Rescanning directory {}", target.c_str()).c_str();
```
KarlStraussberger, line 166: No need for this, if target is still a string.

KarlStraussberger (changes requested):

PR #2182, opened 2021-11-15
use std::pair for checkFileAndSubtitle

neheb:

### DIFF
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -51,28 +51,6 @@ FileRequestHandler::FileRequestHandler(const std::shared_ptr<ContentManager>& co
 {
 }
 
-static bool checkFileAndSubtitle(fs::path& path, const std::shared_ptr<CdsObject>& obj, const std::size_t& resId, std::string& mimeType, struct stat& statbuf, const std::string& rh)
-{
-    bool isSrt = false;
-
-    if (!rh.empty()) {
-        auto resPath = obj->getResource(resId)->getAttribute(R_RESOURCE_FILE);
-        mimeType = MIMETYPE_TEXT;
-        isSrt = !resPath.empty();
-        if (isSrt) {
-            path = resPath;
-        }
-    }
-    int ret = stat(path.c_str(), &statbuf);
-    if (ret != 0) {
-        if (isSrt) {
-            throw SubtitlesNotFoundException(fmt::format("Subtitle file {} is not available.", path.c_str()));
-        }
-        throw_std_runtime_error("Failed to open {}: {}", path.c_str(), std::strerror(errno));
-    }
-    return isSrt;
-}
-
 void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
 {
     log_debug("start: {}", filename);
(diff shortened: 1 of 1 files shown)


src/file_request_handler.cc
```diff
@@ -53,26 +53,26 @@ FileRequestHandler::FileRequestHandler(std::shared_ptr<ContentManager> content,
 {
 }
 
-static bool checkFileAndSubtitle(fs::path& path, const std::shared_ptr<CdsObject>& obj, const std::size_t& resId, std::string& mimeType, struct stat& statbuf, const std::string& rh)
+static std::pair<bool, struct stat> checkFileAndSubtitle(fs::path& path, const std::shared_ptr<CdsObject>& obj, const std::size_t& resId, std::string& mimeType, std::string_view rh)
```
KarlStraussberger, line 56: No need for string_view here. It's converting a string to a new object.

neheb:
Hmmm this mimeType parameter looks like can be removed. The function sets it based on rh.empty().

KarlStraussberger:
> 
> 
> Hmmm this mimeType parameter looks like can be removed. The function sets it based on rh.empty().

It's an output parameter (no const!)

neheb:
I mean this basically:

```
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -53,13 +53,12 @@ FileRequestHandler::FileRequestHandler(std::shared_ptr<ContentManager> content,
 {
 }

-static std::pair<bool, struct stat> checkFileAndSubtitle(fs::path& path, const std::shared_ptr<CdsObject>& obj, const std::size_t& resId, std::string& mimeType, const std::string& rh)
+static std::pair<bool, struct stat> checkFileAndSubtitle(fs::path& path, const std::shared_ptr<CdsObject>& obj, const std::size_t& resId, const std::string& rh)
 {
     std::pair<bool, struct stat> r;

     if (!rh.empty()) {
         auto resPath = obj->getResource(resId)->getAttribute(R_RESOURCE_FILE);
-        mimeType = MIMETYPE_TEXT;
         r.first = !resPath.empty();
         if (r.first) {
             path = resPath;
@@ -101,8 +100,8 @@ void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
     auto item = std::dynamic_pointer_cast<CdsItem>(obj);

     fs::path path = item ? item->getLocation() : "";
-    std::string mimeType;
-    auto [isSrt, statbuf] = checkFileAndSubtitle(path, obj, resId, mimeType, rh);
+    std::string mimeType = !rh.empty() ? MIMETYPE_TEXT : "";
+    auto [isSrt, statbuf] = checkFileAndSubtitle(path, obj, resId, rh);

     UpnpFileInfo_set_IsReadable(info, access(path.c_str(), R_OK) == 0);


```

KarlStraussberger:
Valid point. But maybe there is a way (or already available) to get the mime type from the resource, like `text/srt`. Could be somewhere in metacontent_handler.

KarlStraussberger:
> 
> 
> Valid point. But maybe there is a way (or already available) to get the mime type from the resource, like `text/srt`. Could be somewhere in metacontent_handler.

It can be taken from resource attribute `R_PROTOCOLINFO`. `getMTFromProtocolInfo` sould do the work

neheb:
Right. That gets done later in the code.

KarlStraussberger:
The best solution is to integrate the function in the getInfo code. I created it some time ago to avoid duplicate code. The second call is gone now, so can the function

neheb:
Done. Hope I didn't make a mistake.

KarlStraussberger:
Couldn't see any flaw. Thank you

PR #2186, opened 2021-11-15
move several methods out of line

neheb:
Avoids -Wweak-vtables. Small size reduction.

### DIFF
--- a/src/content/scripting/import_script.cc
+++ b/src/content/scripting/import_script.cc
@@ -49,6 +49,11 @@ ImportScript::ImportScript(std::shared_ptr<ContentManager> content,
     }
 }
 
+script_class_t ImportScript::whoami()
+{
+    return S_IMPORT;
+}
+
 void ImportScript::processCdsObject(const std::shared_ptr<CdsObject>& obj, const std::string& scriptpath)
 {
     processed = obj;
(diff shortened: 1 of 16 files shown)


whyman:
Thanks, can you add the -Wweak-vtables to CMake too?

src/content/scripting/import_script.cc
```diff
@@ -49,6 +49,11 @@ ImportScript::ImportScript(std::shared_ptr<ContentManager> content,
     }
 }
 
+script_class_t ImportScript::whoami()
+{
+    return S_PLAYLIST;
```
KarlStraussberger, line 54: Must be `S_IMPORT` here

KarlStraussberger (changes requested):

neheb:
> Thanks, can you add the -Wweak-vtables to CMake too?

Not all are fixed or should be fixed.

whyman:
> > Thanks, can you add the -Wweak-vtables to CMake too?
> 
> Not all are fixed or should be fixed.

I don't understand why this would be?

neheb:
Basically, it would require placing out of line definitions for virtual functions. This is somewhat problematic for defaulted destructors. Destructors should always be defaulted in their definitions. Not doing so just increases size for no good reason.

neheb:
updated slightly.

PR #2188, opened 2021-11-15
move filesystem declaration after headers

neheb:

### DIFF
--- a/src/cds_objects.h
+++ b/src/cds_objects.h
@@ -32,11 +32,9 @@
 #ifndef __CDS_OBJECTS_H__
 #define __CDS_OBJECTS_H__
 
-#include <filesystem>
 #include <map>
 #include <memory>
 #include <vector>
-namespace fs = std::filesystem;
 
 #include "cds_resource.h"
 #include "common.h"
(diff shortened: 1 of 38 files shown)


src/util/process_executor.h
```diff
@@ -39,11 +39,15 @@
 #include <filesystem>
 #include <unistd.h>
 
+namespace fs = std::filesystem;
+
 #include "executor.h"
 
+namespace fs = std::filesystem;
```
KarlStraussberger, line 46: Why duplicate it?

src/cds_objects.h
```diff
@@ -36,7 +36,6 @@
 #include <map>
 #include <memory>
 #include <vector>
-namespace fs = std::filesystem;
```
KarlStraussberger, line 39: If the alias is not needed, also the include must be in one of the others. Did you try to remove it as well?

KarlStraussberger (commented):
I recommend a new header `utils/grb_fs.h` that encapsulates the include and the alias. We may move some of the declaration from tools.h/.cc to that file.

neheb:
wouldn't common.h make more sense?

KarlStraussberger:
> 
> 
> wouldn't common.h make more sense?

IHMO this would blow up. We should stick to include what we need.

KarlStraussberger:
Thank you

PR #2191, opened 2021-11-16
Set refId for playlist container

### LINKED ISSUE #2144: m3u playlists driving me crazy
I've mentioned this some time ago in an issue I've opened for slow MySQL performance.
Back then it was just a "mild issue" but now it I'm so frustrated that I'm close to give up gerbera :-(

I'm using a shell script (running on Linux here) which every night scans my complete music library, then creates a randomized m3u file holding 200 titles. Doing this, I can just run the playlist on some client and always have a new mix of music.
But gerbera just does not get it :-(

In theory, it should notice that the m3u file has changed, using Inotify. Works as expected.
But for whatever reason that playlist in gerbera then is not what it is supposed to be. It is some random mix of the current playlist file, the previous one or something in between.

Please, how can this be fixed? I want the previous playlist replaced with the current one. Happens once a day.
I wasn't able to get this to work.

Is there a way to simply delete the current playlist in the gerbera DB, then read the new one maybe? All this obviously in an automated way for my case. When I delete the playlist from the GUI, then create a new one, all seems to be fine.

Thanks!

KarlStraussberger:
fixes #2144

### DIFF
--- a/scripts/js/playlists.js
+++ b/scripts/js/playlists.js
@@ -34,7 +34,7 @@ const chain = {
     allPlaylists: { title: 'All Playlists', objectType: OBJECT_TYPE_CONTAINER, upnpclass: UPNP_CLASS_CONTAINER },
     allDirectories: { title: 'Directories', objectType: OBJECT_TYPE_CONTAINER, upnpclass: UPNP_CLASS_CONTAINER },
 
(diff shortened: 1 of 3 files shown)


whyman:
Nice!

I wonder if we should default to linking the imported items to the playlist, I think its more intuitive.

KarlStraussberger:
Just change the config definition 👍

whyman:
Nevermind it is already enabled by default:

```#define DEFAULT_PLAYLIST_CREATE_LINK YES```

PR #2194, opened 2021-11-16
find_if conversion

neheb:

### DIFF
--- a/src/metadata/ffmpeg_handler.cc
+++ b/src/metadata/ffmpeg_handler.cc
@@ -123,17 +123,14 @@ void FfmpegHandler::addFfmpegMetadataFields(const std::shared_ptr<CdsItem>& item
                 continue; // iterate while loop
             }
         }
-        for (auto&& [fld, key] : propertyMap) {
-            if (key == e->key) {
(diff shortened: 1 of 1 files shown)


src/metadata/ffmpeg_handler.cc
```diff
@@ -123,17 +123,13 @@ void FfmpegHandler::addFfmpegMetadataFields(const std::shared_ptr<CdsItem>& item
...
-            if (key == e->key) {
-                // only use ffmpeg meta data if not found by other handler
-                if (item->getMetaData(fld).empty()) {
-                    log_debug("Identified default metadata '{}': {}", key, value);
-                    field = fld;
-                    item->addMetaData(field, sc->convert(trimString(value)));
-                    break; // leave for loop
-                }
-            }
+        auto pIt = std::find_if(propertyMap.begin(), propertyMap.end(), [=](auto&& p) { return p.second == e->key && item->getMetaData(p.first).empty(); });
```
whyman, line 126: Maybe move the logic onto its own line?

Also probably should be capture by ref instead.

neheb:
Done.

PR #2212, opened 2021-11-19
Cleanup mt_inotify

KarlStraussberger:


src/util/mt_inotify.cc
```diff
@@ -121,13 +121,12 @@ struct inotify_event* Inotify::nextEvent()
     static struct inotify_event* ret;
     static int firstByte = 0;
     static ssize_t bytes;
-
-    int fdMax;
+    static std::byte* eventStart = reinterpret_cast<std::byte*>(&event[0]);
```
neheb, line 124: auto can be used here.

neheb:
What happened with this?

PR #2220, opened 2021-11-19
use dynamic cast for derived classes

neheb:

### DIFF
--- a/src/metadata/matroska_handler.cc
+++ b/src/metadata/matroska_handler.cc
@@ -204,7 +204,9 @@ void MatroskaHandler::parseHead(const std::shared_ptr<CdsItem>& item, IOCallback
 
     EbmlElement* dummyEl;
     int iUpperLevel = 0;
-    auto metaSeek = static_cast<LIBMATROSKA_NAMESPACE::KaxSeekHead*>(info);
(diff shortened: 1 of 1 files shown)


src/metadata/matroska_handler.cc
```diff
@@ -213,14 +215,17 @@ void MatroskaHandler::parseHead(const std::shared_ptr<CdsItem>& item, IOCallback
     }
     for (auto&& seekEl : *metaSeek) {
         if (EbmlId(*seekEl) == LIBMATROSKA_NAMESPACE::KaxSeek::ClassInfos.GlobalId) {
-            auto seekPoint = static_cast<LIBMATROSKA_NAMESPACE::KaxSeek*>(seekEl);
+            auto seekPoint = dynamic_cast<LIBMATROSKA_NAMESPACE::KaxSeek*>(seekEl);
+            if (!seekPoint)
+                return;
```
KarlStraussberger, line 220: It should be a `break` only

KarlStraussberger (changes requested):

src/metadata/matroska_handler.cc
neheb: Fixed.

PR #2221, opened 2021-11-19
fix implicit conversions

neheb:
resId is assumed to be an int elsewhere.


neheb:
hmm maybe not.

PR #2222, opened 2021-11-19
reduce lambda size slightly

neheb:
CLion reports that the lambda is too big. Slim it down slightly.


whyman:
That should just not be a lambda I think is the point, rather than making it less lines 😅

neheb:
IIRC the whole point of using find_if here is to reduce written code. Or maybe it was to avoid some clang-tidy error.

PR #2228, opened 2021-11-21
const ref conversion

neheb:


whyman:
Are we sure the strings live longer than these things?

neheb:
I think the reason for this change was that populateCommandLine takes a const ref parameter.

KarlStraussberger:
It's copied to some param in output.

PR #2231, opened 2021-11-21
ffmpeg: const additions

neheb:


whyman:
Lgtm to me apart from the auto.

neheb:
should probably be const auto.

PR #2232, opened 2021-11-21
grb_fs: run through clang-tidy

neheb:
used std::move

Change return type to uintmax_t as that is the type returned by
file_size().

Switch to fseeko to avoid potential 32-bit problems.


whyman:
Filesize should stay off_t. It changes size depending on platform.

neheb:
Not quite: https://en.cppreference.com/w/cpp/filesystem/file_size

neheb:
64-bit on both

neheb:
Rebased.

PR #2233, opened 2021-11-22
Attempt to close leak

KarlStraussberger:


neheb:
wonder if these should be wrapped with unique_ptr or something.

PR #2248, opened 2021-11-23
fix potential shared_ptr problem

neheb:
CLion reports that when this function returns, it points to a bogus
value.


neheb:
hrm guess not.

neheb:
added proper fix.

KarlStraussberger:
This can be integrated into #2251

PR #2251, opened 2021-11-24
avoid some shared_ptr copies

neheb:

### DIFF
--- a/src/database/search_handler.cc
+++ b/src/database/search_handler.cc
@@ -191,28 +191,27 @@ std::shared_ptr<ASTNode> SearchParser::parseSearchExpression()
     std::stack<std::shared_ptr<ASTNode>> nodeStack;
     std::stack<TokenType> operatorStack;
     std::shared_ptr<ASTNode> root;
-    std::shared_ptr<ASTNode> expressionNode;
     TokenType currentOperator = TokenType::INVALID;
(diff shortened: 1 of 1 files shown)


neheb:
This PR is interesting. unique_ptr might be a better fit here.

src/database/search_handler.cc
```diff
@@ -201,18 +201,18 @@ std::shared_ptr<ASTNode> SearchParser::parseSearchExpression()
                     throw_std_runtime_error("Cannot construct ASTAndOperator without lhs");
                 if (!expressionNode)
                     throw_std_runtime_error("Cannot construct ASTAndOperator without rhs");
-                std::shared_ptr<ASTNode> lhs(nodeStack.top());
+                auto lhs = std::move(nodeStack.top());
                 nodeStack.pop();
-                nodeStack.push(std::make_shared<ASTAndOperator>(sqlEmitter, lhs, expressionNode));
+                nodeStack.push(std::make_shared<ASTAndOperator>(sqlEmitter, std::move(lhs), std::move(expressionNode)));
```
KarlStraussberger, line 206: The change would be more obvious if you move the declaration of `expressionNode` inside the while loop.

neheb: Done.

PR #2262, opened 2021-11-26
remove IOHandlerBufferHelper destructor

neheb:
All derived classes call close() already.


KarlStraussberger:
That's true, but there is no destructor to clean up when `close` was not called in usages

PR #2268, opened 2021-11-27
pass shared_ptr constructors by reference

neheb:
Memory Sanitizer indicates memory leaks related to shared_ptrs passed by
value. Passing by const reference seems to fix most of them. While they
seem to be small, they probably add up over time.

Also smaller compiled size.

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -34,7 +34,7 @@
 #include "upnp_xml.h"
 #include "util/upnp_quirks.h"
 
-ActionRequest::ActionRequest(std::shared_ptr<Context> context, UpnpActionRequest* upnpRequest)
+ActionRequest::ActionRequest(const std::shared_ptr<Context>& context, UpnpActionRequest* upnpRequest)
     : upnp_request(upnpRequest)
     , actionName(UpnpActionRequest_get_ActionName_cstr(upnpRequest))
     , UDN(UpnpActionRequest_get_DevUDN_cstr(upnpRequest))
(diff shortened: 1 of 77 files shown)


whyman:
Surely we want to copy them so ref count gets increased, that's how shared_ptr works?

neheb:
This does not prevent copying. It probably reduces them though.

neheb:
Alright small update: I ran modernize-pass-by-value on top of this PR which basically reverts some of the changes. I verified with address sanitizer that the memory leaks that were present are now gone. The net result is basically:

```
~/d/g/addr (1)> rg direct\ leak mem | wc -l
304
SUMMARY: AddressSanitizer: 899986 byte(s) leaked in 10412 allocation(s).
~/d/g/addr (1)> rg direct\ leak mem2 | wc -l
SUMMARY: AddressSanitizer: 357193 byte(s) leaked in 5458 allocation(s).
84
```

The remaining memory leaks seem somewhat difficult to fix. They also seem small. Most of them have to do with libebml:

```
 Direct leak of 340928 byte(s) in 5327 object(s) allocated from:
     #0 0x941c27 in operator new(unsigned long) (/home/mangix/devstuff/gerbera/addr/gerbera+0x941c27)
     #1 0x7ff5e80a9584  (/lib64/libebml.so.5+0x16584)
 
Direct leak of 1440 byte(s) in 20 object(s) allocated from:
    #0 0x941c27 in operator new(unsigned long) (/home/mangix/devstuff/gerbera/addr/gerbera+0x941c27)
    #1 0x7ff5e80a4744  (/lib64/libebml.so.5+0x11744)
```

KarlStraussberger:
For me the result is a lot more readable. It seems to follow a clear line:
- We always pass an object as `const &`. If the called method wants to store a copy in its scope it creates the copy itself. For `shared_ptr` this increases the ref count etc. We don't need moves and don't loose object.
- If a constructor needs a complex object, it may create it directly in its argument list and move it to the target property (c.f. `GrbFile`). This avoids an extra copy.

@whyman  We should keep this somewhere as guideline for arguments.

PR #2270, opened 2021-11-28
remove reserve

neheb:
Memory Sanitizer reports a memory leak here. Probably because of the
function call.


whyman:
Can you post the leak output?

neheb:
Strange. I don't see it anymore. Probably the shared_ptr changes fixed it.

PR #2274, opened 2021-11-30
init shared_ptrs directly

neheb:


KarlStraussberger:
That PR adds addition dependencies to the header which increase compile time.

PR #2278, opened 2021-11-30
use fs::path

neheb:

### DIFF
--- a/src/web/config_save.cc
+++ b/src/web/config_save.cc
@@ -150,7 +150,7 @@ void Web::ConfigSave::process()
     }
 
     // trigger rescan of database after update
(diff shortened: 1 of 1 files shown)


src/web/config_save.cc
```diff
@@ -150,7 +150,7 @@ void Web::ConfigSave::process()
     }
 
     // trigger rescan of database after update
-    std::string target = param("target");
+    fs::path target = param("target");
     if (action == "rescan" && !target.empty()) {
         if (target != "--all") {
             fs::path targetPath(target);
```
KarlStraussberger, line 156: The path is created here. The param may be "--all" which is not a path!

PR #2281, opened 2021-11-30
std::string_view conversions

neheb:


src/metadata/metadata_handler.cc
```diff
@@ -234,7 +234,7 @@ std::string MetadataHandler::mapContentHandler2String(int ch)
     return "Unknown";
 }
 
-metadata_fields_t MetadataHandler::remapMetaDataField(const std::string& fieldName)
+metadata_fields_t MetadataHandler::remapMetaDataField(std::string_view fieldName)
```
whyman, line 237: No reference?

neheb: AFAIK string_view is trivially copyable.

PR #2284, opened 2021-11-30
shared to unique_ptr conversions

neheb:
No copies are needed.

### DIFF
--- a/src/database/search_handler.cc
+++ b/src/database/search_handler.cc
@@ -177,20 +177,20 @@ void SearchParser::getNextToken()
     currentToken = lexer->nextToken();
 }
 
-std::shared_ptr<ASTNode> SearchParser::parse()
+std::unique_ptr<ASTNode> SearchParser::parse()
 {
     getNextToken();
     if (currentToken->getType() == TokenType::ASTERISK)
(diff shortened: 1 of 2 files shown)


src/database/sql_database.cc
```diff
@@ -966,7 +966,7 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::browse(BrowseParam& param)
             for (std::size_t count = 0; count < dynContent->size(); count++) {
                 auto dynConfig = dynContent->get(count);
                 if (parent->getLocation() == dynConfig->getLocation().parent_path() || (parent->getLocation().empty() && dynConfig->getLocation().parent_path() == "/")) {
-                    int dynId = -(parent->getID() + 1) * 10000 - count;
```
KarlStraussberger, line 969: If you cast count to `int` the warning should be gone as well and the code is clearer.

Anyway, you accidently added this commit twice.

KarlStraussberger (changes requested):

src/database/sql_database.cc
neheb: Still complains. Will remove the second commit.

neheb:
Removed second commit and rebased.

PR #2286, opened 2021-12-01
use auto with cast

neheb:
Avoids bugprone-implicit-widening-of-multiplication-result warning.


neheb:
btw removing the inner int64_t cast does not solve the warning.

KarlStraussberger:
Does casting the 10000 help?

neheb:
So clang-tidy suggests two fixes.

```
auto dynId = std::int32_t(std::uint64_t(-(parent->getID() + 1) *) 10000 - count);
```

and

```
auto dynId = std::int32_t(std::int64_t(-(parent->getID() + 1)) * 10000 - count);
```

I think the latter is better. Probably the issue is that count is a size_t and getID returns an int.

PR #2287, opened 2021-12-01
convert string to fs::path

neheb:
Avoids conversions

### DIFF
--- a/src/content/scripting/script.cc
+++ b/src/content/scripting/script.cc
@@ -360,7 +360,7 @@ void Script::defineFunctions(const duk_function_list_entry* functions)
     duk_pop(ctx);
 }
 
-void Script::_load(const std::string& scriptPath)
+void Script::_load(const fs::path& scriptPath)
 {
     std::string scriptText = GrbFile(scriptPath).readTextFile();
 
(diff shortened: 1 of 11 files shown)


src/url_request_handler.cc
```diff
@@ -70,11 +70,11 @@ void URLRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
         UpnpFileInfo_set_FileLength(info, -1);
     } else {
 #ifdef ONLINE_SERVICES
-        std::string url = item->getFlag(OBJECT_FLAG_ONLINE_SERVICE) ? OnlineServiceHelper::resolveURL(item) : item->getLocation().string();
+        fs::path url = item->getFlag(OBJECT_FLAG_ONLINE_SERVICE) ? OnlineServiceHelper::resolveURL(item) : item->getLocation();
```
KarlStraussberger, line 73: It's a URL so it starts with `http(s)://`. We should not declare that as path.

src/content/onlineservice/online_service_helper.cc
```diff
@@ -36,7 +36,7 @@
 #include "config/config_manager.h"
 #include "online_service.h"
 
-std::string OnlineServiceHelper::resolveURL(const std::shared_ptr<CdsItemExternalURL>& item)
+fs::path OnlineServiceHelper::resolveURL(const std::shared_ptr<CdsItemExternalURL>& item)
```
KarlStraussberger, line 39: It's a URL so it starts with `http(s)://`. We should not declare that as path.

src/url_request_handler.cc
neheb: What's interesting is getLocation() returns an fs::path

KarlStraussberger: It's the special situation of external items which inherit from CdsObject we have to live with, but we should make it an URL here (is there some additional lib for that?).

PR #2292, opened 2021-12-04
convert all headers to use pragma once

neheb:
Easier that having to match names.


whyman:
This is not standardised, so lets not use it

PR #2293, opened 2021-12-04
use operator< instead of creating temporaries

neheb:


neheb:
```
Could NOT find GTest (missing: GTEST_LIBRARY GTEST_MAIN_LIBRARY)
```

PR #2297, opened 2021-12-04
simplify some if statements

neheb:

### DIFF
--- a/src/metadata/ffmpeg_handler.cc
+++ b/src/metadata/ffmpeg_handler.cc
@@ -357,10 +357,10 @@ std::unique_ptr<IOHandler> FfmpegHandler::serveContent(const std::shared_ptr<Cds
     if (!config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_ENABLED))
         return nullptr;
 
-    const auto cacheEnabled = config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_CACHE_DIR_ENABLED);
-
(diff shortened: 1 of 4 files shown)


src/metadata/ffmpeg_handler.cc
```diff
@@ -357,13 +357,11 @@ std::unique_ptr<IOHandler> FfmpegHandler::serveContent(const std::shared_ptr<Cds
...
 
-    const auto cacheEnabled = config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_CACHE_DIR_ENABLED);
-
-    if (cacheEnabled) {
-        if (auto data = readThumbnailCacheFile(item->getLocation())) {
-            log_debug("Returning cached thumbnail for file: {}", item->getLocation().c_str());
-            return std::make_unique<MemIOHandler>(data->data(), data->size());
-        }
+    auto cacheEnabled = config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_CACHE_DIR_ENABLED);
+    auto data = readThumbnailCacheFile(item->getLocation());
```
KarlStraussberger, line 361: You cannot call read... if cache is not enabled. Costs time and it may cause exception on cache file access!

KarlStraussberger (changes requested):
ffmpeg changes are breaking

src/metadata/ffmpeg_handler.cc
neheb: I noticed that too. I'll fix.

neheb: Fixed.

PR #2298, opened 2021-12-04
replace findinotify with KDE's version

neheb:
Seems to be better maintained.

Modified to support SunOS as done before.


whyman:
Would be nice to have a target if we are going to change it

neheb:
KDE's version does not have a target.

OTOH, it's available on most distros as extra-cmake-modules or some variation.

neheb:
Anyway, https://invent.kde.org/frameworks/extra-cmake-modules/-/merge_requests/209

neheb:
I added a target thing. No idea if I did it right. Copy paste from the iconv module.

cmake/FindInotify.cmake
```diff
@@ -0,0 +1,66 @@
...
+                Inotify_INCLUDE_DIRS
+        )
+        mark_as_advanced(Inotify_LIBRARIES Inotify_INCLUDE_DIRS)
+        include(FeatureSummary)
+        set_package_properties(Inotify PROPERTIES
+            URL "https://github.com/libinotify-kqueue/"
+            DESCRIPTION "inotify API on the *BSD family of operating systems."
+        )
+    endif()
+  if(NOT TARGET Inotify::Inotify)
```
whyman, line 57: Can you fix the whitespace?

Looks good

neheb: What whitespace?

PR #2303, opened 2021-12-06
simplify shared_ptr assignment

neheb:
Avoids nullptr assignment.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -493,10 +493,9 @@ int main(int argc, char** argv, char** envp)
 
         installSignalHandler();
 
-        std::shared_ptr<Server> server;
+        auto server = std::make_shared<Server>(configManager);
+
         try {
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -470,13 +470,13 @@ int main(int argc, char** argv, char** envp)
 
         log_debug("Datadir is: {}", dataDir.value_or("unset"));
 
-        std::shared_ptr<ConfigManager> configManager;
+        auto configManager = std::make_shared<ConfigManager>(
+            configFile.value_or(""), home.value_or(""), confdir.value_or(""),
```
KarlStraussberger, line 474: ConfigManager constructor throws `std::runtime_error` in case the file is not readable. This is not catched by your change.

KarlStraussberger (changes requested):

src/main.cc
neheb: Removed.

PR #2304, opened 2021-12-06
pair conversion

neheb:
Avoids having to use a non const ref


KarlStraussberger:
Nice!

PR #2309, opened 2021-12-06
const ref conversion

neheb:

### DIFF
--- a/src/transcoding/transcode_ext_handler.cc
+++ b/src/transcoding/transcode_ext_handler.cc
@@ -51,7 +51,7 @@
 #endif
 
 std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(const std::shared_ptr<TranscodingProfile>& profile,
-    fs::path location, const std::shared_ptr<CdsObject>& obj, const std::string& range)
+    const fs::path& location, const std::shared_ptr<CdsObject>& obj, const std::string& range)
 {
     log_debug("Start transcoding file: {}", location.c_str());
     if (!profile)
@@ -77,11 +77,12 @@ std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(const std::sha
 #endif
 
     std::vector<std::shared_ptr<ProcListItem>> procList;
+    fs::path fifoLocation;
 
     bool isURL = obj->isExternalItem();
     if (isURL && !profile->acceptURL()) {
 #ifdef HAVE_CURL
-        openCurlFifo(location, procList);
+        fifoLocation = openCurlFifo(location, procList);
 #else
         throw_std_runtime_error("Compiled without libcurl support, data proxying for profile {} is not available", profile->getName());
 #endif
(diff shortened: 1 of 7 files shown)


src/transcoding/transcode_ext_handler.cc
```diff
@@ -51,7 +51,7 @@
 #endif
 
 std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(const std::shared_ptr<TranscodingProfile>& profile,
-    fs::path location, const std::shared_ptr<CdsObject>& obj, const std::string& range)
```
KarlStraussberger, line 54: You remove the side effect that location points to the temp fifo after the call. I'm quite sure this is needed for everything to work.

neheb: wouldn't that need a non const reference?

KarlStraussberger: Good point 😄

src/transcoding/transcode_ext_handler.cc
```diff
@@ -77,11 +77,12 @@ std::unique_ptr<IOHandler> TranscodeExternalHandler::serveContent(const std::sha
 #endif
 
     std::vector<std::shared_ptr<ProcListItem>> procList;
+    fs::path loc2 = location;
 
     bool isURL = obj->isExternalItem();
     if (isURL && !profile->acceptURL()) {
 #ifdef HAVE_CURL
-        openCurlFifo(location, procList);
+        openCurlFifo(loc2, procList);
```
KarlStraussberger, line 85: In that case change openCurlFifo to fs::path openCurlFifo(const fs::path& location, ..) so that the call is `auto fifoLocation = openCurlFifo(location, procList);` make the code more obvious.

neheb: Reworked. I think this is what you meant.

neheb:
As a sidenote, converting to non const reference also requires this:

```
--- a/src/url_request_handler.cc
+++ b/src/url_request_handler.cc
@@ -118,18 +118,18 @@ std::unique_ptr<IOHandler> URLRequestHandler::open(const char* filename, enum Up

     auto item = std::static_pointer_cast<CdsItemExternalURL>(obj);
 #ifdef ONLINE_SERVICES
-    std::string url = item->getFlag(OBJECT_FLAG_ONLINE_SERVICE) ? OnlineServiceHelper::resolveURL(item) : item->getLocation().string();
+    fs::path url = item->getFlag(OBJECT_FLAG_ONLINE_SERVICE) ? OnlineServiceHelper::resolveURL(item) : item->getLocation().string();
 #else
-    std::string url = item->getLocation();
+    fs::path url = item->getLocation();
 #endif
-    log_debug("Online content url: {}", url);
+    log_debug("Online content url: {}", url.c_str());

     std::string trProfile = getValueOrDefault(params, URL_PARAM_TRANSCODE_PROFILE_NAME);
     if (!trProfile.empty()) {
         auto tp = config->getTranscodingProfileListOption(CFG_TRANSCODING_PROFILE_LIST)
                       ->getByName(trProfile);
         if (!tp)
-            throw_std_runtime_error("Transcoding of file {} but no profile matching the name {} found", url, trProfile);
+            throw_std_runtime_error("Transcoding of file {} but no profile matching the name {} found", url.c_str(), trProfile);

         auto trD = std::make_unique<TranscodeDispatcher>(content);
         auto ioHandler = trD->serveContent(tp, url, item, "");
```

this was declined before.

Issue #2305, opened 2021-12-06
BufferedIOHandler exception on threadStart


KarlStraussberger:
found a core file from gerbera in the container, but this exceeds my skills:
```
Core was generated by `gerbera --port 49494 --config /var/run/gerbera/config.xml'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0xb6eff2da in pthread_mutex_lock () from /lib/ld-musl-armhf.so.1
[Current thread is 1 (LWP 35)]
(gdb) bt
#0  0xb6eff2da in pthread_mutex_lock () from /lib/ld-musl-armhf.so.1
#1  0x005f8e36 in __gthread_mutex_lock (__mutex=0xb0)
    at /usr/include/c++/10.3.1/armv7-alpine-linux-musleabihf/bits/gthr-default.h:749
#2  std::mutex::lock (this=0xb0) at /usr/include/c++/10.3.1/bits/std_mutex.h:100
#3  std::unique_lock<std::mutex>::lock (this=0xb324b968) at /usr/include/c++/10.3.1/bits/unique_lock.h:138
#4  std::unique_lock<std::mutex>::unique_lock (__m=..., this=0xb324b968) at /usr/include/c++/10.3.1/bits/unique_lock.h:68
#5  ThreadRunner<std::condition_variable, std::mutex>::uniqueLock (this=0x0) at /gerbera_build/src/util/thread_runner.h:146
#6  BufferedIOHandler::threadProc (this=0xb33fe5a0) at /gerbera_build/src/iohandler/buffered_io_handler.cc:76
#7  0x00591b20 in operator() (arg=<optimized out>, __closure=0x0) at /gerbera_build/src/iohandler/io_handler_buffer_helper.cc:179
#8  _FUN () at /gerbera_build/src/iohandler/io_handler_buffer_helper.cc:181
#9  0x00507370 in ThreadRunner<std::condition_variable, std::mutex>::startThread()::{lambda(void*)#1}::operator()(void*) const (
    this=0x0, arg=<optimized out>) at /gerbera_build/src/util/thread_runner.h:198
#10 ThreadRunner<std::condition_variable, std::mutex>::startThread()::{lambda(void*)#1}::_FUN(void*) ()
    at /gerbera_build/src/util/thread_runner.h:196
#11 0xb6eff0a4 in ?? () from /lib/ld-musl-armhf.so.1
```

[...]

PR #2314, opened 2021-12-06
Reduce Docker image size

ngosang:
Final Docker image size is reduced from 209 MB => 126 MB (uncompressed)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY . .
 RUN mkdir build && \
     cd build && \
     cmake .. \
-        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DCMAKE_BUILD_TYPE=Release \
         -DWITH_MAGIC=YES \
         -DWITH_MYSQL=YES \
         -DWITH_CURL=YES \
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ RUN mkdir build && \
     cd build && \
     cmake .. \
         -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_CXX_FLAGS=-g1 \
         -DWITH_MAGIC=YES \
         -DWITH_MYSQL=YES \
         -DWITH_CURL=YES \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY . .
 RUN mkdir build && \
     cd build && \
     cmake .. \
-        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DCMAKE_BUILD_TYPE=Release \
```
KarlStraussberger, line 15: It's smaller, but there is no way to debug a core if the server crashes like in #2305.

We may set up a whole group of images without debugging symbols.

ngosang: Official packages in Debian, Ubuntu, Archlinux... don't include debug symbols.
* Is it possible to reduce the size of Debug symbols? I mean, the binary is just 3MB and debug symbols are 100 MB...
* If the previous point is not possible I think it's a good idea to release 2 Docker images. Keep the current tag for slim version and add debug suffix for the fat one eg:`1.9.2-debug`

ngosang:
@KarlStraussberger take a look at `glevel` => https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html
do you think `-g1` will provide enough information for debugging? Could you make it crash and see what happens?

1. `-DCMAKE_BUILD_TYPE=RelWithDebInfo` (glevel = 2 by default)
gerbera exec => 112 MB
2. `-DCMAKE_BUILD_TYPE=Release && -DCMAKE_CXX_FLAGS=-g1` (glevel = 1)
gerbera exec => 28 MB
3. `-DCMAKE_BUILD_TYPE=Release` (glevel = 0, no debug symbols)
gerbera exec => 4.1 MB

KarlStraussberger:
@ngosang glevel=1 sounds great. Enough to see where it crashed. No intention to debug from the cores.

ngosang:
@KarlStraussberger I think this is ready.
BTW I noticed that there are some packages in the final image that are not required in runtime. The size decreases from 126 MB to 119 MB and Gerbera starts correctly:
* util-linux
* sqlite
* zlib
* fmt
* ffmpeg-libs
* libebml
I can open another PR with this change if you are interested.

KarlStraussberger:
From my perspective they should be needed.

There is a reasonable request to add ffmpeg and even vlc for transcoding.

ngosang:
Ok, you can merge.

KarlStraussberger:
Thanks

PR #2317, opened 2021-12-09
set by move

neheb:
Avoids sonarlint warning about using string_view.

### DIFF
--- a/src/util/xml_to_json.h
+++ b/src/util/xml_to_json.h
@@ -45,8 +45,8 @@ class Xml2Json {
 public:
     class Hints {
     public:
-        void setArrayName(const pugi::xml_node& node, const std::string& name) { asArray[node] = name; }
-        void setFieldType(const std::string& node, const std::string& type) { asType[node] = type; }
+        void setArrayName(const pugi::xml_node& node, std::string_view name) { asArray.insert_or_assign(node, name); }
+        void setFieldType(const std::string& node, std::string_view type) { asType.insert_or_assign(node, type); }
 
     private:
         std::map<pugi::xml_node, std::string> asArray;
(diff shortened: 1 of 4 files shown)


KarlStraussberger:
Can you please check the usages. It may happen that string_view is usable here (all calls have const strings?).

neheb:
not sure what you mean here. change the map to contains string views?

KarlStraussberger:
The map must have strings but the parameter can be string_view if all usages use constants. So the string in the map is created just once.

neheb:
All usages of these functions are C strings.

neheb:
hrm maybe went overboard.

src/util/xml_to_json.h
```diff
@@ -45,22 +45,22 @@ class Xml2Json {
...
-        void setArrayName(const pugi::xml_node& node, const std::string& name) { asArray[node] = name; }
-        void setFieldType(const std::string& node, const std::string& type) { asType[node] = type; }
+        void setArrayName(const pugi::xml_node& node, std::string_view name) { asArray[node] = name; }
+        void setFieldType(std::string_view node, std::string_view type) { asType[node] = type; }
 
     private:
-        std::map<pugi::xml_node, std::string> asArray;
-        std::map<std::string, std::string> asType;
+        std::map<pugi::xml_node, std::string_view> asArray;
+        std::map<std::string_view, std::string_view> asType;
```
KarlStraussberger, line 53: Keep these as strings, to create copies. `string_view` stores raw pointers which can become invalid.

KarlStraussberger (changes requested):
`string_view` should only be used for parameters we know to be constants. Storage should use `string`

PR #2320, opened 2021-12-09
remove unused include

neheb:

### DIFF
--- a/src/web/config_load.cc
+++ b/src/web/config_load.cc
@@ -31,7 +31,6 @@
 #include "config/client_config.h"
 #include "config/config_definition.h"
 #include "config/config_setup.h"
-#include "config/directory_tweak.h"
 #include "config/dynamic_content.h"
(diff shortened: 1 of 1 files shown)


src/web/config_load.cc
```diff
@@ -31,7 +31,6 @@
 #include "config/client_config.h"
 #include "config/config_definition.h"
 #include "config/config_setup.h"
-#include "config/directory_tweak.h"
```
KarlStraussberger, line 34: It's needed for https://github.com/gerbera/gerbera/blob/e3a66e274fdb7f4de46d877fbf3579c4e273f43e/src/web/config_load.cc#L210

May happen that some other include get's it.

neheb: That's probably what's happening.

KarlStraussberger: So we should keep it for clarity (others are there as well)

PR #2323, opened 2021-12-10
fmt c++20 fixes

neheb:


neheb:
hrm requires bumping fmt to version 8

neheb:
removed fmt::runtime stuff for now.

PR #2266, opened 2021-11-26
add missing ixmlDocument_free call

neheb:
Address Sanitizer reports a memory leak without it.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

PR #2325, opened 2021-12-10
Delay release of ixmlDocument

KarlStraussberger:
Avoid early ixmlDocument_free.
Make sure object is not needed anymore.

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -108,6 +108,8 @@ void ActionRequest::update()
         if (err != IXML_SUCCESS) {
             log_error("ActionRequest::update(): could not convert to iXML");
             UpnpActionRequest_set_ErrCode(upnp_request, UPNP_E_ACTION_FAILED);
+            if (result)
+                ixmlDocument_free(result);
         } else {
             log_debug("ActionRequest::update(): converted to iXML, code {}", errCode);
             UpnpActionRequest_set_ActionResult(upnp_request, result);
--- a/src/server.cc
+++ b/src/server.cc
@@ -337,6 +337,8 @@ void Server::shutdown()
     timer->shutdown();
     timer = nullptr;
 
+    context = nullptr;
+
     mime = nullptr;
     clients = nullptr;
 }


whyman:
Wouldnt a standard `shared_ptr` work here?

KarlStraussberger:
> Wouldnt a standard `shared_ptr` work here?

The raw pointer is set to some libpupnp internal property which we do not see. It is used after the scope (our area of control) is left.

So the idea is to queue the raw pointer and have it released later.

neheb:
this fails compilation with NPUPNP which does not use IXML. Needs to be ifdef'd out.

```
../src/context.h:85:49: error: unknown type name 'IXML_Document'
    void delayRelease(const std::string source, IXML_Document* doc)
                                                ^
../src/context.h:102:16: error: use of undeclared identifier 'IXML_Document'
    FiFoBuffer<IXML_Document> delay_release_buffer;

```

whyman:
My understanding is that `Context` is pretty much only deleted when `Server` is deleted, and realistically that only happens when we shut down.

So I am wondering if this is an improvement as the allocations hang around still?

It feels like an issue for LibUPnP

src/upnp_mrreg.cc
```diff
@@ -131,7 +132,7 @@ void MRRegistrarService::processSubscriptionRequest(const std::unique_ptr<Subscr
         config->getOption(CFG_SERVER_UDN).c_str(),
```
whyman, line 132: So, `UpnpAcceptSubscriptionExt` this calls `genaInitNotifyExt` which calls  `genaInitNotifyCommon`, which says "We take ownership of propertySet and will free it"

So I think we should only free this if we get an error back.

src/upnp_mrreg.cc
```diff
@@ -160,6 +161,6 @@ void MRRegistrarService::prcoessSubscriptionUpdate(std::string sourceProtocol_CS
             config->getOption(CFG_SERVER_UDN).c_str(),
             serviceID.c_str(), event);
```
whyman, line 163: Same here, this is passed into `genaNotifyAllCommon` eventually, which will free it. Should be freed on error though.

src/upnp_mrreg.cc
KarlStraussberger: I'm sorry, but from the code perspective the above mentioned `propertySet` is a `DOMString` which is `propertySet = ixmlPrintNode((IXML_Node *)PropSet)`.

But contrary to `UpnpActionRequest_set_ActionResult` the content of `PropSet` can be freed right after the call because it is copied to the `DOMString`.

For actionResult there is a `handle_invoke_action` in `soap_device.cc` which does `ixmlDocument_free(actionResultDoc);` So the scan result in #2266 is misleading.

whyman: Ah, yep, you are right, I skimmed the code to hastily

KarlStraussberger:
Well, there is not much left of the initial change set

PR #2332, opened 2021-12-15
add throw_fmt_system_error define

neheb:
Avoids having to call strerror manually.


src/util/mt_inotify.cc
```diff
@@ -60,14 +60,14 @@ Inotify::Inotify()
 #endif
 {
     if (inotify_fd < 0)
-        throw_std_runtime_error("Unable to initialize inotify");
+        throw_fmt_system_error("Unable to initialize inotify");
```
KarlStraussberger, line 63: Sure this sets `errno`?

src/util/grb_fs.cc
```diff
@@ -155,7 +155,7 @@ void GrbFile::writeTextFile(std::string_view contents)
 
     std::size_t bytesWritten = std::fwrite(contents.data(), 1, contents.length(), fd);
     if (bytesWritten < contents.length()) {
-        throw_std_runtime_error("Error writing to {}: {}", path.c_str(), std::strerror(errno));
```
whyman, line 158: Why removing this?

KarlStraussberger: Should be `throw_fmt_system_error`

neheb: Yes it should

src/util/mt_inotify.cc
neheb: man page says they do.

src/util/grb_fs.cc
neheb: wait... fwrite does not set errno.

neheb:
rebased.

PR #2339, opened 2021-12-23
replace various tests with static_assert

neheb:
No need to test these at runtime.


whyman:
I guess you want ASSERT_TRUE/FALSE

neheb:
That's just a normal assert.

whyman:
Okay, EXPECT_TRUE then

neheb:
replaced with ASSERT

KarlStraussberger:
The advantage of `EXPECT...` is that the test framework handles stuff and reports test failures correctly. Please switch over to the proper methods.

PR #2340, opened 2021-12-23
make all mutexes mutable

neheb:
Allows using them in const member functions.


KarlStraussberger:
Nice

PR #2345, opened 2021-12-29
turn getDLNAT stuff into lambdas

neheb:
They're used in one place each.


whyman:
I personally think we should keep the named methods here. The methods encapsulate and name the logic, which is more human friendly.

PR #2346, opened 2021-12-29
lambda conversion

neheb:
Allows scoping some variables and simplifying by using return.


KarlStraussberger:
I'd prefer that whole labda in a separate function. Makes the code more readable.

neheb:
like this?

KarlStraussberger:
Yes, thank you.

PR #2349, opened 2021-12-29
deque removal

neheb:
No benefit over vector here. Small size reduction.

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -1071,7 +1071,7 @@ std::pair<int, bool> ContentManager::addContainerTree(const std::vector<std::sha
     }
     std::string tree;
     int result = CDS_ID_ROOT;
-    std::deque<int> createdIds;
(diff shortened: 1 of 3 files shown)


whyman:
Im 99% sure you converted this from `vector` to `deque`

neheb:
At the time push_front was being used. It seems that since then it has been reworked to not so that.

cweiske:
Doesn't that make #2351, #2350 and this here useless?
Just keep the semantically correct types and do not care how they are implemented under the hood, since this can change tomorrow.

neheb:
There's no benefit to deque here. It's slower than vector. push and pop_front were removed at some point.

KarlStraussberger:
I don't see the need for deque here but remind everybody to improve code by functionality and resource consumption (time and storage) as well. Everybody's time is too valuable to discuss a change in container types. Anyway, even small changes in the container type may break things.

PR #2350, opened 2021-12-29
remove std::stack

neheb:
No real benefit over vector. Small size reduction.

### DIFF
--- a/src/database/search_handler.cc
+++ b/src/database/search_handler.cc
@@ -25,7 +25,6 @@
 #include "search_handler.h" // API
 
 #include <algorithm>
-#include <stack>
 
 #include "database/sql_database.h"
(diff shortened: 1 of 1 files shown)


whyman:
Does using a stack for a stack not make sense?

KarlStraussberger:
In my opinion `stack` semantics are clearer to understand

neheb:
std::stack is implemented on top of vector. No real difference besides a size increase.

neheb:
> In my opinion `stack` semantics are clearer to understand

I actually disagree. When I first converted to vector, I was getting runtime crashes. Turns out top needed to be replaced with back. It makes more sense to me this was by having back everywhere.

KarlStraussberger:
It's used as a stack, it's called stack, so we keep it a stack

PR #2351, opened 2021-12-29
replace queue with vector

neheb:


whyman:
For what reason?

neheb:
The former is implemented in terms of the latter. No real difference besides a size increase.

whyman:
For humans its clearer to use a `queue` when you are dealing with a queue.

KarlStraussberger:
It's used as a queue so it should be a queue.

PR #2356, opened 2021-12-30
remove erase usage

neheb:
Allows conversion to a range based loop.

Also add some std::move to avoid copies.

I *think* this code is equivalent.


neheb:
hmm never mind.

PR #2358, opened 2021-12-31
Refactor file_request_handler

whyman:
Reverts the changes to cache stuff across get_info()/open() calls, this
is not feasible because libupnp may not call open() in the case of HEAD
requests, giving us no opportunity to clean up resources.

This sucks because we have to parse/load everything twice per request,
but it should stop transcoding processes being leaked.

Fixes: #2139

### DIFF
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -55,90 +55,91 @@ void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
 {
     log_debug("start: {}", filename);
 
-    const struct sockaddr_storage* ctrlPtIPAddr = UpnpFileInfo_get_CtrlPtIPAddr(info);
-    // HINT: most clients do not report exactly the same User-Agent for UPnP services and file request.
-    std::string userAgent = UpnpFileInfo_get_Os_cstr(info);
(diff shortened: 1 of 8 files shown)


src/file_request_handler.cc
```diff
@@ -55,90 +55,91 @@ void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
...
+        if (isResourceFile) {
             path = resPath;
         }
+    } else {
+        path = item->getLocation();
     }
+
+    // Check the path (if its real) is accessible
+    struct stat statbuf {
+    };
```
neheb, line 89: I think this should be one line?

whyman: It was but clang format complained

KarlStraussberger:
Good to have a fix, bad to duplicate

KarlStraussberger:
Happy New Year

KarlStraussberger:
As far I can see we broke fanart display. Something to fix this year.

whyman:
Yeah I suspected as much, I couldn't test album art as my NAS was/is having issues.

I'll try to take a look later today

whyman:
See #2359

PR #2359, opened 2022-01-01
Drop resourceHandler concept

whyman:
Nowadays thanks to Karls great work we no longer have the virtual
resource mess to deal with as things like album art now live in
the DB, so we can simplify this code a bit.

Fixes broken artwork.

In fact with this change the `rct` handler is also ignored, just
`res_id/1` is enough to get the image.

### DIFF
--- a/src/file_request_handler.cc
+++ b/src/file_request_handler.cc
@@ -53,35 +53,36 @@ FileRequestHandler::FileRequestHandler(const std::shared_ptr<ContentManager>& co
 
 void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
 {
-    log_debug("start: {}", filename);
+    log_debug("Start: {}", filename);
 
     auto quirks = getQuirks(info);
 
     auto params = parseParameters(filename, LINK_FILE_REQUEST_HANDLER);
     auto obj = loadObject(params);
 
-    auto [resourceHandler, resourceId] = parseResourceInfo(params);
+    auto resourceId = parseResourceInfo(params);
 
-    // If its not an Item and there is no handler, we cant service this request.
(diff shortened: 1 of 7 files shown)


whyman:
Still need to validate things like subs and container art work as expected. I should write some tests too.

src/metadata/ffmpeg_handler.cc
```diff
@@ -255,7 +255,6 @@ void FfmpegHandler::addFfmpegResourceFields(const std::shared_ptr<CdsItem>& item
             std::string thumbMimetype = it != mappings.end() && !it->second.empty() ? it->second : "image/jpeg";
 
             auto ffres = std::make_shared<CdsResource>(CH_FFTH);
-            ffres->addParameter(RESOURCE_HANDLER, fmt::to_string(CH_FFTH));
```
whyman, line 258: This is set in the constructor anyway.

src/file_request_handler.cc
```diff
@@ -105,16 +106,8 @@ void FileRequestHandler::getInfo(const char* filename, UpnpFileInfo* info)
...
-    // so we can not load such a resource for a particular item, we will have
-    // to trust the resource handler parameter
-    if (!resourceHandler.empty()) {
-        if (resourceId == CH_DEFAULT || resourceId >= obj->getResourceCount()) {
-            throw_std_runtime_error("Requested resource not found");
-        }
-
-        auto resource = obj->getResource(resourceId);
-        auto metadataHandler = getResourceMetadataHandler(obj, resourceHandler, resource);
+    if (resourceId > 0) {
```
KarlStraussberger, line 109: You should be able to use `resource->getHandlerType() != CH_DEFAULT` instead. Resource number should not be significant anymore.

Issue #2326, opened 2021-12-11
HTTP interface not available when using hostname


cweiske:
I have a host name entry in my DNS server `mediatomb.home.cweiske.de` that points to the server running gerbera.

Using the browser to access gerbera's web interface via the IP address works: `http://192.168.3.3:49152`.
Using the host name does not work: `http://mediatomb.home.cweiske.de:49152/`

curl says:
```
url -vi http://mediatomb.home.cweiske.de:49152
*   Trying 192.168.3.3:49152...
* Connected to mediatomb.home.cweiske.de (192.168.3.3) port 49152 (#0)
> GET / HTTP/1.1
> Host: mediatomb.home.cweiske.de:49152
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```

This did work with 1.9.2 and earlier, but the git version since 2 or 4 weeks is broken.

Gerbera version: `v1.9.2-274-gffb6e173`

whyman:
I suspect you have upgraded your libupnp version?

ngosang:
Same issue here.
Gerbera master ffb6e173c065e8833a970d8905cb6886e45ac681 
PUPNP=1.14.12

```
curl -vvv http://myserv:49494/
*   Trying 192.168.1.191:49494...
* Connected to myserv (192.168.1.191) port 49494 (#0)
> GET / HTTP/1.1
> Host: myserv:49494
> User-Agent: curl/7.80.0
> Accept: */*
> 
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server


curl -vvv http://192.168.1.191:49494/
*   Trying 192.168.1.191:49494...
* Connected to 192.168.1.191 (192.168.1.191) port 49494 (#0)
> GET / HTTP/1.1
> Host: 192.168.1.191:49494
> User-Agent: curl/7.80.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< CONTENT-LENGTH: 25834
< Accept-Ranges: bytes
< CONTENT-TYPE: text/html
< DATE: Sat, 11 Dec 2021 16:47:59 GMT
< LAST-MODIFIED: Sat, 13 Nov 2021 13:07:40 GMT
< SERVER: Linux/5.10.63-v7l+, UPnP/1.0, Portable SDK for UPnP devices/1.14.12
< X-User-Agent: redsonic
[...]

PR #2365, opened 2022-01-03
Implement libupnp hostname validation

whyman:
Set a `UpnpSetHostValidateCallback` to validate that host names match a
configured virtualURL. Produces a warning if requests do not match with link
to the docs to make debugging non-loading UIs more obvious.

Interestingly, libupnp currently seems to have a bug where the returned
UPNP_E_BAD_HTTPMSG is not actually respected, so for now this change
will allow all hostnames until the libupnp bug is fixed...

Fixes: #2326


KarlStraussberger:
You need at least libupnp 1.14.10 or so.

whyman:
Sigh, `1.14.6` is the earliest. Conan doesn't seem to have anything newer than `1.14.2` :cry: .

neheb:
I assume libnpupnp would also not work.

whyman:
> I assume libnpupnp would also not work.

It should do, this API was actually invented by them and backported

neheb:
Oh interesting.

neheb:
```
../src/server.cc:200:5: error: ‘UpnpSetHostValidateCallback’ was not declared in this scope
  200 |     UpnpSetHostValidateCallback(

```

whyman:
Interestingly for some reason in libnpupnp its called `UpnpSetWebRequestHostValidateCallback`.

I wonder if our friend @medoc92 would mind providing a compatible alias to avoid another ifdef

medoc92:
I'll add an alias.

whyman:
Conan bump PR: https://github.com/conan-io/conan-center-index/pull/8645

PR #2366, opened 2022-01-03
convert pointer to reference

neheb:
Nothing is null or being modified.


KarlStraussberger:
In the end you get the address of the reference so this is no real improvement: It is a pointer returned from UPnP call and should be handled this way throughout the calls.

PR #2372, opened 2022-01-04
replace make_unique with pointer_cast

neheb:
the last parameter is a shared_ptr. This happens to reduce compiled size
quite a bit.


neheb:
Oh interesting. reinterpret_pointer_cast was introduced in C++17. Maybe that's why it wasn't used before.

neheb:
Incredible. This was only added to libcxx in 2020... https://github.com/llvm/llvm-project/commit/ce195fb22b542c20745e3ed46d1516e8b4e3a76f

neheb:
Fixed FreeBSD compilation. Turns out the version of libcxx is too old.

whyman:
Perhaps lets keep the single line rather than 5 lines of ifdefs

neheb:
I mean...I could have just kept the top version only.

PR #2374, opened 2022-01-05
remove various shared_ptr copies

neheb:

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -386,7 +386,7 @@ void ContentManager::addVirtualItem(const std::shared_ptr<CdsObject>& obj, bool 
     addObject(obj, true);
 }
 
-std::shared_ptr<CdsObject> ContentManager::createSingleItem(const fs::directory_entry& dirEnt, const fs::path& rootPath, bool followSymlinks, bool checkDatabase, bool processExisting, bool firstChild, const std::shared_ptr<CMAddFileTask>& task)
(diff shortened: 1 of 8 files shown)


neheb:
address sanitizer shows no memory leaks.

src/content/content_manager.cc
```diff
@@ -351,7 +351,9 @@ std::deque<std::shared_ptr<GenericTask>> ContentManager::getTasklist()
         return taskList;
 
     taskList.push_back(std::move(t));
-    std::copy_if(taskQueue1.begin(), taskQueue1.end(), std::back_inserter(taskList), [](auto&& task) { return task->isValid(); });
+    for (auto&& task : taskQueue1)
+        if (task->isValid())
+            taskList.push_back(std::move(task));
```
KarlStraussberger, line 356: Doesn't the have an effect on the items in `taskQueue1`?

neheb: yeah they get removed and added to taskList.

whyman: I think Carl is saying that as std::move doesn't actually move stuff, its just a cast, the items remain in the original collection

neheb: hrm I don't know how true that is. The way I did this part is by temporarily making taskQueue1 and 2 unique_ptrs. It errored on std::copy_if. This loop fixed that error.

neheb:
Got rid of the offending section. Please review.

PR #2375, opened 2022-01-06
make getValueOrDefault more generic

neheb:
Allows usage with more containers, including array.


src/util/tools.h
```diff
@@ -234,19 +234,22 @@ class EnumIterator {
...
+        return defval;
+    return it->second;
 }
 
-template <typename K, typename V>
-V getValueOrDefault(const std::map<K, V>& m, const K& key, const V& defval)
-{
-    auto it = m.find(key);
-    return (it == m.end()) ? defval : it->second;
-}
```
KarlStraussberger, line 249: Does it work also, if you leave this? I don't like the `constexpr` - it needs explanation like 
```
// if container is map, use map methods instead
```

neheb: if constexpr is specifically made to work with templates.

neheb: I tried to make it more generic (if find() method exists, use that). Couldn't find a simple way to do that. On that note, it seems only std::map has one.

Issue #2377, opened 2022-01-09
Transcoding is broken again


kyak:
Transcoding is broken once again (see #2341). This is not a very good sign. Either the code base is so fragile, or there is a lack of testing or attention to the transcoding feature.

The last known working release is 71b8827abd85fd7352040959bbf68b08855b4ecc. The 9d23b5b7415767bf5744418ccc93725db63dab3e is already not working. The latest master is not working as well.

I will attach the debug log shortly.

whyman:
Hey @kyak, thanks for using master and making these reports, we appreciate it and it improves Gerbera for everyone.

kyak:
Attached is the log file.

[gerb.log](https://github.com/gerbera/gerbera/files/7835518/gerb.log)

whyman:
I can't see many possibly related commits, my guess would be https://github.com/gerbera/gerbera/commit/91dce72f62a2dc857b88b93e808854d12f884a2e, based only on the files changed

kyak:
> I can't see many possibly related commits, my guess would be [91dce72](https://github.com/gerbera/gerbera/commit/91dce72f62a2dc857b88b93e808854d12f884a2e), based only on the files changed

There are not so many commits, just 10 of them. And the one you refer to is older than the one that actually works.

whyman:
Hah, you are correct, apologies

kyak:
This is the faulty commit: 1ec9b7ce64c5cf2bc81ae1449dd26c4a055e31d2. Its parent, 71b8827abd85fd7352040959bbf68b08855b4ecc, works fine as I mentioned.

KarlStraussberger:
@kyak Yes, the old code is a total mess that's why we need to rework it with the risk of breaking stuff. In hope of some friendly folk to find the issues. Thanks

@whyman What about returning `-99` for transcoding https://github.com/gerbera/gerbera/blob/4ff062f408e0bb3acc3190a23ff16a11bc2bb6bc/src/file_request_handler.cc#L240 in order to make sure it is handled correctly

whyman:
I tested that transcoding worked with that change though and indeed for mj profiles it's working in master. 

So I am wondering what edge caae I am missing here

whyman:
So looking at your log it looks like transcoding is getting invoked correctly, and testing on master I get transcoded content.

[...]

PR #2379, opened 2022-01-10
Only set DNLA headers for resources we know about

whyman:
Possible fix for https://github.com/gerbera/gerbera/issues/2377


KarlStraussberger:
So we avoid setting content header for transcoded streams.

Issue #1498, opened 2021-06-01
Support for transcoding in docker


hmoffatt:
Could you add ffmpeg into the docker image so that transcoding could be enabled?

(Maybe vlc too but it doesn't seem to be available on armhf in the Alpine repository.)

jwatzk:
+1 vote

PR #2387, opened 2022-01-14
Docker: Add transcoding tag

whyman:
For now includes ffmpeg, vlc and heic-tools
Adds about about 100MB but saves people having to roll their own.

Fixes: #1498


whyman:
hmm, well this a new error I havent seen before! Worked fine on my local machine.

PR #2389, opened 2022-01-15
Feature/dockerfile

tiescher:
Hi all.

I would like to introduce a change to the Dockerfile to be able to easily change the base image.
That would be very helpful for building on other platforms, eg. like ARMv7 with this command w/out modifying the Dockerfile:

  buildah bud -t gerbera:test --build-arg BASE_IMAGE=arm32v7/alpine:3.15 . 

Thanks in advance

tiescher


KarlStraussberger:
Thank you for your contribution.

whyman (approved):

tiescher:
Thanks for accepting this change ... and for all the great contributions and efforts you have done to provide such a great UPNP streamer.

whyman:
BTW, we do already have ARMv7 and ARM64 images on Dockerhub :+1:

tiescher:
Yes, saw these ... but I prefer to build my own ones 👍 ...keep me on touch with podman and buildah

PR #2394, opened 2022-01-29
replace make_unique with pointer_cast

neheb:
the last parameter is a shared_ptr. This happens to reduce compiled size
quite a bit.


neheb:
Anything missing here?

KarlStraussberger:
It's unreadable to me.

neheb:
Right. This is supposed to be reinterpret_pointer_cast. libc++ versions less than 10 do not support it.

whyman:
Code is for humans, lets keep the clearer version.

PR #2399, opened 2022-01-31
Block negative track numbers

### LINKED ISSUE #2398: Crash with js layout when track number is -1
If I use  `<virtual-layout type="js">` and the track number of one my music is `-1`, I get this error:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoi
```

I have seen this issue on 1.9.2 and on master, but I think it didn't exist on revision 1.7.0.

If I comment all calls to `addCdsObject()`, the problem disappears.

I tried to trick `obj.meta[M_TRACKNUMBER]`, but it seems it is not sufficient.

KarlStraussberger:
detects overflow from taglib

fix #2398

### DIFF
--- a/src/cds_objects.h
+++ b/src/cds_objects.h
@@ -416,7 +416,11 @@ public:
     std::string getMimeType() const { return mimeType; }
 
     /// \brief Sets the upnp:originalTrackNumber property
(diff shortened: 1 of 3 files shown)


whyman:
How odd!

PR #2403, opened 2022-02-05
remove mimetype parameter

neheb:
As a result of the last commit, it is unused.

### DIFF
--- a/src/content/layout/builtin_layout.cc
+++ b/src/content/layout/builtin_layout.cc
@@ -482,7 +482,7 @@ std::string BuiltinLayout::mapGenre(const std::string& genre)
     return genre;
 }
 
-void BuiltinLayout::processCdsObject(const std::shared_ptr<CdsObject>& obj, const fs::path& rootpath, const std::string& mimetype, const std::string& contentType)
+void BuiltinLayout::processCdsObject(const std::shared_ptr<CdsObject>& obj, const fs::path& rootpath, const std::string& contentType)
 {
     log_debug("Process CDS Object: {}", obj->getTitle());
     auto clone = CdsObject::createObject(obj->getObjectType());
@@ -507,17 +507,18 @@ void BuiltinLayout::processCdsObject(const std::shared_ptr<CdsObject>& obj, cons
     } else {
 #endif
 
-        if (startswith(mimetype, "video"))
-            addVideo(clone, rootpath);
-        else if (startswith(mimetype, "image"))
-            addImage(clone, rootpath);
-        else if (startswith(mimetype, "audio") && (contentType != CONTENT_TYPE_PLAYLIST))
-            addAudio(clone, rootpath);
-        else if (contentType == CONTENT_TYPE_OGG) {
+        auto objCls = std::static_pointer_cast<CdsItem>(obj)->getClass();
+        if (contentType == CONTENT_TYPE_OGG) {
             if (obj->getFlag(OBJECT_FLAG_OGG_THEORA))
                 addVideo(clone, rootpath);
             else
                 addAudio(clone, rootpath);
+        } else if (objCls == UPNP_CLASS_VIDEO_ITEM) {
+            addVideo(clone, rootpath);
+        } else if (objCls == UPNP_CLASS_IMAGE_ITEM) {
+            addImage(clone, rootpath);
+        } else if (objCls == UPNP_CLASS_MUSIC_TRACK && contentType != CONTENT_TYPE_PLAYLIST) {
+            addAudio(clone, rootpath);
         }
 
 #ifdef ONLINE_SERVICES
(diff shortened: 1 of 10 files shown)


KarlStraussberger:
Did you consider the effect on mimetypes like `application/ogg`?

neheb:
I did not. Does this code not result in the same behavior?

src/content/layout/builtin_layout.cc
```diff
@@ -507,11 +507,12 @@ void BuiltinLayout::processCdsObject(const std::shared_ptr<CdsObject>& obj, cons
...
+        auto objCls = std::static_pointer_cast<CdsItem>(obj)->getClass();
+        if (objCls == UPNP_CLASS_VIDEO_ITEM)
             addVideo(clone, rootpath);
-        else if (startswith(mimetype, "image"))
+        else if (objCls == UPNP_CLASS_IMAGE_ITEM)
             addImage(clone, rootpath);
-        else if (startswith(mimetype, "audio") && (contentType != CONTENT_TYPE_PLAYLIST))
+        else if (objCls == UPNP_CLASS_MUSIC_TRACK && contentType != CONTENT_TYPE_PLAYLIST)
             addAudio(clone, rootpath);
         else if (contentType == CONTENT_TYPE_OGG) {
```
KarlStraussberger, line 517: This has to be the first clause otherwise class==video/music catches OGG

src/metadata/metadata_handler.cc
```diff
@@ -112,7 +113,7 @@ void MetadataHandler::extractMetaData(const std::shared_ptr<Context>& context, c
 #endif
 
 #ifdef HAVE_FFMPEG
-    if (contentType != CONTENT_TYPE_PLAYLIST && ((contentType == CONTENT_TYPE_OGG && item->getFlag(OBJECT_FLAG_OGG_THEORA)) || startswith(item->getMimeType(), "video") || startswith(item->getMimeType(), "audio"))) {
+    if (contentType != CONTENT_TYPE_PLAYLIST && ((contentType == CONTENT_TYPE_OGG && item->getFlag(OBJECT_FLAG_OGG_THEORA)) || itemCls == UPNP_CLASS_MUSIC_TRACK || itemCls == UPNP_CLASS_VIDEO_ITEM)) {
```
KarlStraussberger, line 116: This can be reduced to

``if (contentType != CONTENT_TYPE_PLAYLIST && (itemCls == UPNP_CLASS_MUSIC_TRACK || itemCls == UPNP_CLASS_VIDEO_ITEM))``

src/content/layout/builtin_layout.cc
KarlStraussberger: If we can rely on objectClass this may go completely

KarlStraussberger (changes requested):
There are another usages of `startswith` in ffmpeg_handler.cc which can be handled in the same way.

src/metadata/metadata_handler.cc
whyman: Would a the itemCls checks on their own even be enough?

KarlStraussberger: > Would a the itemCls checks on their own even be enough?

If mime type to itemCls is complete as it is with default config. 

Makes everything clearer: we set item class in the beginning and then handle files according to their class. At the moment esp. the check for ogg is done very often - only to confirm that video is handled as video.

neheb:
I messed up so I squashed both commits. Fixed all comments.

neheb:
```

FAILED TESTS:
  Gerbera DirTweaks
    submitDirTweak()
      ✖ collects all the form data from the dirtweak modal to call the server
        Chrome Headless 97.0.4692.99 (Linux x86_64)
      Unhandled promise rejection: [object Object] thrown

```

hmmmm

src/metadata/ffmpeg_handler.cc
```diff
@@ -241,7 +241,7 @@ void FfmpegHandler::addFfmpegResourceFields(const std::shared_ptr<CdsItem>& item
     }
 
 #if defined(HAVE_FFMPEG) && defined(HAVE_FFMPEGTHUMBNAILER)
-    if (config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_ENABLED) && (startswith(item->getMimeType(), "video") || item->getFlag(OBJECT_FLAG_OGG_THEORA))) {
+    if (config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_ENABLED) && (item->getClass() == UPNP_CLASS_VIDEO_ITEM || item->getFlag(OBJECT_FLAG_OGG_THEORA))) {
```
KarlStraussberger, line 244: You can drop the check for the flag because it is used to detect a video

neheb: Done.

KarlStraussberger (approved):

PR #2404, opened 2022-02-05
fix memory leak

neheb:
Using unique_ptr with a ternary operator seems to allocate both pointers
and only one ends up getting freed. Changed to lambda to avoid this.

Found with clang-tidy.

### DIFF
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -632,7 +632,11 @@ std::pair<bool, int> UpnpXMLBuilder::insertTempTranscodingResource(const std::sh
             if (tp->hideOriginalResource())
                 hideOriginalResource = true;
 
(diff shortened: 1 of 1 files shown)


neheb:
I tried this on godbolt:

```
void a(bool t)
{
    auto b = t ? std::make_unique<std::string_view>("haha") : std::make_unique<std::string_view>("hmmm");
    auto c = std::make_unique<std::string_view>("haha");
}
```

comment out auto c and no code gets generated. auto b generates code:

```
        sub     rsp, 8
        mov     edi, 16
        call    operator new(unsigned long)
        mov     esi, 16
        add     rsp, 8
        mov     rdi, rax
        jmp     operator delete(void*, unsigned long)
```

PR #2406, opened 2022-02-05
fix bugs in addFfmpegMetadataFields()

mlaerd:
I found 2 bugs in FfmpegHandler::addFfmpegMetadataFields().

1. Because both propertyMap.second and FFMpeg::metadata->key are `const char *`, we need to use `strcmp()` instead of `==`.
2.  If the metadata which is not included in `propertyMap` is found after this method processed M_TRACKNUMBER (or M_PARTNUMBER), `item->setTrackNumber(stoiString(value));` is called again incorrectly because `field` is still `M_TRACKNUMBER`.

I fixed the bugs and confirmed metadata of DSF file are set to DB correctly.

### DIFF
--- a/src/metadata/ffmpeg_handler.cc
+++ b/src/metadata/ffmpeg_handler.cc
@@ -111,6 +111,7 @@ void FfmpegHandler::addFfmpegMetadataFields(const std::shared_ptr<CdsItem>& item
         std::pair(M_COMPOSER, "composer"),
     };
     while ((e = av_dict_get(pFormatCtx->metadata, "", e, AV_DICT_IGNORE_SUFFIX))) {
+        std::string key = e->key;
         std::string value = e->value;
         auto it = specialPropertyMap.find(e->key);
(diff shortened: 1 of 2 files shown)


neheb:
First bug was introduced with https://github.com/gerbera/gerbera/commit/3f5ccd68bacc24c489b23ab4a7ecf1b5483bd7b7

I wonder if it would make more sense to turn the pair from const char * to std::string_view. Would allow == to work again.

mlaerd:
@neheb 
Thanks for your comment.
I found [std:string provides operator== for const char*](https://stackoverflow.com/questions/2931704/how-to-compare-string-with-const-char) and updated the code.
I confirmed it works.

mlaerd:
I want to rerun checks. Could someone help me?

neheb:
git rebase origin/master
git push --force

mlaerd:
@neheb 
Thanks a lot!

whyman:
Nice patch, thanks!

Issue #2401, opened 2022-02-01
DSD sent as PCM 352800 Hz


fandangos:
I've been digging this issue with DLNA and DSD for a while. 
This isn't just a Gerbera issue, other software have the same problem, including ReadyDLNA, Serviio, Tonido, Twonky, except the ancient UMS (Universal Media Server). 

I'm aware there's no guideline to streaming DSD over DLNA and Gerbera uses the safe path of sending it as audio/x-dsd, even if the content is DSF and DFF. 

The problem is that renderers like JRiver, Bubble UPnP, USB Audio Player Pro (on android) read the metadata information from the server that I'm guessing uses ffmpeg that reads on the DSF/DFF files and put out at the max PCM it has available as 352khz to output this content to DSD capable DACs.
So, DSD is converted to PCM even when it should not. 

The workaround I found was to change the profile to audio/x-dsf and to show it as *.dsf. 

With default settings DSF/DFF files are streamed as track.dsd, changing it to track.dsf forces Jriver, USB Audio Player Pro to output as DSD. 
But as I cannot read the internals of the player and as the metadata sent by the server is the same I'm not 100% sure if it's not just packing the PCM container and outputing as DSD, which Jriver is capable of with it's DSP studio. 

Here you can see an example with Bubble UPnP
<a href="https://ibb.co/mNMDwBp"><img src="https://i.ibb.co/8YSKpBk/dsd2.jpg" alt="dsd2" border="0"></a>

And as I mentioned, this isn't the case with UMS
<a href="https://ibb.co/3SQ1r2B"><img src="https://i.ibb.co/gthyjNz/dsd3.jpg" alt="dsd3" border="0"></a>

if this can be solved by the end user, please, let me know as Gerbera seems to be the most reliable DLNA media server so far.

neheb:
I don't think Gerbera does any transcoding by default. The metadata is most definitely handled by ffmpeg as there doesn't seem to be any other available: https://github.com/gerbera/gerbera/blob/master/src/metadata/metadata_handler.cc#L115 . Note the audio part.

I don't see anything in src/metadata/ffmpeg_handler that implies transcoding.

fandangos:
it's not transcoding, the file goes intact.

I'm not familiar with DLNA but I believe that along with mime type and other DLNA specific information there's some kind of metadata involved that is send with the file. 
That's why it's the same file but the player will play it as PCM because of this metadata sent along with the file. 

[...]

PR #2407, opened 2022-02-06
Multiply out sample rate reported by FFMpeg

whyman:
When there are multiple bits per sample we would have reported an
incorrect sample rate. For PCM bits per raw/coded are both 0.

Bug: https://github.com/gerbera/gerbera/issues/2401

### DIFF
--- a/src/metadata/ffmpeg_handler.cc
+++ b/src/metadata/ffmpeg_handler.cc
@@ -167,7 +167,7 @@ void FfmpegHandler::addFfmpegMetadataFields(const std::shared_ptr<CdsItem>& item
 // ffmpeg library calls
 void FfmpegHandler::addFfmpegResourceFields(const std::shared_ptr<CdsItem>& item, const AVFormatContext* pFormatCtx) const
 {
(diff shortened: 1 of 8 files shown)


neheb:
tests need updating.

src/metadata/ffmpeg_handler.cc
```diff
@@ -238,7 +238,7 @@ void FfmpegHandler::addFfmpegResourceFields(const std::shared_ptr<CdsItem>& item
                 int bitsPerSample = as_codecpar(st)->bits_per_coded_sample;
                 log_debug("Bits per coded sample: {}", bitsPerSample);
 
-                sampleFreq = sampleFreq * bitsPerSample > 0 ? bitsPerSample : 1;
+                sampleFreq = sampleFreq * (bitsPerSample > 0 ? bitsPerSample : 1);
```
neheb, line 241: you can avoid the () if you use *= instead of =.

whyman:
_Finally_ had the time to fix the tests, sorry for the wait.

PR #2409, opened 2022-02-06
Fix libary links in compilation docs

cweiske:
https://docs.gerbera.io/en/stable/compile.html#dependencies


cweiske:
Fixed links in https://gerbera--2409.org.readthedocs.build/en/2409/compile.html#dependencies

PR #2416, opened 2022-02-08
transform conversions

neheb:
Found with cppcheck


whyman:
The transforms are more characters and less clear than simple for loops IMO

PR #2417, opened 2022-02-08
cppcheck fixes

neheb:


neheb:
it should be safe. It's not a requirement to pass those quote usages with namespace::quote.

PR #2422, opened 2022-02-11
don't bother finding before inserting.

neheb:
Found with cppcheck.


KarlStraussberger:
This was added because of a crash without: If the map does not have an entry, the push_back is called on a null object

neheb:
Hmm OK.

PR #2427, opened 2022-02-12
replace copy with move

neheb:
More efficient.

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -1603,7 +1603,7 @@ std::vector<std::shared_ptr<AutoscanDirectory>> ContentManager::getAutoscanDirec
     auto all = autoscan_timed->getArrayCopy();
 
     auto ino = autoscan_inotify->getArrayCopy();
-    std::copy(ino.begin(), ino.end(), std::back_inserter(all));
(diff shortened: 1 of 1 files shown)


whyman:
Where can I read about the 3 argument version of std::move?

neheb:
https://en.cppreference.com/w/cpp/algorithm/move

It's actually the same as std::copy. It just calls std::move on the argument.

src/content/content_manager.cc
```diff
@@ -351,7 +351,10 @@ std::deque<std::shared_ptr<GenericTask>> ContentManager::getTasklist()
         return taskList;
 
     taskList.push_back(std::move(t));
-    std::copy_if(taskQueue1.begin(), taskQueue1.end(), std::back_inserter(taskList), [](auto&& task) { return task->isValid(); });
+
+    for (auto&& task : taskQueue1)
+        if (task->isValid())
+            taskList.push_back(std::move(task));
```
KarlStraussberger, line 357: This moves the contents of taskQueue1 to taskList, which is not used in case taskQueue2 still contains valid entries. In this case the former entries of taskQueue1 are lost.

neheb:
Removed offending sections. Please review.

PR #2429, opened 2022-02-13
FFMpegThumbnailer: Split into its own handler

whyman:
For some reason this code was in the ffmpeg handler, when really there
is no dependency there, so split it out to clean things up.


KarlStraussberger:
Great idea. I still want to use it for images without thumb. Made it a lot easier.

whyman:
@KarlStraussberger looks like the web test is also failing in master for some reason

KarlStraussberger:
> @KarlStraussberger looks like the web test is also failing in master for some reason

Had that issue after bumping some javascript lib. Needs an extra kick to get over the timing in headless.

src/metadata/ffmpeg_thumbnailer_handler.cc
```diff
@@ -0,0 +1,142 @@
...
+        y = config->getIntOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_THUMBSIZE) * y / x;
+        x = config->getIntOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_THUMBSIZE);
+        std::string resolution = fmt::format("{}x{}", x, y);
+        ffres->addAttribute(R_RESOLUTION, resolution);
+        item->addResource(ffres);
+        log_debug("Adding resource for video thumbnail");
+    }
+}
+
+FfmpegThumbnailerHandler::FfmpegThumbnailerHandler(const std::shared_ptr<Context>& context)
```
neheb, line 137: This looks suspicious. I think you can replace this with a using declaration.

neheb:
Well needed cleanup.

src/metadata/ffmpeg_thumbnailer_handler.cc
neheb:
```
--- a/src/metadata/ffmpeg_thumbnailer_handler.cc
+++ b/src/metadata/ffmpeg_thumbnailer_handler.cc
@@ -134,9 +134,4 @@ void FfmpegThumbnailerHandler::fillMetadata(const std::shared_ptr<CdsObject>& ob
     }
 }

-FfmpegThumbnailerHandler::FfmpegThumbnailerHandler(const std::shared_ptr<Context>& context)
-    : MetadataHandler(context)
-{
-}
-
 #endif
diff --git a/src/metadata/ffmpeg_thumbnailer_handler.h b/src/metadata/ffmpeg_thumbnailer_handler.h
index 6a498036..520950e2 100644
--- a/src/metadata/ffmpeg_thumbnailer_handler.h
+++ b/src/metadata/ffmpeg_thumbnailer_handler.h
@@ -32,8 +32,8 @@ Gerbera - https://gerbera.io/
 #include "metadata_handler.h"

 class FfmpegThumbnailerHandler : public MetadataHandler {
+    using MetadataHandler::MetadataHandler;
 public:
-    explicit FfmpegThumbnailerHandler(const std::shared_ptr<Context>& context);
     void fillMetadata(const std::shared_ptr<CdsObject>& obj) override;
     std::unique_ptr<IOHandler> serveContent(const std::shared_ptr<CdsObject>& obj, int resNum) override;


```

src/metadata/ffmpeg_thumbnailer_handler.cc
```diff
@@ -0,0 +1,142 @@
...
+#include <libffmpegthumbnailer/filmstripfilter.h>
+#include <libffmpegthumbnailer/videothumbnailer.h>
+
+fs::path FfmpegThumbnailerHandler::getThumbnailCacheBasePath(const Config& config_)
+{
+    auto configuredDir = config_.getOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_CACHE_DIR);
+    if (!configuredDir.empty())
+        return configuredDir;
+
+    auto home = config_.getOption(CFG_SERVER_HOME);
```
neheb, line 36: This looks like it can be reduced to one return statement.

src/metadata/ffmpeg_thumbnailer_handler.cc
```diff
@@ -0,0 +1,142 @@
...
+    return fs::path(home) / "cache-dir";
+}
+
+fs::path FfmpegThumbnailerHandler::getThumbnailCachePath(const fs::path& base, const fs::path& movie)
+{
+    assert(movie.is_absolute());
+
+    auto path = base / movie.relative_path();
```
neheb, line 44: this looks like it can be a single return statement.

src/metadata/ffmpeg_thumbnailer_handler.cc
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    auto thumbLock = std::scoped_lock<std::mutex>(thumb_mutex);
+
+    auto th = ffmpegthumbnailer::VideoThumbnailer(config->getIntOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_THUMBSIZE), false, true, config->getIntOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_IMAGE_QUALITY), false);
+    std::vector<uint8_t> img;
+
+    th.setSeekPercentage(config->getIntOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_SEEK_PERCENTAGE));
+    if (config->getBoolOption(CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_FILMSTRIP_OVERLAY))
+        th.addFilter(new ffmpegthumbnailer::FilmStripFilter());
```
neheb, line 91: I can't tell if this causes a memory leak. Where's the corresponding delete?

whyman: This is the existing code, if you want to double check that it would be great.

whyman:
@neheb happy to accept patches for your suggestions. I just don't have time to make them myself currently ☹️

neheb:
Sure.

Issue #358, opened 2018-10-02
How does deletion from the database work?


stheine:
I'm currently looking at Gerbera 1.2.0.

In the Database view, what does 'Delete Item' or 'Delete All' actually do? I guess the source files are not getting deleted from the filesystem, so the system must just flag the files somehow internally. The documentation states "Deleting the item removes it from the virtual Database View". So what is actually flagged as removed?
a) the source file location from the filesystem?
b) or the artist & title combination?

in case of a), if I would rename the file in the filesystem, would it come back into the database?
in case of b), would it delete all occurrences from various albums?

is there a way to query that 'virtual Database View', or the files that have been deleted there?

can I un-delete a file that was accidentally flagged? I guess that a 'Delete All' on a higher level folder could remove a large bulk of data.

elmodaddyb:
@stheine - I will explain my understanding from working with the code.  _Albeit, some assumptions could be incorrect._

### Adding a Gerbera virtual object

Let's take a scenario of adding `/home/gerbera/myfile.mp3`

1. Gerbera adds a virtual object that represents the file in the 

    `Database View --> PC Directory -->  home --> gerbera --> myfile.mp3`

    > There is now a virtual `objectID` for this media file.

2. Gerbera runs the `import.js` creating any number of virtual containers which create virtual objects that point back to `objectID` via a `refID`
   * This creates any number of artist, title combinations(_refs_) for that `objectID`

### When deleting a virtual object reference in the virtual database view (**not PC Directory**)

1. **Delete** just deletes that item by its `refID`
2. **Delete All** deletes recursively all children that belong to the `refID`

### When deleting a reference in the `Database View --> PC Directory`

1. Removes **ALL** references to the objectID, including any referenced virtual objects
2. This file is no longer part of the Gerbera media library
3. The file still exists on the file system

### Deleting

When a delete is completed, it is removed from the data record....not flagged (_from what i can tell_).  This may result in:

1. Files be re-added via auto-scan at a later time
2. Gerbera does not delete the filesystem file (_the real file_)
3. You can add back deleted files using the filesystem view

> Please feel free to add any details I missed, or if I misspoke on the capability.

Thanks,
Eamonn

stheine:
> Files be re-added via auto-scan at a later time

I can confirm that deleting files from the database, that are contained in an autoscan folder, will only work temporary. latest, on the next server restart, they're coming back into the database.
I think this is fine from the autoscan perspective.

[...]

PR #2430, opened 2022-02-13
Document the special role of "PC Directory"

cweiske:
Information taken from https://github.com/gerbera/gerbera/issues/358


whyman:
Nice!

PR #2431, opened 2022-02-15
remove string_view quote

neheb:
It's not clear that it's a good idea.

### DIFF
--- a/src/database/sql_database.h
+++ b/src/database/sql_database.h
@@ -95,10 +95,9 @@ class SQLDatabase : public Database {
 public:
     /* methods to override in subclasses */
     virtual std::string quote(const std::string& str) const = 0;
-    /* wrapper functions for different types */
-    std::string quote(std::string_view str) const { return quote(std::string(str)); }
     // required to handle #defines
     std::string quote(const char* str) const { return quote(std::string(str)); }
+    /* wrapper functions for different types */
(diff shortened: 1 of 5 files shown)


whyman:
Why is it not a good idea?

KarlStraussberger:
It used to be const char*. Actually, I prefer the `string_view` because of the comparison operators.

neheb:
sure. but there's no point here. All usages use the data() method.

src/database/sql_database.h
```diff
@@ -96,10 +96,9 @@ class SQLDatabase : public Database {
     /* methods to override in subclasses */
     virtual std::string quote(const std::string& str) const = 0;
     /* wrapper functions for different types */
-    std::string quote(std::string_view str) const { return quote(std::string(str)); }
-    // required to handle #defines
     std::string quote(const char* str) const { return quote(std::string(str)); }
     std::string quote(char val) const { return quote(fmt::to_string(val)); }
+    // required to handle #defines
```
KarlStraussberger, line 101: The comment belongs to the `const char*` version

neheb: Interesting... Wonder if it makes sense to replace those defines with string_view instead of this PR.

neheb: hmm if I remove that const char version, it still compiles.

PR #2437, opened 2022-02-20
don't use std::map 's merge

neheb:
libcxx 7 does not support it.

I'm unsure about this commit. Debian 10, which comes with libcxx 7 by default, does not compile without it. OTOH buster-backports has libcxx 8 and 11. So maybe moot.


whyman:
It's part of the c++17 spec, so I think we just don't support that old version of libcxx

PR #2442, opened 2022-02-20
UI: Hide more when not logged in

whyman:
Currently it looks like you can do things, when you cant


whyman:
Now looks like 
image

PR #2444, opened 2022-02-20
Thumbnails for images and grid view for items

KarlStraussberger:
It is possible to switch from the known table/list view to grid view
grafik

or large grid view
grafik


KarlStraussberger:
The config option to turn off ffmpegthumbnailer for images is still missing

src/metadata/metadata_handler.cc
```diff
@@ -121,16 +121,17 @@ void MetadataHandler::extractMetaData(const std::shared_ptr<Context>& context, c
...
     }
 #endif // HAVE_FFMPEG
 
 #ifdef HAVE_FFMPEGTHUMBNAILER
     // Thumbnails for videos
     if (itemCls == UPNP_CLASS_VIDEO_ITEM)
-        FfmpegThumbnailerHandler(context).fillMetadata(item);
+        FfmpegThumbnailerHandler(context, CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_VIDEO_ENABLED).fillMetadata(item);
+    else if (itemCls == UPNP_CLASS_IMAGE_ITEM)
+        FfmpegThumbnailerHandler(context, CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_IMAGE_ENABLED).fillMetadata(item);
```
whyman, line 134: I think we should keep the "enabled/disabled" logic out of the handlers themselves, we should just not call them if they are not enabled, and because bool type arguments suck.

Something like `if (itemCls == UPNP_CLASS_IMAGE_ITEM && CFG_SERVER_EXTOPTS_FFMPEGTHUMBNAILER_IMAGE_ENABLED)`

whyman:
Looks nice!

src/metadata/metadata_handler.cc
whyman: Ah, nevermind, I see you pass the Option to use as a flag.

KarlStraussberger: Thaught about this, but `serveContent` is called from too many places to squeeze it in here

KarlStraussberger:
> Looks nice!

My wife loves it too. A lot easier to find the proper holiday picture way back to Australia-2005 😄

Issue #357, opened 2018-10-02
Want indicator for 'added item'


stheine:
I'm currently looking at Gerbera 1.2.0.

When I'm in the FileSystem view, I would like to see any indication on what directories I have flagged by 'Add Item' or 'Add Autoscan Item'. Currently there is nothing.
What would happen if I would add a directory using 'Add Item' multiple times?
Is 'recursive' always set in 'Add Item'? In 'Add Autoscan Item' I'm asked if I want recursive or not.

elmodaddyb:
This is a great feature idea.  Visibility of the added items would be helpful from the filesystem view.

#### To answer your other questions...

> What would happen if I would add a directory using 'Add Item' multiple times?

Gerbera finds the filesystem object in the virtual database and will not re-add multiple items.

> Is 'recursive' always set in 'Add Item'?

Yes --> [add.cc](https://github.com/gerbera/gerbera/blob/master/src/web/add.cc#L62)

> In 'Add Autoscan Item' I'm asked if I want recursive or not.

Yes, you can choose to be recursive here through the UI [web_autoscan.cc](https://github.com/gerbera/gerbera/blob/master/src/web/web_autoscan.cc#L116)

stheine:
I just found this 'autoscan' indicator in the 'database/PC Directory' tree:
![image](https://user-images.githubusercontent.com/19614566/46579619-a7044900-ca15-11e8-8519-6602bf2c7338.png)

This clearly goes in the right direction. I just would love to see this in the filesystem tree as well.

cweiske:
The file list also should show an indicator if the file is already known by Gerbera, not only the tree.

KarlStraussberger:
Files are normally imported by autoscan, so if a file is missing it was deleted manually. On the other hand it is quite expensive with respect to calculation to check each file.

KarlStraussberger:
I fully unserstand your issue but it means adding a feature that impacts runtime for rare situations which can also be solved by reimporting part of the files

PR #2446, opened 2022-02-21
Also display autoscan badge in filesystem view

KarlStraussberger:
See #357

### DIFF
--- a/src/web/directories.cc
+++ b/src/web/directories.cc
@@ -31,6 +31,7 @@
 
 #include "pages.h" // API
 
+#include "content/content_manager.h"
 #include "database/database.h"
 #include "util/string_converter.h"
 #include "util/tools.h"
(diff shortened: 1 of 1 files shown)


src/web/directories.cc
```diff
@@ -87,16 +89,21 @@ void Web::Directories::process()
...
         /// \todo replace hexEncode with base64_encode?
         std::string id = hexEncode(filepath.c_str(), filepath.string().length());
-        filesMap.emplace(id, std::pair(filepath.filename(), hasContent));
+        filesMap.emplace(id, std::pair(filepath, hasContent));
     }
 
     auto f2i = StringConverter::f2i(config);
     for (auto&& [key, val] : filesMap) {
-        auto&& [file, has] = val;
+        auto file = val.first;
```
neheb, line 97: :) Curious about this. I was under the impression structured bindings we're passed by value.

KarlStraussberger: I tried several options in vain.

PR #2450, opened 2022-02-23
get rid of more startswith

neheb:
operator>= can be used.

ping @KarlStraussberger


KarlStraussberger:
It does lexicographical comparison /a < /b. Been trying this.

PR #2451, opened 2022-02-23
replace for loop with while

neheb:
More readable.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -2473,12 +2473,13 @@ void SQLDatabase::generateResourceDBOperations(const std::shared_ptr<CdsObject>&
             resId++;
         }
         // res_id in db resources but not obj resources, so needs a delete
-        for (; resId < dbResources.size(); resId++) {
+        while (resId < dbResources.size()) {
             if (dbResources.at(resId)) {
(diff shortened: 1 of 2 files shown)


src/database/sql_database.cc
```diff
@@ -2473,11 +2473,12 @@ void SQLDatabase::generateResourceDBOperations(const std::shared_ptr<CdsObject>&
             resId++;
         }
         // res_id in db resources but not obj resources, so needs a delete
-        for (; resId < dbResources.size(); resId++) {
+        while (resId < dbResources.size()) {
             if (dbResources.at(resId)) {
                 std::map<std::string, std::string> resourceSql;
                 resourceSql["res_id"] = quote(resId);
                 operations.emplace_back(RESOURCE_TABLE, std::move(resourceSql), Operation::Delete);
+                ++resId;
```
KarlStraussberger, line 2481: You need to increase outside the if scope.

KarlStraussberger (changes requested):

src/database/sql_database.cc
neheb: yikes good catch.

neheb:
Fixed.

PR #2452, opened 2022-02-23
treewide: replace suffix ++ with prefix ++

neheb:
The latter is recommended as the former can create copies.

Intentionally left out usages where the return value is used.


whyman:
This is going to cause bugs, I don't think it's worth it

PR #2462, opened 2022-02-26
simplify some bools

neheb:


neheb:
Wonder if that port thing is even right. Documentation says 49152 and up.

PR #2470, opened 2022-03-06
doc/install.rst: drop Entware

ffontaine:
gerbera has been dropped from Entware since https://github.com/Entware/rtndev/commit/6d6803f1b5fca0f03623c14a4bd924cbe8000166


KarlStraussberger:
Thanks again. We should try to get back on routers

PR #2471, opened 2022-03-06
Add Clients to database

KarlStraussberger:


src/util/upnp_clients.h
```diff
@@ -115,6 +116,8 @@ class Clients {
     std::vector<ClientCacheEntry> cache;
 
     std::vector<ClientInfo> clientInfo;
+    std::shared_ptr<Database> database;
+    int cacheThreshold;
```
neheb, line 120: maybe use std::chrono::hours here?

KarlStraussberger: Good point

whyman:
Nice!

Issue #2176, opened 2021-11-14
[Question] Scanning/indexing FLAC


aanno:
I've got a big music collection (around 12000 tracks from about 1100 interprets) on HDD, mostly ripped from CDs with `asunder` to FLAC (with maximal compression). Using gerbera 1.9.2, I recognize that nearly all my collection is _not_ indexed. At first glance, the index only includes some MP3 and OGG files.

I'm not sure if I stepped upon https://github.com/gerbera/gerbera/issues/401 - is there a simple way to produce helpful debug logs for the issue?

KarlStraussberger:
Just run gerbera with `--debug` and there's plenty of log.

aanno:
Well, I encounter very many 

```
select Error: [/__w/gerbera/gerbera/src/database/sqlite3/sqlite_database.cc:333] addTask Error: sqlite3 task queue is already closed","code":800}}
```

entries in the logs. In addition WebUI on http://192.168.10.28:49152 is able to display the filesystem part with my track, but refuse to give me 'Config' or 'Database'.

aanno:
I can see things like

```
2021-11-14 18:27:25.429  debug: [web_request_handler.cc:211] open(): output-----------------------{"success":true,"files":{"parent_id":"2f68
6f6d652f7470617363682f4d7573696b2f644555532f466f6c6c6f77696e6720536561","location":"/home/tpasch/Musik/dEUS/Following Sea","file":[{"id":"2f
686f6d652f7470617363682f4d7573696b2f644555532f466f6c6c6f77696e67205365612f303120517561747265204d61696e732e666c6163","filename":"01 Quatre Ma
ins.flac"},{"id":"2f686f6d652f7470617363682f4d7573696b2f644555532f466f6c6c6f77696e67205365612f303220536972656e732e666c6163","filename":"02 S
irens.flac"},{"id":"2f686f6d652f7470617363682f4d7573696b2f644555532f466f6c6c6f77696e67205365612f30332048696464656e20576f756e64732e666c6163",
"filename":"03 Hidden Wounds.flac"},{"id":"2f686f6d652f7470617363682f4d7573696b2f644555532f466f6c6c6f77696e67205365612f3034204769726c73204b6
56570204472696e6b696e672e666c6163","filename":"04 Girls Keep Drinking.flac"},{"id":"2f686f6d652f7470617363682f4d7573696b2f644555532f466f6c6c
[...]

PR #2477, opened 2022-03-08
Fix tag matches with ffmpeg_handler

whyman:
Some of the keys were uppercase e.g. ARTIST when returned from the FLAC
parser, so lowercase the keys first.

Fixes: https://github.com/gerbera/gerbera/issues/2176


jeeger:
Awesome, thanks!

PR #2486, opened 2022-03-14
Ensure search sort by title

KarlStraussberger:
Also fix broken mysql upgrade statement


whyman:
@KarlStraussberger this may be my favourite patch in some time.

So small, yet so useful!

PR #2495, opened 2022-03-17
Support searching playlists containers

### LINKED ISSUE #2492: Playlist items are not returned from uPNP searches
**_Summary:_**  Playlist items are not returned from searches initiated on my uPNP client. The playlist is visible when browsing the Gerbera web UI under Playlists->All Playlists->Playlist Name and Playlists->Directories->Playlists->Playlist Name

**_Action:_** In a uPNP client, search using the criteria `upnp:class = "object.container.playlistContainer" and dc:title = "Playlist Name"`,

**_Expected Results:_**  Playlist matching search criteria is returned
**_Actual Results:_**  No results are returned

**_Investigation:_**  
From the Gerbera debug log, the search being executed is:

`SELECT DISTINCT "c"."id", "c"."ref_id", "c"."parent_id", "c"."object_type", "c"."upnp_class", "c"."dc_title", "c"."mime_type"
        , "c"."flags", "c"."part_number", "c"."track_number", "c"."location", "c"."last_modified", "c"."last_updated"
FROM "mt_cds_object" "c"
INNER JOIN "mt_metadata" "m" ON "c"."id" = "m"."item_id"
INNER JOIN "grb_cds_resource" "re" ON "c"."id" = "re"."item_id"
WHERE (LOWER("c"."upnp_class")=LOWER('object.container.playlistContainer')) AND (LOWER("c"."dc_title")=LOWER('Playlist Name'))
ORDER BY "c"."dc_title" ASC;`

The playlists do not have a row in the grb_cds_resource table, so the "INNER JOIN" means the query returns no records.

I am able to work around this issue by manually inserting playlist items into the grb_cds_resource table

`INSERT INTO grb_cds_resource (item_id, res_id, handlerType) VALUES (1235555,0,0);`

It seems like either playlist items need to be added to the grb_cds_resource table or the query should use a left outer join with grb_cds_resource.

KarlStraussberger:
fixes #2492


src/web/edit_load.cc
```diff
@@ -47,6 +47,38 @@ Web::EditLoad::EditLoad(const std::shared_ptr<ContentManager>& content, std::sha
 {
 }
 
+constexpr auto upnpFlags = std::array {
```
neheb, line 50: static or anonymous namespace.

src/web/edit_load.cc
```diff
@@ -47,6 +47,38 @@ Web::EditLoad::EditLoad(const std::shared_ptr<ContentManager>& content, std::sha
...
+};
+
+static std::string mapFlags(int flags)
+{
+    if (!flags)
+        return "None";
+
+    std::vector<std::string> myFlags;
+
+    for (auto uFlag : upnpFlags) {
```
neheb, line 68: looks like this should use structured bindings.

PR #2503, opened 2022-03-19
string to string_view conversion

neheb:
clang-tidy suggests const std::string&. Probably better to pass a view.


whyman:
I don't think we should pass views to those c functions, strings are always null terminated

PR #2510, opened 2022-03-20
Refactor handling of network addresses

KarlStraussberger:
Improves readability and avoids pointer handling.


neheb:
erm sounds like C pointers get changed to shared_ptrs.

I'd need to look at this more closely, but sounds like you can use references with deleted copy operations.

PR #2512, opened 2022-03-23
ResourceAttribute new style enum

whyman:


src/content/scripting/script_names.h
```diff
@@ -31,23 +31,23 @@
...
-    std::pair(R_PROTOCOLINFO, "R_PROTOCOLINFO"),
-    std::pair(R_RESOURCE_FILE, "R_RESOURCE_FILE"),
-    std::pair(R_BITS_PER_SAMPLE, "R_BITS_PER_SAMPLE"),
-    std::pair(R_TYPE, "R_TYPE"),
-    std::pair(R_FANART_OBJ_ID, "R_FANART_OBJ_ID"),
-    std::pair(R_FANART_RES_ID, "R_FANART_RES_ID"),
-    std::pair(R_LANGUAGE, "R_LANGUAGE"),
-    std::pair(R_AUDIOCODEC, "R_AUDIOCODEC"),
-    std::pair(R_VIDEOCODEC, "R_VIDEOCODEC"),
+static auto res_names = std::map<ResourceAttribute, std::string_view> {
```
neheb, line 34: hmm?

whyman: Maps are great and fit what we want to achieve here, e.g. lookup by value. We dont need the constexpr stuff because we dont do any compile time computation with these maps, they are only queried at runtime, so static does the job here.

KarlStraussberger:
What about moving the declaration to cds_resource? All calls need resources but not metadata.

KarlStraussberger:
Do you mind if I merge the transcoding PR? It causes some conflicts.

src/upnp_xml.cc
```diff
@@ -841,10 +845,12 @@ void UpnpXMLBuilder::addResources(const std::shared_ptr<CdsItem>& item, pugi::xm
             continue;
         }
 
+        // TODO FIXME Should these be attributes or something else?
```
KarlStraussberger, line 848: The goal in this method is to calculate temporary properties depending on the client and return them. We should not make them permanent, e.g. the raw data for _protocolInfo_ is in attributes, extended by more details and returned to the client.

We may add another attribute set `std::map<std::string, std::string> responseAttributes` to the resource for this purpose allowing more sophisticated handling and storage along with the request.

I recommend doing this separately.

src/upnp_xml.cc
```diff
@@ -646,38 +646,42 @@ std::pair<bool, int> UpnpXMLBuilder::insertTempTranscodingResource(const std::sh
...
                     targetMimeType.append(fmt::format(";channels={}", chan));
                 }
             }
 
-            tRes->addAttribute(R_PROTOCOLINFO, renderProtocolInfo(targetMimeType));
+            tRes->addAttribute(CdsResource::Attribute::PROTOCOLINFO, renderProtocolInfo(targetMimeType));
 
-            tRes->mergeAttributes(tp->getAttributes());
+            if (tp->isThumbnail())
+                tRes->addOption(RESOURCE_CONTENT_TYPE, EXIF_THUMBNAIL);
```
KarlStraussberger, line 682: This is already in line 644

whyman: The joys of rebasing

src/upnp_xml.cc
```diff
@@ -845,10 +846,12 @@ void UpnpXMLBuilder::addResources(const std::shared_ptr<CdsItem>& item, pugi::xm
...
         }
 
+        // TODO FIXME Should these be attributes or something else?
+        std::map<std::string, std::string> subtitleQuirks;
         if (handlerType == CH_DEFAULT && !captionInfoEx.empty() && quirks && quirks->checkFlags(QUIRK_FLAG_PV_SUBTITLES)) {
             auto captionInfo = captionInfoEx[0];
-            resAttrs["pv:subtitleFileType"] = toUpper(captionInfo["sec:type"]);
-            resAttrs["pv:subtitleFileUri"] = captionInfo[""];
+            subtitleQuirks["pv:subtitleFileType"] = toUpper(captionInfo["sec:type"]);
+            subtitleQuirks["pv:subtitleFileUri"] = captionInfo[""];
```
KarlStraussberger, line 854: You need to pass the quirks map to `renderResource` for output as xml attributes. Currently this is dead code.

Other option is to return `xml_node` from `renderResource`

KarlStraussberger (commented):
pv:subtitle is still open

src/upnp_xml.cc
whyman: `clientSpecficAttrs` added in latest rev

src/upnp_xml.cc
whyman: Yep, too hasty on the push

KarlStraussberger:
Ready?

whyman:
Should be!

PR #2521, opened 2022-03-28
WavPack: Get metadata with original library

KarlStraussberger:


src/metadata/wavpack_handler.cc
```diff
@@ -0,0 +1,169 @@
...
+        was created. For all files created prior to WavPack 5 this will be zero (WP_FORMAT_WAV). Introduced
+        in WavPack 5.0.0. The following formats are currently defined in wavpack.h:
+        - WP_FORMAT_WAV: Microsoft Waveform Audio Format, including BWF and RF64
+        - WP_FORMAT_W64: Sony Wave64
+        - WP_FORMAT_CAF: Apple Core Audio
+        - WP_FORMAT_DFF: Philips DSDIFF
+        - WP_FORMAT_DSF: Sony DSD format
+    */
+    auto fileFormat = WavpackGetFileFormat(context);
+    resource->addAttribute(R_VIDEOCODEC, fileFormats[fileFormat].data());
```
whyman, line 79: AUDIO?

README.md
```diff
@@ -14,7 +14,7 @@ View our documentation online at [https://docs.gerbera.io](https://docs.gerbera.
 ## Features
 * Browse and playback your media via your network on all kinds of devices.
 * Web UI with a tree view of the database and the file system, allowing to add/remove/edit/browse your media
-* Metadata extraction from MP3, OGG, AAC, M4A, FLAC, JPG (and many more!) files.
+* Metadata extraction from MP3, OGG, AAC, M4A, FLAC, wavpack, JPG (and many more!) files.
```
whyman, line 17: 🎉

src/metadata/wavpack_handler.cc
KarlStraussberger: That's the current hack to get it into the filtering for dlna profiles (that needs an update as well). I didn't want to create a new attribute while your still refactoring them.

whyman:
LGTM

src/metadata/wavpack_handler.cc
```diff
@@ -0,0 +1,285 @@
...
+/// \file wavpack_handler.cc
+
+#ifdef HAVE_WAVPACK
+#include "wavpack_handler.h" // API
+
+#include "cds_objects.h"
+#include "iohandler/mem_io_handler.h"
+#include "util/mime.h"
+
+static auto fileFormats = std::map<int, std::string_view> {
```
neheb, line 33: why not const?

neheb:
Doesn't taglib support wavpack?

src/metadata/wavpack_handler.cc
```diff
@@ -0,0 +1,285 @@
...
+    char value[MAX_WV_IMAGE_SIZE];
+    auto size = WavpackGetBinaryTagItem(context, artTag.c_str(), value, MAX_WV_IMAGE_SIZE);
+    if (size > MAX_WV_IMAGE_SIZE) {
+        log_warning("file {}: wavpack binary tag {} value '{}' truncated (full size {})", item->getLocation().c_str(), artTag, value, size);
+        size = MAX_WV_IMAGE_SIZE;
+    }
+    if (startswith(artTag, "Cover Art")) {
+        log_debug("Found image '{}': {}", artTag, value);
+        return std::make_unique<MemIOHandler>(value + strlen(value) + 1, size - strlen(value) - 1);
+    } else {
```
neheb, line 278: I don't think else after return is needed.

KarlStraussberger:
> Doesn't taglib support wavpack?

It does not handle the file format und album art.

neheb:
how strange... What does it handle then?

KarlStraussberger:
Only regular tags but not technical details of wv files.

src/cds_resource.h
```diff
@@ -166,6 +168,7 @@ class CdsResource {
         { CdsResource::Attribute::LANGUAGE, "dc:language" },
         { CdsResource::Attribute::AUDIOCODEC, "sec:acodec" },
         { CdsResource::Attribute::VIDEOCODEC, "sec:vcodec" },
+        { CdsResource::Attribute::FILEFORMAT, "fileFormat" },
```
whyman, line 171: fileFormat here would be DSD if I am reading it right, but the file on disk is actually a wavpack?

IMO We need to _store_ 3 bits of info:
`ContainerType`, `VideoCodec` and `AudioCodec`

Here I guess ContainerType is WavPack, and AudioCodec is DSD?

whyman:
> > Doesn't taglib support wavpack?
> 
> It does not handle the file format und album art.

Taglib supports the artwork part - we just havent coded it our side to add support (and we should).

But it doesnt seem to handle the different types of codec that can be embedded though that this change hopes to address.

src/cds_resource.h
KarlStraussberger: It's more complicated here.

Wavpack is some sort of ZIP file that contains compressed audio (no video). The audio file can be WAV, DFF or DSD, etc. The audio like DSD may use a codec. So we cannot store DSD it as acodec - vcodec was just the hack to get dlnaprofile selection tested. `fileFormat` describes the media type of packed file.

whyman: I think we are agreeing 🤔 

We need to know the format inside, which is what the acodec field is for in my understanding.

In my head it looks like
Typical MKV video `mkv(h264,aac)`: acodec=aac, vcodec=h264 and fileformat=mkv
OGV file `ogg(h264,mp3)`: acodec=mp3, vcodec=h264 and fileformat=ogg
OGA file `ogg(null,vorbis)`: acodec=vorbis, vcodec=null and fileformat=ogg
M4A file `mp4(null,aac)`: acodec=aac, vcodec=null and fileformat=mp4

So with wavpack: `wv(null,dsd)`: acodec=dsd, vcodec=null and fileformat=wv

We need all 3 bits of information at least for transcoding effectively.

Say we have a client that supports DSD audio streams but only in its "native" DSF container, we could create a profile that simply copies that stream out of the wavpack with ffmpeg into the new container for that client.

KarlStraussberger: Almost audio codec is e.g. lsbf_planar (the codec of the original file) and fileformat is DSD. fileFormat can be content type for other media. So it may happen that we need to transcode because the client does not support lsbf_planar inside a DSD inside a wavpack...

KarlStraussberger: As we are already below Resource we can just use `format`

PR #2525, opened 2022-03-29
Restore friendly name in UI

### LINKED ISSUE #2522: V1.10.0 Web UI Doesn't show the Gerbera server name any more
Just testing V1.10.0 and the first thing I have noticed was a much faster and cleaner web ui but the name of the server is nowhere to be seen.  It used to be in the title bar after the icon and the word gerbera in the 'Home [server name] field before 'database'.

I also noticed that clicking on the word Gerbera takes you to the Web UI home instead of the GIT like it used to.  Don't know if this was always part of the plan for cleaning up the UI.

I know it's unusiual to have more than one server but I have a live and testing server andit's really easy to get confused between the two sometimes.  Is it possible to get this back on the next rev please.

whyman:
Fixes #2522

Thanks to @BobManMasher for the report

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -18,7 +18,10 @@
               aria-expanded="false" aria-label="Toggle navigation">
         <span class="navbar-toggler-icon"></span>
       </button>
-      <a id="nav-home" class="navbar-brand" data-gerbera-menu-cmd="HOME" title="Gerbera"><img src="assets/theme/gerbera-logo-white.png" class="d-inline-block align-top" alt="Gerbera Logo"></a>
+      <a id="nav-home" class="navbar-brand" data-gerbera-menu-cmd="HOME" title="Gerbera" href="#">
+        <img src="assets/theme/gerbera-logo-white.png" class="d-inline-block align-top" alt="Gerbera Logo">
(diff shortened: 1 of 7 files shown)


web/index.html
```diff
@@ -18,7 +18,10 @@
               aria-expanded="false" aria-label="Toggle navigation">
         <span class="navbar-toggler-icon"></span>
       </button>
-      <a id="nav-home" class="navbar-brand" data-gerbera-menu-cmd="HOME" title="Gerbera"><img src="assets/theme/gerbera-logo-white.png" class="d-inline-block align-top" alt="Gerbera Logo"></a>
+      <a id="nav-home" class="navbar-brand" data-gerbera-menu-cmd="HOME" title="Gerbera" href="#">
```
KarlStraussberger, line 21: We do not need the href because it's handled by jquery.

whyman: An `a` element without the href will not get the pointer styling on hover, plus it does no harm.

KarlStraussberger:
@whyman Seems you need to take a look at the tests
```
1) Config Suite
[1]        The gerbera config
[1]          a gerbera config item is changed:
```

whyman:
Urgh, I dont really see how I could have broken that :-1:

PR #2528, opened 2022-03-31
Find DNLA profiles by resource attributes

KarlStraussberger:
Extracted ConfigSetup classes to subdir
Add new ConfigSetupVector to store any type of key-value pairs


whyman:
I wonder if this config based mapping will be effective.

We need to work out the types in code IMO because there are so many rules:

`PNG_SM` = Smaller than 48px, no interlacing
`PNG_SM_ICO` = Smaller than 48px, no interlacing and not to to use for a Content Item in CDS.
`PNG LRG` = Smaller than 4096px, no interlacing
`PNG LRG_ICO` = Smaller than 120px, no interlacing  and not to to use for a Content Item in CDS.
`PNG_TN` =  PNG Thumbnnail must be smaller than 160px, only for thumbnails

(via https://github.com/da2ce7/libdlna/blob/master/src/dlna/image_png.c)

KarlStraussberger:
Your link can be a good source for ideas to make the code more readable for images.

doc/config-import.rst
```diff
@@ -957,13 +957,19 @@ This section holds the content type to dlnaprofile mappings.
...
-attributes are case sensitive. There are profiles for special codecs which can be access by the pattern ``<content-type>-<video-codec>-<audio-codec>``.
-Video and audio codecs can be seen in the resources properties the web UI.
+attributes are case sensitive.
+In order to access special profiles you can specify a resource attribute with its required value. Mappings with more details 
+are preferred to simple from-to mappings. Resource attributes can be seen in the details page for an item on the web UI.
+
+* Example:
+::
+
+    <map from="mp4" sec:vcodec="h264" sec:acodec="aac" to="AVC_MP4_MP_HD_720p_AAC"/>
```
whyman, line 972: How about `audio_codec` or `audioCodec`, seems a bit "low level" for config

KarlStraussberger: This is the spelling of the resource attribute. Each attribute can be referenced here.

whyman: Yeah but I don't think that is a very friendly thing to have user config

whyman: How we will we handle the resolution case? Or say SD Vs HD profiles? We would need GT/LT, but resolution is a string.

Simple thing would be to split it to height and width, but no Fields exist for that.

I am thinking something less directly tied to the tags is preferabe

whyman (approved):

doc/config-import.rst
KarlStraussberger: Good point. So let's see this as an intermediate step and I think about some transformations for attributes and values to deal with the usability requirement.

PR #2535, opened 2022-04-01
use constexpr parameter for fmt::format

neheb:
Fixes compilation with fmt 8+ and c++20.

### DIFF
--- a/src/database/sqlite3/sqlite_database.cc
+++ b/src/database/sqlite3/sqlite_database.cc
@@ -164,7 +164,7 @@ void Sqlite3Database::init()
     } catch (const std::runtime_error& e) {
         log_error("prematurely shutting down.");
         shutdown();
-        throw_std_runtime_error(e.what());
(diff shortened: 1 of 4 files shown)


neheb:
No real idea how to handle the remaining ones:

https://github.com/neheb/gerbera/commit/3652e9f58f5091c248eae99b51ec4bf0aea712f0
https://github.com/neheb/gerbera/commit/b676e3500130f87eb1b7f757a104b6c1de51b177

neheb:
any other issue?

neheb:
Because fmt 8 uses consteval for the first parameter.

whyman:
Right but with one arg you dont need to format anyway

neheb:
It's equivalent to printf(foo) vs printf("%s", foo). Compiler complains about the former.

This pattern is already done elsewhere with fmt. Might as well maintain it.

neheb:
Rebased.

Issue #2531, opened 2022-04-01
Gerbera 1.10.0 files not supported on UPNP device (Panasonic TV)


lukesoft76:
I have updated to Gerbera 1.10.0 on Debian via deb-multimedia.org packages.
With older version, my Panasonic TV did show the folders, but know it claims that the files(folders) are not supported.

lukesoft76:
After digging deeper, it looks like the Panasonic TV has a problem with the virtual containers and how they are created.
I switched back to 1.1.0+dfsg-3+b6 provided by the official debian packages system. 
The new versions show nice features but the documentation is not really helpful to create an own import.js files.

whyman:
That Debian version is very old.

Is it particular containers that aren't working? Or just any container?

KarlStraussberger:
Did you recreate your config file with `gerbera --create-config` and merge your old settings? There have been quite some changes. If you did so, you can try to remove `containers` which cause trouble with clients that cache all metadata before showing any files. Next stop could be to rebuild the database from scratch.

TaniGit:
I had same problem. But I resolved already.
Problem1. Panasonic VIERA do not accept none ascii <res>(like utf-8 filenaem)</res>
Problem2. require SortCriteria [+upnp:class,+dc:date]. usualy Gerbera not set dc:date.

now I am making simple patch for 1.10.0.
[gerbera-1.10.0-viera.patch.txt](https://github.com/gerbera/gerbera/files/8402381/gerbera-1.10.0-viera.patch.txt)

now I can use on client info:  'Panasonic MIL DLNA CP UPnP/1.0 DLNADOC/1.50'

KarlStraussberger:
@TaniGit Thank you for the detailed analysis and attempt to fix. I'm sure we can work out a solution that suits all of our users.

TaniGit:
oops... sorry  my patch was forgot update src/database/sqlite3/sqlite3.sql....
need insert  below

INSERT INTO \`mt_metadata\` (
  \`id\`, \`item_id\`, \`property_name\`, \`property_value\`
) VALUES (NULL, \'1\', \'dc:date\', \'1900-01-01T00:00:00Z\');
[gerbera-1.10.0-viera-r1.patch.txt](https://github.com/gerbera/gerbera/files/8402658/gerbera-1.10.0-viera-r1.patch.txt)

And for a brief description of patch,

- for Problem1
  - return filename is stick file.\'ext\'. remove real filename.

- for Problem2
[...]

Issue #2017, opened 2021-10-11
Samsung subtitles are broken again


druban:
They used to work a month or so ago. What kind of debug info should I provide?

KarlStraussberger:
Looks like your client is not handled as SAMSUNG device anymore: `getInfo(): client info: 192.168.6.109:47995 'DLNADOC/1.50' -> 'Manual Setup for IP 192.168.6.109'`. There is no `CaptionInfo.sec`

druban:
Here's relevant config entry:
```
  <clients enabled="yes">
    <!-- <client ip="192.168.6.109" userAgent="DLNADOC/1.50" flags="SAMSUNG"></client> -->
    <client ip="192.168.6.109" flags="SAMSUNG"></client>
  </clients>
```
I haven't changed it in quite some time ....

KarlStraussberger:
I don't see real issues, but found a small artefact which is in PR #2018: `sec:type` was empty

KarlStraussberger:
Can you try opening the url, please. It should download the sub file.

KarlStraussberger:
We've split the SAMSUNG flag to allow control of response and improve X_GetFeatureList support.

Can you try current head, please?

druban:
No, didn't help. But I uncovered another issue - gerbera crashes now because logfile is not initialized.
To fix:
```
diff --git a/src/main.cc b/src/main.cc
index 84fde55f..389f0ad2 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -219,6 +219,7 @@ int main(int argc, char** argv, char** envp)
 
         std::optional<fs::path> logfile;
         if (opts.count("logfile") > 0) {
+            logfile = opts["logfile"].as<fs::path>();
             if (!fs::directory_entry(logfile->parent_path()).exists()) {
                 log_warning("Log dir {} missing", logfile->parent_path().c_str());
             }
```

druban:
As for the subtitles, I suspect it might be related to libraries I'm linking against. I tried reverting to the code line when I knew subtitles worked, but I could get them to work.
Here's output of ldd:
```
/usr/local/bin/gerbera:
	libupnp.so.17 => /usr/local/lib/libupnp.so.17 (0x800858000)
	libixml.so.11 => /usr/local/lib/libixml.so.11 (0x8008a3000)
	libspdlog.so.1 => /usr/local/lib/libspdlog.so.1 (0x8008b3000)
[...]

PR #2540, opened 2022-04-03
Add Quirk to block filename in item uri

KarlStraussberger:
second part of solution for #2531
may even help with #2017

### DIFF
--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -471,31 +471,31 @@ std::pair<std::string, bool> UpnpXMLBuilder::renderItemImage(const std::string& 
     return {};
 }
 
-std::pair<std::string, bool> UpnpXMLBuilder::renderSubtitle(const std::string& virtualURL, const std::shared_ptr<CdsItem>& item)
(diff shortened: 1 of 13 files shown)


src/upnp_xml.cc
```diff
@@ -471,31 +471,31 @@ std::pair<std::string, bool> UpnpXMLBuilder::renderItemImage(const std::string&
     return {};
 }
 
-std::pair<std::string, bool> UpnpXMLBuilder::renderSubtitle(const std::string& virtualURL, const std::shared_ptr<CdsItem>& item)
+std::pair<std::string, bool> UpnpXMLBuilder::renderSubtitle(const std::string& virtualURL, const std::shared_ptr<CdsItem>& item, const std::unique_ptr<Quirks>& quirks)
```
whyman, line 474: Maybe just pass a reference to the Quirks here, not sure it makes sense to pass a reference to a unique_ptr

KarlStraussberger: Sometimes you do not see the obvious

KarlStraussberger: Needed to roll it back, because quirks are nullable. Crashed in tests

KarlStraussberger: Pointers work

PR #2543, opened 2022-04-06
Autoscan: Tidy validation logic, nest enum

whyman:


src/web/containers.cc
```diff
@@ -86,7 +86,7 @@ void Web::Containers::process()
 #ifdef HAVE_INOTIFY
             if (config->getBoolOption(CFG_IMPORT_AUTOSCAN_USE_INOTIFY)) {
                 auto adir = database->getAutoscanDirectory(cont->getID());
-                if (adir && (adir->getScanMode() == ScanMode::INotify))
+                if (adir && (adir->getScanMode() == AutoscanDirectory::ScanMode::INotify))
                     autoscanMode = "inotify";
```
KarlStraussberger, line 90: Shouldn't we call `AutoscanDirectory::mapScanmode` here as well

whyman: Yes, good idea.

whyman: Actually, no this is mapping the CDSObjects autoscan field, not an autoscan itself, its still just a bunch of #defines

I am hoping to clean this up when collapsing the multiple "AutoscanList" into one

PR #2545, opened 2022-04-07
ContentHandler to enum

KarlStraussberger:


src/metadata/metadata_handler.h
```diff
@@ -46,20 +46,23 @@ class CdsObject;
...
-#define CH_FFTH 6
-#define CH_FLAC 7
-#define CH_FANART 8
-#define CH_MATROSKA 9
-#define CH_SUBTITLE 10
-#define CH_CONTAINERART 11
-#define CH_WAVPACK 12
-#define CH_RESOURCE 20
+enum class ContentHandler {
+    UNKNOWN = -1,
```
whyman, line 50: Why do we need a UNKNOWN type? Wouldn't DEFAULT work better? and then it can be unsigned too.

src/metadata/taglib_handler.cc
```diff
@@ -640,7 +640,7 @@ void TagLibHandler::extractFLAC(TagLib::IOStream& roStream, const std::shared_pt
 
     auto audioProps = flac.audioProperties();
     auto temp = audioProps->bitsPerSample();
-    auto res = item->getResource(0);
+    auto res = item->getResource(ContentHandler::DEFAULT);
```
whyman, line 643: So much nicer.

src/metadata/metadata_handler.cc
```diff
@@ -201,42 +218,48 @@ std::string MetadataHandler::getMimeType() const
...
 
-int MetadataHandler::remapContentHandler(const std::string& contHandler)
+ContentHandler MetadataHandler::remapContentHandler(const std::string& contHandler)
 {
     auto chEntry = std::find_if(chKeys.begin(), chKeys.end(), [contHandler](auto&& entry) { return contHandler == entry.second; });
     if (chEntry != chKeys.end()) {
         return chEntry->first;
     }
-    return -1;
+    return ContentHandler::UNKNOWN;
```
whyman, line 254: I think it would be better to make this method throw and not have the unknown value.

whyman (commented):
Nice!

src/metadata/metadata_handler.cc
KarlStraussberger: This is one reason for it: https://github.com/KarlStraussberger/gerbera/blob/c419bb294f3260742f76ee8376f2910e6f52be7b/src/upnp_xml.cc#L55
We do not have a check for the config yet.

And the extra check for https://github.com/KarlStraussberger/gerbera/blob/c419bb294f3260742f76ee8376f2910e6f52be7b/src/cds_resource.h#L120-L123. This is easy to fix.

whyman: re `isMetaResource` IMO we should get rid of that method (not in this change)

My thinking is that we should add a Purpose enum to Resource too:

`Purpose { Content, Thumbnail, Subtitle }`

Then you can tell the purpose of the resource explicitly.

whyman: And for the other `CFG_IMPORT_RESOURCES_ORDER` it simply skips it if its not mapped anyway, which could be as handled with a try/catch.

Or if you wanted it to be explicit it should return an optional.

KarlStraussberger: > And for the other `CFG_IMPORT_RESOURCES_ORDER` it simply skips it if its not mapped anyway, which could be as handled with a try/catch.
> 
> Or if you wanted it to be explicit it could return an optional.

Moved that into the check function of the config value. So it throws on startup now

whyman (approved):

PR #2546, opened 2022-04-07
Remove DB field from autoscan list

whyman:


src/config/setup/config_setup_autoscan.cc
```diff
@@ -197,7 +197,7 @@ void ConfigAutoscanSetup::makeOption(const pugi::xml_node& root, const std::shar
 
 std::shared_ptr<ConfigOption> ConfigAutoscanSetup::newOption(const pugi::xml_node& optValue)
```
neheb, line 198: maybe match pointer types with result?

neheb: nvm this is not returned.

PR #2549, opened 2022-04-10
fix bad URL

neheb:
: is used to specify port, not the URL.


KarlStraussberger:
Nice finding.

PR #2550, opened 2022-04-10
Revert "fix bad URL"

KarlStraussberger:
Reverts gerbera/gerbera#2549

The documentation says it is appended to the IP to set the port


whyman:
Yep, I was going to comment.

Could do witha test or two

PR #2551, opened 2022-04-10
Config: Autoscan list to plain vector

whyman:


whyman:
Im not really sure it makes sense to allow editing the autoscan items from config via the config UI.

If you want to edit them via the UI just do it via the usual method...

1st stage of moving out autoscan management out of content manager.

src/config/config_setup.h
```diff
@@ -662,9 +662,9 @@ class ConfigAutoscanSetup : public ConfigSetup {
     /// \brief Creates an array of AutoscanDirectory objects from a XML nodeset.
     /// \param element starting element of the nodeset.
     /// \param scanmode add only directories with the specified scanmode to the array
-    bool createOptionFromNode(const pugi::xml_node& element, std::shared_ptr<AutoscanList>& result);
+    bool createOptionFromNode(const pugi::xml_node& element, std::vector<AutoscanDirectory> result);
```
KarlStraussberger, line 665: You need a reference otherwise no result is returned

whyman: Good spot

src/config/config_options.h
```diff
@@ -233,15 +234,15 @@ class ArrayOption : public ConfigOption {
 
 class AutoscanListOption : public ConfigOption {
 public:
-    explicit AutoscanListOption(std::shared_ptr<AutoscanList> option)
-        : option(std::move(option))
+    explicit AutoscanListOption(std::vector<AutoscanDirectory> autoscans_)
+        : autoscans(std::move(autoscans_))
```
neheb, line 238: shouldn't this be opposite? WebKit style says _ is for member variables AFAIK.

src/config/config_setup.h
whyman: Yep

src/config/config.h
```diff
@@ -30,8 +30,8 @@
 #include "util/grb_fs.h"
 
 // forward declarations
-class AutoscanDirectory;
 class AutoscanList;
```
KarlStraussberger, line 33: Isn't this obsolete now?

KarlStraussberger:
> Im not really sure it makes sense to allow editing the autoscan items from config via the config UI.

We should keep the option to check the settings, but normally this is done in config.xml

whyman:
Yeah, I didnt remove it, was more a general comment.

KarlStraussberger:
Ready to merge?

whyman:
I think so!

There isnt great test coverage for this, but I've run it locally and done some things and it seemed to work as I would expect.

PR #2553, opened 2022-04-12
Change ResourceContentType to Resource::purpose

### LINKED ISSUE #2547: Add purpose to CdsResource
Re. `isMetaResource` IMO we should get rid of that method (not in this change)

My thinking is that we should add a Purpose enum to Resource too:

`Purpose { Content, Thumbnail, Subtitle }`

Then you can tell the purpose of the resource explicitly.

_Originally posted by @whyman in https://github.com/gerbera/gerbera/pull/2545#discussion_r844983905_

KarlStraussberger:
Closes #2547


KarlStraussberger:
It may handle some weird stuff (like EXT_URL) different now. But that's only worth digging in if some users complain.

KarlStraussberger:
@whyman can you please take a look at https://builds.sr.ht/~whyman/job/734962.

src/content/autoscan.h
```diff
@@ -148,7 +148,7 @@ class AutoscanDirectory {
     }
     bool isValid() const
     {
-        return scanID == INVALID_SCAN_ID;
+        return scanID != INVALID_SCAN_ID;
```
whyman, line 151: Whoops! 🤦

PR #2558, opened 2022-04-13
replace make_unique with reinterpret_pointer_cast

neheb:
content is a shared_ptr. reinterpret_pointer_cast is the correct thing
to do here. There's also a quite significant size reduction.


neheb:
hmm this seems wrong. The constructor parameter is also const std::shared_ptr<ContentManager>& content, which means no need for std::move. clang-tidy can't see it because the constructor is implemented with "using TranscodeHandler::TranscodeHandler;"

PR #2560, opened 2022-04-13
move propertyMap out of header

neheb:
Avoids placing array in a header.


whyman:
Perhaps it should just be a private static member?

PR #2564, opened 2022-04-14
Try to fix git issue with safe-directory

KarlStraussberger:


whyman:
Took a few attempts but should be fixed by #2566

PR #2565, opened 2022-04-14
Add item class filter and mapping to autoscan directory

### LINKED ISSUE #2542: Cover and Folder-Arts are imported to Photos Section
The folder images/art images from from other sources like music/videos folders are imported to photo section. Cluttering the photo section with unwanted images like:

folder.jpg
AlbumArt*.jpg
discart.png
thumb.jpg

... and many more.

A solution from other UpNP Servers would be to specify the content of added folders to only search for the correct filetypes inside instead of everything.

KarlStraussberger:
The files are still imported to PC Directory but the layout is not processes.

closes #2542


whyman:
Interesting approach, I was thinking of adding an "MediaType" to the Autoscan object, because there are cases where we want to disambiguate the type of audio/video too.

```c++
enum class MediaType {
Any,
Music,
Photos,
Movie,
TV,
AudioBook,
}
```

Then we can decide based on that

KarlStraussberger:
> Interesting approach, I was thinking of adding an "MediaType" to the Autoscan object, because there are cases where we want to disambiguate the type of audio/video too.
> 
> ```c++
> enum class MediaType {
> Any,
> Music,
> Photo,
> Movie,
> TV,
> AudioBook,
> }
> ```
> 
> Then we can decide based on that
> 
> Also enables us to present TV/Film and Music/Audiobooks differently (Handle series etc)

I intended to disable any kind of item. If we have TV and Audiobook as UpnpClass  it can be added in the same way. If we chose the MediaType approach we might make it a bitfield to allow any combination (there are music videos in my audio tree; there are movies in my images tree taken by mobile etc.)

whyman:
> music videos

Yes, that is a case I missed from the enum, I think a bitmask would work nicely.

We cant detect what the "content" of the files is, we need a hint from the user to do the right thing.

I also have videos in my photos as you would expect, but I would want them presented in my photo albums, not treated like say a DVD rip with actor metadata and subtitltes.

When I read https://github.com/gerbera/gerbera/issues/2542 what they want is not to filter out images specifically (though that would be a fix, admittedly), but not have Gerbera treat those random images as photos which we could have done if we had known that import directory is "Music"

KarlStraussberger:
This ends in mapping the mediaType to layout functions (and splitting them)
- Music (still UPNP_CLASS_MUSIC_TRACK ?) -> AddAudio
- AudioBook (new UPNP_CLASS_AUDIO_BOOK) -> AddAudioBook
- Photo -> AddImage
- Movie -> AddVideo
- ...

Will rework it to handle current types.

KarlStraussberger:
It now allows filtering by the detailed media type.

I'm still thinking about the best way to set the detailed upnp class for an entry
- somewhere in the layout or
- in content manager or
- in metadata logic

KarlStraussberger:
> It now allows filtering by the detailed media type.
> 
> I'm still thinking about the best way to set the detailed upnp class for an entry
> 
>     * in content manager or

Can be configured as
```
<mimetype-upnpclass>
	<map from="application/ogg" to="object.item.audioItem.musicTrack"/>
	<map from="audio/*" to="object.item.audioItem"/>
	<map from="audio/*;tracknumber>0" to="object.item.audioItem.musicTrack"/>
	<map from="audio/*;upnp:genre=Book" to="object.item.audioItem.audioBook"/>
	<map from="image/*" to="object.item.imageItem"/>
	<map from="image/*;location=Camera" to="object.item.imageItem.photo"/>
	<map from="video/*" to="object.item.videoItem"/>
</mimetype-upnpclass>
```

Tests are still pending. Code needs some cleanup.

KarlStraussberger:
Runtime tests were successful now. Autoscan modal is still pending.

KarlStraussberger:
grafik

whyman:
This is a big one!

KarlStraussberger:
Small changes in lots of places.

PR #2566, opened 2022-04-14
CI: Debian: whitelist /build

whyman:


whyman:
Build is running over at https://github.com/whyman/gerbera/runs/6022985954

PR #2584, opened 2022-04-16
remove const from map

neheb:
Avoids clang-tidy warning about the return value of template being const
V.


whyman:
This map should be const.

neheb:
The map is const.

whyman:
Ah yes, so just entirely inaccurate commit message and PR title....

PR #2587, opened 2022-04-17
clang-tidy: use at() instead of []

neheb:

I changed ++apos to apos++ since the latter seems to be the correct thing to do. Please check.


whyman:
It's not the correct thing, consider they are arguments

neheb:
OK

PR #2589, opened 2022-04-17
pass various setters by value

neheb:
Allows using std::move().


whyman:
I dont see how this better? Its still a single copy either way.

Issue #2590, opened 2022-04-17
search-item-result not applied unless mark-played-items is also enabled.


mill1000:
Looks like commit a59ccafe61e63d93f356b1ce5b2080ac8d3feaeb introduced a regression in the handling of `<search-item-result>` .  The item's title isn't updated unless `mark-played-items` is enabled and the item needs to be marked.

PR #2591, opened 2022-04-17
Fix search-item-result regression.

mill1000:
Fix regression described in #2590 by always updating the item's title.


KarlStraussberger:
Thank you.

Issue #2595, opened 2022-04-18
libexif: Non-exif JPEG size parsing not working


whyman:
![image](https://user-images.githubusercontent.com/345034/163878813-237c40c2-d136-4620-a7b9-b8a5585f184f.png)

Broken thumbnail for some reason, incorrect resolution for both thumbnail and the main content.

neheb:
does exiv2 work fine?

whyman:
Havent tried exiv2, the files are all valid, I think it something to do with the fallback manual parsing code in `jpeg_resolution.cc`

PR #2596, opened 2022-04-19
Fix built in JPEG resolution parsing.

whyman:
Looks like it was broken with the std::byte changes.
Added a test to make fixing the bug easier.
Added Resolution object and tests
Various related tidy ups.

Fixes: https://github.com/gerbera/gerbera/issues/2595


src/metadata/resolution.cc
```diff
@@ -0,0 +1,68 @@
...
+    You should have received a copy of the GNU General Public License
+    along with Gerbera.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/// \file resolution.cc
+
+#include "resolution.h"
+#include "util/tools.h"
+
+Resolution::Resolution(std::string string)
```
neheb, line 27: const string ref

neheb:
```
 /home/runner/.conan/data/gerbera/dev/_/_/build/d9eb8116466780ac9e52023ad71eb294759c4452/src/util/jpeg_resolution.cc:124:17: error: comparison of integer expressions of different signedness: ‘long unsigned int’ and ‘int’ [-Werror=sign-compare]
  124 |         if (got != itemLen - 2) {
```

web/js/jquery.gerbera.editor.js
```diff
@@ -162,7 +162,7 @@
     if (item) {
       reset(modal);
       if (item.image) {
-        modal.find('#mediaimage').prop('src', item.image.value);
+        modal.find('#mediaimage').prop('src', '//' + item.image.value);
```
KarlStraussberger, line 165: What's the background?

`image` is always the full path starting with virtual Url?

whyman: Yes, but Firefox refused to load it without a protocol element (http://) so I used the "//" which means match the pages protocol

KarlStraussberger: What's your virtualUrl setting in config?

whyman: Good question, I'm on holiday for a few days

KarlStraussberger: Enjoy it!

KarlStraussberger: Did you check you config. The change breaks my WebUI

whyman: So, I have virtualURL as `hostname:port` in config.

I guess you have `http://hostname:port` which would have broken.

I will put together a PR to fix it.

PR #2604, opened 2022-04-24
Respect ContainerID when performing search

### LINKED ISSUE #2364: Search does not respect ContainerID
Another search issue.

Gerbera v1.9.2-290-g1ec9b7ce
Hi-Fi Cast v1.127

It looks like the ContainerID of the Search action is not being respected.

Hi-Fi Cast is generating a search with the following parameters.
> 2022-01-03 14:20:08.729  debug: [upnp_cds.cc:171] doSearch(): Search received parameters: ContainerID [83] SearchCriteria [upnp:class derivedfrom "object.item.audioItem.musicTrack"] StartingIndex [0] RequestedCount [30] RequestedCount [30]

It is attempting to a get a list of all tracks in a given container. In this case, an album that was marked as searchable upon import.

Gerbera converts the search to the follow query.
> 2022-01-03 14:20:08.729  debug: [sql_database.cc:1029] search(): Search count resolves to SQL [SELECT COUNT(DISTINCT "c"."id") FROM "mt_cds_object" "c" INNER JOIN "mt_metadata" "m" ON "c"."id" = "m"."item_id" INNER JOIN "grb_cds_resource" "re" ON "c"."id" = "re"."item_id" WHERE (LOWER("c"."upnp_class") LIKE LOWER('object.item.audioItem.musicTrack%')) AND ("c"."flags" & 2 = 2 OR "c"."object_type" != 1)]

This returns all the entries in the database since the ContainerID was not respected.

Gerbera should be mapping ContainerID to parent_id but its more complicated than that.

In my DB, ContainerID 83 points to a virtual item. The album which was marked searchable.
image
But it only contains virtual objects that reference the actual track items.
image

The actual tracks have a parent_id of 80
image
which is the ID of the physical container
image

mill1000:
Respect ContainerID when performing searches against the database. Close #2364.

This is implemented using a recursive query that finds all children of a given parent_id. The children's ref_id's are then followed to find the actual physical items.

I am a little unclear on the desired styles of this project. There appears to be mix of `snake_case` and `camelCase` for variable names. Some accessors also take the form`getSomeVariable()` while others are `someVariable()`. I will gladly update the changes to match the desired style.

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -386,10 +386,11 @@ void SQLDatabase::init()
 
         auto join1 = fmt::format("LEFT JOIN {} ON {} = {}", metaColumnMapper->tableQuoted(), searchColumnMapper->mapQuoted(UPNP_SEARCH_ID), metaColumnMapper->mapQuoted(UPNP_SEARCH_ID));
         auto join2 = fmt::format("LEFT JOIN {} ON {} = {}", resourceColumnMapper->tableQuoted(), searchColumnMapper->mapQuoted(UPNP_SEARCH_ID), resourceColumnMapper->mapQuoted(UPNP_SEARCH_ID));
+        this->sql_search_query = fmt::format("{} {} {}", searchColumnMapper->tableQuoted(), join1, join2);
 
-        // Build container query and final search query
-        auto sql_container_query = fmt::format(container_query_raw, searchColumnMapper->tableQuoted(), searchColumnMapper->getAlias(), searchColumnMapper->mapQuoted(UPNP_SEARCH_PARENTID, true), searchColumnMapper->mapQuoted(UPNP_SEARCH_ID, true), searchColumnMapper->mapQuoted(UPNP_SEARCH_REFID, true));
-        this->sql_search_query_format = fmt::format("{} {} {}", sql_container_query, join1, join2);
+        // Build container query format string
+        auto sql_container_query = fmt::format(sql_search_container_query_raw, searchColumnMapper->tableQuoted(), searchColumnMapper->getAlias(), searchColumnMapper->mapQuoted(UPNP_SEARCH_PARENTID, true), searchColumnMapper->mapQuoted(UPNP_SEARCH_ID, true), searchColumnMapper->mapQuoted(UPNP_SEARCH_REFID, true));
+        this->sql_search_container_query_format = fmt::format("{} {} {}", sql_container_query, join1, join2);
     }
     // Statement for metadata
     {
@@ -1040,11 +1041,21 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::search(const SearchParam& p
             searchColumnMapper->mapQuoted(SearchCol::Flags), OBJECT_FLAG_SEARCHABLE, searchColumnMapper->mapQuoted(SearchCol::ObjectType), OBJECT_TYPE_CONTAINER));
     }
 
-    beginTransaction("search");
-    const std::string countSelect = fmt::format("COUNT(DISTINCT {})", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID));
-    auto countSQL = fmt::format(this->sql_search_query_format, param.getContainerID(), countSelect);
-    countSQL += fmt::format(" WHERE {}", searchSQL);
+    bool rootContainer = param.getContainerID() == "0";
+
+    std::string countSQL;
+    if (rootContainer) {
+        // Use fater, non-recursive search for root container
+        countSQL = fmt::format("SELECT COUNT(DISTINCT {}) FROM {} WHERE {}", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID), sql_search_query, searchSQL);
+    } else {
+        // Use recursive container search
+        const std::string countSelect = fmt::format("COUNT(DISTINCT {})", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID));
+        countSQL = fmt::format(this->sql_search_container_query_format, param.getContainerID(), countSelect);
+        countSQL += fmt::format(" WHERE {}", searchSQL);
+    }
+
     log_debug("Search count resolves to SQL [{}]", countSQL);
+    beginTransaction("search");
     auto sqlResult = select(countSQL);
     commit("search");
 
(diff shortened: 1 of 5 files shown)


mill1000:
It appears this implementation has a pretty substantial impact on the performance when the root container is queried.

src/database/sql_database.cc
```diff
@@ -1037,9 +1041,21 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::search(const SearchParam& p
             searchColumnMapper->mapQuoted(SearchCol::Flags), OBJECT_FLAG_SEARCHABLE, searchColumnMapper->mapQuoted(SearchCol::ObjectType), OBJECT_TYPE_CONTAINER));
     }
 
-    beginTransaction("search");
-    auto countSQL = fmt::format("SELECT COUNT(DISTINCT {}) FROM {} WHERE {}", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID), sql_search_query, searchSQL);
+    bool rootContainer = param.getContainerID() == "0";
+
+    std::string countSQL;
+    if (rootContainer) {
+        // Use fater, non-recursive search for root container
```
KarlStraussberger, line 1048: just a little typo

src/database/sql_database.cc
```diff
@@ -1037,9 +1041,21 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::search(const SearchParam& p
...
+        countSQL = fmt::format("SELECT COUNT(DISTINCT {}) FROM {} WHERE {}", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID), sql_search_query, searchSQL);
+    } else {
+        // Use recursive container search
+        const std::string countSelect = fmt::format("COUNT(DISTINCT {})", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID));
+        countSQL = fmt::format(this->sql_search_container_query_format, param.getContainerID(), countSelect);
```
KarlStraussberger, line 1053: no need for `this` inside normal methods.

src/database/sql_database.h
```diff
@@ -226,9 +226,25 @@ class SQLDatabase : public Database {
...
+SELECT {1}.* FROM {0} JOIN containers AS cont ON {1}.{2} = cont.{3}
+),
+-- Find all physical items and de-reference any virtual item (i.e. follow ref-id)
+items AS (SELECT * from containers AS cont WHERE {4} IS NULL
+UNION
+SELECT {1}.* FROM containers AS cont JOIN {0} ON cont.{4} = {1}.{3}
+)
+-- Select desired cols from items
+SELECT {{}} FROM items AS {1})";
+
```
KarlStraussberger, line 243: Try to move that to the code file to make have all SQL statements in the same spot. It should you `identifier("containers")` etc. calls for code safety,

mill1000: Took a wack at what you suggest with the `identifier` calls. Not sure I see the benefit but I don't see the harm either.

src/database/sql_database.cc
```diff
@@ -1037,9 +1056,21 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::search(const SearchParam& p
...
+
+    std::string countSQL;
+    if (rootContainer) {
+        // Use faster, non-recursive search for root container
+        countSQL = fmt::format("SELECT COUNT(DISTINCT {}) FROM {} WHERE {}", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID), sql_search_query, searchSQL);
+    } else {
+        // Use recursive container search
+        const std::string countSelect = fmt::format("COUNT(DISTINCT {})", searchColumnMapper->mapQuoted(UPNP_SEARCH_ID));
+        countSQL = fmt::format(sql_search_container_query_format, param.getContainerID(), countSelect);
+        countSQL += fmt::format(" WHERE {}", searchSQL);
```
KarlStraussberger, line 1069: If you add ` WHERE {} ` to `sql_search_container_query_format` you can merge the statements

src/database/sql_database.cc
```diff
@@ -1069,7 +1100,16 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::search(const SearchParam& p
...
-    auto retrievalSQL = fmt::format("SELECT DISTINCT {} {} FROM {} {} WHERE {}{}{}", sql_search_columns, addColumns, sql_search_query, addJoin, searchSQL, orderBy, limit);
+    std::string retrievalSQL;
+    if (rootContainer) {
+        // Use faster, non-recursive search for root container
+        retrievalSQL = fmt::format("SELECT DISTINCT {} {} FROM {} {} WHERE {}{}{}", sql_search_columns, addColumns, sql_search_query, addJoin, searchSQL, orderBy, limit);
+    } else {
+        // Use recursive container search
+        const std::string retrievalSelect = fmt::format("DISTINCT {} {}", sql_search_columns, addColumns);
+        retrievalSQL = fmt::format(sql_search_container_query_format, param.getContainerID(), retrievalSelect);
+        retrievalSQL += fmt::format(" {} WHERE {}{}{}", addJoin, searchSQL, orderBy, limit);
```
KarlStraussberger, line 1111: You may need a second property to avoid that `+=` here - or add empty parameters above.

src/database/sql_database.cc
```diff
@@ -299,6 +299,21 @@ static const std::vector<std::pair<std::string, AutoscanColumn>> autoscanTagMap
...
+WITH {0} AS (SELECT * FROM {2} WHERE {4} = {{}}
+UNION
+SELECT {3}.* FROM {2} JOIN {0} AS {1} ON {3}.{4} = {1}.{5}
+),
+-- Find all physical items and de-reference any virtual item (i.e. follow ref-id)
+items AS (SELECT * from {0} AS {1} WHERE {6} IS NULL
+UNION
+SELECT {3}.* FROM {0} AS {1} JOIN {2} ON {1}.{6} = {3}.{5}
+)
+-- Select desired cols from items
```
KarlStraussberger, line 314: DId you check `--` comments in statements for mysql and sqlite APIs?

mill1000: Yes it looks like [MySQL](https://dev.mysql.com/doc/refman/5.7/en/comments.html) and [SQLite](https://www.sqlite.org/lang_comment.html) support `--` as comments

src/database/sql_database.cc
mill1000: Feels wasteful to create an near duplicate property just to avoid a format call and concatenation.

KarlStraussberger:
Thank you

mill1000:
Happy to help. With #2364 closed I think I can make Gerbera my primary media server.

whyman:
> I am a little unclear on the desired styles of this project. There appears to be mix of snake_case and camelCase for variable names. 

Older stuff uses a mix, we are gradually moving everything to camelCase.

> Some accessors also take the formgetSomeVariable() while others are someVariable(). I will gladly update the changes to match the desired style.

Mostly we would go with get prefixes.

Generally we follow https://webkit.org/code-style-guidelines.

We should extend the CONTRIBUTING a bit.

mill1000:
Cool thanks for the heads up.

PR #2606, opened 2022-04-25
CMake: Fetch GTest if needed

whyman:


whyman:
Conan is great in principle, painful in reality!

PR #2618, opened 2022-04-30
clang-tidy fixes

neheb:


whyman:
Please revert statics for public methods

neheb:
Hmm? Any particular issue?

whyman:
Breaks method access encapsulation, you shouldnt be able to call those methods without an instance

neheb:
Removed.

PR #2621, opened 2022-05-03
github workflows: make cmake stuff consistent

neheb:
Also changed to build with Ninja for speed.


neheb:
having to pass -DWITH_INOTIFY=0 on macOS sounds like a bug.

neheb:
could be replaced by libfswatch. But that would be a separate PR.

whyman:
Can you make ninja print the compiler commands like the `VERBOSE_MAKEFILE` flag does?

whyman:
We cant link to libfswatch because its GPLv3 or Apache 2, both are not compatible with GPLv2.

neheb:
> 

Let's find out.

neheb:
Anything else missing?

.builds/freebsd.yml
```diff
@@ -25,10 +26,9 @@ tasks:
   - build: |
       mkdir build
       cd build
-      CXXFLAGS="-DLIBICONV_PLUG" cmake ../gerbera \
-      -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
+      CXXFLAGS="-DLIBICONV_PLUG" cmake ../gerbera -G Ninja \
       -DWITH_INOTIFY=1 -DWITH_MAGIC=1 -DWITH_MYSQL=0 -DWITH_CURL=1 -DWITH_JS=1 \
       -DWITH_TAGLIB=1 -DWITH_AVCODEC=1 -DWITH_FFMPEGTHUMBNAILER=1 \
       -DWITH_EXIF=1 -DWITH_LASTFM=0 -DWITH_SYSTEMD=0 -DWITH_DEBUG=1 -DWITH_TESTS=1
```
KarlStraussberger, line 32: Can you use or create some cmake preset here as well. Forgot that.

neheb: Actually I’m curious why this is not using actions. vmactions supports FreeBSD.

neheb: Yeah this isn't used looks like...

PR #2640, opened 2022-05-20
clang-tidy: const ref to value

neheb:


KarlStraussberger:
I don't see the benefit. The copy is created anyway and then moved.

neheb:
clang-tidy flags this.

PR #2642, opened 2022-05-23
Update node packages

KarlStraussberger:
Always in hope of getting rid of the random test issues...


KarlStraussberger:
```
deprecated eslint-plugin-standard@5.0.0: standard 16.0.0 and eslint-config-standard 16.0.0 no longer require the eslint-plugin-standard package.
More info here: https://github.com/standard/standard/issues/1316
```

PR #2644, opened 2022-05-25
shared_ptr conversions

neheb:
matches the type instead of dereferencing.


KarlStraussberger:
The changes to `UpnpXMLBuilder` were deliberate because it cannot be nullptr

whyman:
Also generally you don't want to pass a `shared_ptr` unless you want the function to share ownership, which we do not in this case.

PR #2650, opened 2022-06-02
Coverity fixes

neheb:
Mostly nullptr deref.

Found with Coverity.

### DIFF
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -359,7 +359,8 @@ void ContentManager::shutdown()
     log_debug("start");
     auto lock = threadRunner->uniqueLock();
     log_debug("updating last_modified data for autoscan in database...");
-    autoscanList->updateLMinDB(*database);
+    if (autoscanList)
+        autoscanList->updateLMinDB(*database);
 
 #ifdef HAVE_JS
     destroyJS();
@@ -569,6 +570,9 @@ int ContentManager::_addFile(const fs::directory_entry& dirEnt, fs::path rootPat
 
 bool ContentManager::updateAttachedResources(const std::shared_ptr<AutoscanDirectory>& adir, const std::shared_ptr<CdsObject>& obj, const fs::path& parentPath, bool all)
 {
+    if (!adir)
+        return false;
+
     bool parentRemoved = false;
     int parentID = database->findObjectIDByPath(parentPath, false);
     if (parentID != INVALID_OBJECT_ID) {
@@ -624,8 +628,8 @@ std::vector<int> ContentManager::_removeObject(const std::shared_ptr<AutoscanDir
         if (changedContainers) {
             session_manager->containerChangedUI(changedContainers->ui);
             update_manager->containersChanged(changedContainers->upnp);
+            return changedContainers->upnp;
         }
-        return changedContainers->upnp;
     }
 
     if (obj)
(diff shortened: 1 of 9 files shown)


src/util/upnp_quirks.cc
```diff
@@ -40,7 +40,7 @@
 Quirks::Quirks(const UpnpXMLBuilder& xmlBuilder, const ClientManager& clients, const std::shared_ptr<GrbNet>& addr, const std::string& userAgent)
     : xmlBuilder(xmlBuilder)
 {
-    if (addr || !userAgent.empty())
+    if (addr && !userAgent.empty())
```
KarlStraussberger, line 43: That's breaking things. getInfo has to be called if either `addr` is set or `userAgent`

KarlStraussberger (changes requested):
Fine but the quirks change is breaking!

neheb:
Fixed.

src/content/content_manager.cc
```diff
@@ -768,7 +772,7 @@ void ContentManager::_rescanDirectory(const std::shared_ptr<AutoscanDirectory>&
             continue;
         }
 
-        if (shutdownFlag || (task && !task->isValid()))
+        if (shutdownFlag || !task->isValid())
```
KarlStraussberger, line 775: `task` may be unset

src/database/search_handler.cc
```diff
@@ -677,7 +677,7 @@ std::string SortParser::parse(std::string& addColumns, std::string& addJoin)
         } else {
             log_warning("Unknown sort direction '{}' in '{}'", seg, sortCrit);
         }
-        if (!colMapper || !colMapper->mapQuotedList(sort, seg, (desc ? "DESC" : "ASC"))) {
+        if (!colMapper->mapQuotedList(sort, seg, (desc ? "DESC" : "ASC"))) {
```
KarlStraussberger, line 680: can you add a check for `colMapper` and `metaMapper` in the begin of the method
```
if (sortCrit.empty() || !colMapper || !metaMapper)
```

KarlStraussberger (changes requested):
New changes - new remarks

src/content/content_manager.cc
neheb: Fixed.

src/database/search_handler.cc
neheb: Fixed.

KarlStraussberger (changes requested):
Found one more issue

src/web/edit_load.cc
```diff
@@ -165,6 +165,8 @@ void Web::EditLoad::process()
     xml2JsonHints->setFieldType("rawvalue", "string");
 
     auto objItem = std::dynamic_pointer_cast<CdsItem>(obj);
+    if (!objItem)
+        return;
```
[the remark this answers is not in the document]
neheb: Moved down. I think this should do it.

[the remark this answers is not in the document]
KarlStraussberger: no, there is still container code after it. Anyway, `obj` is used before.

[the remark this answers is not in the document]
neheb: removed.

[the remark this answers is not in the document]
KarlStraussberger: Thank you

PR #2652, opened 2022-06-04
remove lock to avoid waiting while locked

neheb:
Reported by Coverity.


KarlStraussberger:
It should become `atomic_flag` https://en.cppreference.com/w/cpp/atomic/atomic_flag and we should use `test_and_set`. Might solve transaction issues. I actually don't use it anymore

PR #2654, opened 2022-06-06
replace some memcpy by std::copy_n

neheb:
The latter compiles to the former. It's more restrictive as the types
have to match.

### DIFF
--- a/src/util/grb_net.cc
+++ b/src/util/grb_net.cc
@@ -75,7 +75,7 @@ GrbNet::GrbNet(const std::string& addr, int af)
 
 GrbNet::GrbNet(const struct sockaddr_storage* addr)
 {
-    std::memcpy(&sockAddr, addr, sizeof(sockAddr));
+    sockAddr = *addr;
 }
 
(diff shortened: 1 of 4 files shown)


src/util/grb_net.cc
```diff
@@ -75,7 +75,7 @@ GrbNet::GrbNet(const std::string& addr, int af)
 
 GrbNet::GrbNet(const struct sockaddr_storage* addr)
 {
-    std::memcpy(&sockAddr, addr, sizeof(sockAddr));
+    sockAddr = *addr;
```
KarlStraussberger, line 78: You have to use `copy_n` as well, otherwise you may copy the `ss_data` pointer intead of the array contents:

```
 struct sockaddr_storage {
     sa_family_t ss_family;
     uint8_t ss_data[SOCKADDR_MAX_DATA_LEN]; 
 };
```

neheb: Looks the same to me:

https://godbolt.org/z/84Man3GMW

PR #2657, opened 2022-06-09
Simplify Xml2Json

KarlStraussberger:
May be a little slower than the stream version but it is readable.


whyman:
I doubt it will be too much slower, but a tad more memory 👍

PR #2665, opened 2022-06-13
more sonarlint

neheb:

### DIFF
--- a/src/content/scripting/playlist_parser_script.cc
+++ b/src/content/scripting/playlist_parser_script.cc
@@ -94,9 +94,9 @@ duk_ret_t jsReadln(duk_context* ctx)
     }
 
     try {
-        auto line = self->readLine();
-        duk_push_string(ctx, line.first.c_str());
-        return line.second ? 1 : 0;
+        auto [line, trimmed] = self->readLine();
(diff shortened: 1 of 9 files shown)


neheb:
that find_if lambda though...

Need to add a constexpr library to simplify with find().

src/content/scripting/playlist_parser_script.cc
```diff
@@ -94,9 +94,9 @@ duk_ret_t jsReadln(duk_context* ctx)
     }
 
     try {
-        auto line = self->readLine();
-        duk_push_string(ctx, line.first.c_str());
-        return line.second ? 1 : 0;
+        auto [line, trimmed] = self->readLine();
+        duk_push_string(ctx, line.c_str());
+        return trimmed;
```
KarlStraussberger, line 99: We should not implicitely convert `bool` to `duk_ret_t`

neheb: How about now?

KarlStraussberger: I'd leave the original `? 1 : 0`. You see what you get.

neheb: OK.

Issue #2605, opened 2022-04-25
Prematurely shutting down: UNIQUE constraint failed


mill1000:
Running master (183beee1fa6b3eaf598766607e7028b1452303c6) with PR #2604. This happened when I tried to access the server from my control point.
```
> Apr 24 18:58:21 RockHole gerbera[24618]: 2022-04-24 18:58:21  error: prematurely shutting down.
> Apr 24 18:58:21 RockHole gerbera[24618]: 2022-04-24 18:58:21  error: prematurely shutting down.
> Apr 24 18:58:21 RockHole gerbera[24618]: 2022-04-24 18:58:21  error: Exception: [/home/rock/gerbera/gerbera/src/database/sqlite3/sqlite_database.cc:264] virtual int Sqlite3Database::exec(cons
> t string&, bool) Error: [/home/rock/gerbera/gerbera/src/database/sqlite3/sqlite_database.cc:405] void SLTask::waitForTask() Error: SQLITE3: (19, : 1555) UNIQUE constraint failed: grb_playstat
> us.group, grb_playstatus.item_id
> Apr 24 18:58:21 RockHole gerbera[24618]: Query: INSERT INTO "grb_playstatus" ("group","item_id","playCount","lastPlayed","lastPlayedPosition","bookMarkPos") VALUES ('default',13870,3,16508483
> 01,0,0)
> Apr 24 18:58:21 RockHole gerbera[24618]: error: UNIQUE constraint failed: grb_playstatus.group, grb_playstatus.item_id
> Apr 24 18:58:21 RockHole gerbera[24618]: 2022-04-24 18:58:21  error: prematurely shutting down.
> Apr 24 18:58:21 RockHole gerbera[24618]: 2022-04-24 18:58:21  error: Exception: [/home/rock/gerbera/gerbera/src/database/sqlite3/sqlite_database.cc:264] virtual int Sqlite3Database::exec(cons
> t string&, bool) Error: [/home/rock/gerbera/gerbera/src/database/sqlite3/sqlite_database.cc:405] void SLTask::waitForTask() Error: Sorry, sqlite3 thread is shutting down
> Apr 24 18:58:21 RockHole gerbera[24618]: 2022-04-24 18:58:21  error: Exception: [/home/rock/gerbera/gerbera/src/database/sqlite3/sqlite_database.cc:264] virtual int Sqlite3Database::exec(cons
[...]

PR #2666, opened 2022-06-14
Cleanup tasks on removeObject

KarlStraussberger:
Ensure that tasks are cleaned up after each object remove


KarlStraussberger:
Hope the closes #2605

PR #2694, opened 2022-08-19
Fix build with fmt 9.0

### LINKED ISSUE #2681: does not compile with fmt 9.0.0
https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/gerbera/compile.txt

felixonmars:
Fixes #2681

Tested to build fine with both fmt 9.0 and 8.1.1.


kloczek:
Is it possible to make the new release because of that commit?

PR #2702, opened 2022-09-06
Add support for ip subnets in client config

KarlStraussberger:
As inspired by @GinTonic99 in #2700


GinTonic99:
I cannot give any useful comment on the code, this is far beyond my understanding. However let me thank you for your effort. Looking forward to use it in real life.
best regards
gin

PR #2705, opened 2022-09-18
Fix static init

### LINKED ISSUE #2704: Sorting by track number broken
Gerbera 1.11 sorts track numbers alphabetically instead of numerically. As I have a dumb client that simply shows/plays the titles in the order as provided by the server, this is very annoying.

Workaround (hack) is to (pre)"fix" the track number at the beginning of `addAudio`

```js
    if (obj.metaData[M_TRACKNUMBER] && obj.metaData[M_TRACKNUMBER][0]) {
        var track = obj.metaData[M_TRACKNUMBER][0];
        if (track.length == 1) {
            track = '0' + track;
        }
        obj.metaData[M_TRACKNUMBER][0] = track;
    }
```

KarlStraussberger:
Closes #2704

### DIFF
--- a/scripts/versions.sh
+++ b/scripts/versions.sh
@@ -36,7 +36,7 @@ else
     EBML="1.4.2"
     EXIV2="v0.27.5"
     FFMPEGTHUMBNAILER="2.2.2"
-    FMT="9.0.0"
+    FMT="9.1.0"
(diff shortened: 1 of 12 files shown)


KarlStraussberger:
Contains a bit of cleanup in layout handling.

PR #2717, opened 2022-10-10
Path handling cleanup

KarlStraussberger:


KarlStraussberger:
Found while updating my image library

PR #2726, opened 2022-10-21
Fix build on systems that have const in second argument to iconv

niclasr:
Make sure ICONV_CONST is defined if the system has const in its iconv prototype definition.

Use the right variable names when calling iconv with const.


KarlStraussberger:
Thank you for supplying this PR

PR #2729, opened 2022-10-24
Quirks: Avoid passing references

KarlStraussberger:
References from shared_ptr may bypass destruction and usage count


KarlStraussberger:
Added entry for X_SetBookmark in cds.xml to allow UpnpTools to send signals

Issue #2763, opened 2022-12-10
 Segmentation fault on first request on Raspberry PI


vicziani:
After first request from browser/DLNA client Gerbera 1.12.0 throws a segfault. No problem with 1.11.0.

(On my previous Pi Gerbera version 1.12.0 worked for months. But I've reinstalled everything, Raspberry OS 32 bit -> 64 bit Lite, some network modifications, I installed everything with Ansible.)

Environment: 

* Raspberry Pi 3 Model B Plus Rev 1.3
* Debian GNU/Linux 11 (bullseye)
* Docker version 20.10.21
* Gerbera UPnP Server 1.12.0

Log:

```
2022-12-10 13:14:55.217  debug: [thread_runner.h:134] lockGuard(): ThreadRunner: Guard for SQLiteThread - addTask SelectTask
2022-12-10 13:14:55.217  debug: [thread_runner.h:124] notify(): ThreadRunner: Notifying SQLiteThread
2022-12-10 13:14:55.218  debug: [sl_task.cc:118] run(): Running: SELECT "group", "item_id", "playCount", "lastPlayed", "lastPlayedPosition", "bookMarkPos" FROM "grb_playstatus" WHERE "group" = 'default' AND "item_id" = 28
2022-12-10 13:14:55.218  debug: [thread_runner.h:94] wait(): ThreadRunner: Waiting for SQLiteThread
2022-12-10 13:14:55.218  debug: [file_request_handler.cc:255] parseResourceInfo(): Resource ID: 1
Segmentation fault (core dumped)
```

Than I ran with gdb:

```
docker run --network host -v /media/adata/Music:/content:ro --name gerbera-debug gerbera/gerbera tail -f /dev/null
# In another terminal
docker exec -it gerbera-debug  sh
apk update && apk add gdb
gdb /bin/gerbera
(gdb) handle SIGILL pass nostop noprint

(gdb) run --port 49494 --config /var/run/gerbera/config.xml
```

The result:

```
2022-12-10 13:20:35.563  debug: [thread_runner.h:124] notify(): ThreadRunner: Notifying SQLiteThread
2022-12-10 13:20:35.564  debug: [sl_task.cc:118] run(): Running: SELECT "group", "item_id", "playCount", "lastPlayed", "lastPlayedPosition", "bookMarkPos" FROM "grb_playstatus" WHERE "group" = 'default' AND "item_id" = 28
[...]

PR #2764, opened 2022-12-11
MetadataHandler: Use resource instead of id

KarlStraussberger:
Also avoid opening Null resource

See #2763


whyman:
Nice!

KarlStraussberger:
Was just stang by the comment when adding the safety catch. Came out really quick and saves us quite some calls.

PR #2779, opened 2023-01-02
README: fix CI badge

whyman:
Apply change from https://github.com/badges/shields/issues/8671


KarlStraussberger:
Already prepared that locally.

Issue #1440, opened 2021-05-09
Import merges albums with same name even if they have different artists and paths


mill1000:
I have a structure that looks somewhat like this. There are 2 unique artists, and 2 unique tracks that share an album name. They are not on the same album, but simply share a name.
```
Artist1
└── Album
    └── Artist1 - Song.flac
Artist2/
└── Album
    └── Artist2 - Different Song.flac
```

The default JS import merges these tracks into a single album like so. I understand how this behavior might be expected and useful for compilation albums, but I would prefer separate entries.
![image](https://user-images.githubusercontent.com/6924622/117579026-5746fc00-b0ae-11eb-8c7b-118c383ef1af.png)

How can I customize my import scripts to leave these tracks in separate albums **without** changing the albums names? 

I'm trying to mimic the behavior of my previous server. Here's a comparison of how they represent the structure on my control point.
![image](https://user-images.githubusercontent.com/6924622/117579530-83637c80-b0b0-11eb-8254-b35ad5def99b.png)

whyman:
Have you customised your import script?

mill1000:
I have. I've seen this behavior on both the included JS import script and my customized script.

KarlStraussberger:
If you want to avoid unnecessary nesting, I recommend renaming the album.

The brute force solution is to add the artist sub folder in all cases. 

A more usable solution can be a name pattern like `%album title% - %album artist%` which only work if you have album artist tags set in your media. A little bit of java script should be fine here.

mill1000:
Hm...not a fan of changing the displayed album titles. I will experiment with adding artists subfolders and see how that works out with my control points.

Thanks.

mill1000:
Hi Karl.

Adding artist subfolders to albums with shared names is a workable solution, but certainly not ideal. 

For my collection, I can selectively use the subfolders when the album is "Unknown". For these unknown albums I clear the M_ARTIST and M_ALBUMARTIST metadata before adding them to the DB.

There is a quirk with how these collections are represented to control points. I've verified this behavior on Hi-Fi Cast and BubbleUpnp. Depending on the import order, the "Unknown" album containers will inherit artists unexpectedly. I've even seen them inherit artists that don't have any "Unknown" albums.

For example. 
Import order 1
![Import Order 1](https://user-images.githubusercontent.com/6924622/118374600-4567cc00-b57a-11eb-8ca4-07f3e8fe2b36.png)

Import order 2
[...]

PR #2783, opened 2023-01-05
Import: Add staged import process

KarlStraussberger:
1. scan all files
2. create objects for files and directories
3. create layout

Further changes:
- Rework Configuration Mechanism for Enums and Paths
- Allow setting configuration values via command line
- Make virtual folders unique by path
- Reduce overhead of loading data by passing around object references instead of objectID.


.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           python-version: "3.x"
       - name: Install conan
         run: |
-          pip install -U --no-cache-dir conan
+          pip install -U --no-cache-dir conan==1.59
```
gobater, line 44: Not sure if this and further (conan) changes below are related to this feature. If not, for the sake of separation of concerns shall be (whenever possible) extracted into a different PR allowing the changes (if needed) to be integrated as soon as possible into the master -> so other contributions could benefit of it

KarlStraussberger: Good point. It's the time issue again.

src/content/import_service.cc
```diff
@@ -570,31 +607,51 @@ std::pair<int, bool> ImportService::addContainerTree(int parentContainerId, cons
...
+            if (!dirKeyValues.empty()) {
+                subTree = fmt::format("{}@{}", tree, fmt::join(dirKeyValues, "@"));
+            } else {
+                subTree = tree;
+            }
+            auto cont = database->findObjectByPath(subTree, DbFileType::Virtual);
+            if (cont && cont->isContainer())
+                containerMap[subTree] = std::dynamic_pointer_cast<CdsContainer>(cont);
+            else
+                containerMap[subTree] = nullptr;
```
gobater, line 627: I'm not able to follow all changes in this commit, but I guess this is the key part that creates a unique ID within the Albums virtual container...
Considering the changes in "src/config/config_definition.cc", it seems that the album artist and the year are considered and therefore it will probably solve Issue #1440.
Looking forward to see this "in action".
Is this part only of the "grb" import mode or also the "traditional"?

KarlStraussberger: It will be available for both. I'm still testing in real life.

gobater:
@KarlStraussberger ,
I was thinking about adding a new virtual folder (similar to "all songs") within the albums of the artist to list the albums chronologically. It's quite easy to do in Javascript. Do you think it would make sense also to have such a feature in the native importer?

KarlStraussberger:
> @KarlStraussberger , I was thinking about adding a new virtual folder (similar to "all songs") within the albums of the artist to list the albums chronologically. It's quite easy to do in Javascript. Do you think it would make sense also to have such a feature in the native importer?

That's something for JavaScript. Maybe it can be adapted for the release version

whyman:
Wowzers, this is a beast, nice work.

Issue #2679, opened 2022-07-13
Samsung TV not playing videos any more


aleho:
<details> 
<summary>Compile Info</summary>

```
Gerbera UPnP Server 1.11.0
Copyright 2016-2022 Gerbera Contributors.
Licence GPLv2: GNU GPL version 2.
This is free software: you are free to change and redistribute it.

Compile info
-------------
WITH_NPUPNP=NO
WITH_MAGIC=YES
WITH_MYSQL=YES
WITH_CURL=YES
WITH_INOTIFY=YES
WITH_JS=YES
WITH_TAGLIB=YES
WITH_AVCODEC=YES
WITH_FFMPEGTHUMBNAILER=YES
WITH_EXIF=YES
WITH_EXIV2=NO
WITH_MATROSKA=YES
WITH_WAVPACK=YES
WITH_SYSTEMD=NO
WITH_LASTFM=NO
WITH_DEBUG=YES
WITH_TESTS=NO

Git info:
-------------
Git Branch: HEAD
Git Commit: d73d8c1346213d784753c02ea771970500e0df2a
```

</details>

With 1.11.0 / latest in Docker no videos can be played at all.
There are also no thumbnails shown or generated.

Sometimes when starting to play a video, Gerbera just crashes.
[gerbera_1.11.0_tv.log](https://github.com/gerbera/gerbera/files/9104848/gerbera.log)

Using the web UI videos can be played just fine, thumbnails are also generated.

Log file with `docker run -ti --rm --network host --volumes-from gerbera gerbera/gerbera:latest gerbera --port=49494 --config /var/run/gerbera/config.xml --debug`:
[gerbera_1.11.0_web.log](https://github.com/gerbera/gerbera/files/9104850/gerbera.log)

---

With 1.10.0 videos can be played, but after exiting, there's always a crash (apparently trying to set the bookmark).

[gerbera_1.10.0.log](https://github.com/gerbera/gerbera/files/9104893/gerbera1.10.0.log)

jfikar:
The same for LG OLED C9. Gerbera 1.10.0 works, while 1.11.0 does not. For 1.11.0 TV shows no thumbnails and says file format not recognized, when I try to play a file.

Both gerberas use the same pupnp 1.14.12 and I compile it by myself on armhf following:

```
./bootstrap.sh
./configure --prefix=~/upnp --disable-blocking-tcp-connections --enable-reuseaddr
```
and for gerbera:
```
[...]

PR #2799, opened 2023-02-09
Client: remove references to shared_ptrs

whyman:
My theory being that the shared_ptrs are expiring leaving invalid references

Needs a decent amount of testing, not confirmed it actually helps with the issue.

Knocked up whilst looking at https://github.com/gerbera/gerbera/issues/2679

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -39,15 +39,15 @@
 #include "util/grb_net.h"
 #include "util/upnp_quirks.h"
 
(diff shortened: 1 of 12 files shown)


whyman:
Still crashes when built as a docker image, so strange.

KarlStraussberger:
What about just removing the `&` at the `shared_ptr` arguments so we get proper reference handling. 

Also quirks might be passed as `shared_ptr`.

Sounds reasonable that the context of the `ActionRequest` is dropped, so the objects go away because the reference counter does not _see_ our calls.

PR #2802, opened 2023-02-14
Update install-lastfm.sh

wareck:
Hello
I think there is a space char make some issue.
(ubuntu build, with this fix no problem)

best regards
Olivier


KarlStraussberger:
I don't think this is a problem. Because `. script` loads the content (esp. variables) into the current environment.

If this is a problem all other script have the same issue. You should not activate lastfm unless you've got account. It's partially unmaintained from our side.

PR #2803, opened 2023-02-15
Keep our unique_ptrs for the xml in scope until we are finished with …

whyman:
…them

Otherwise the temporaries get cleaned up and our node refs are invalid

Fixes: https://github.com/gerbera/gerbera/issues/2679


whyman:
Confirmed to fix the crash under docker with upnptool

KarlStraussberger:
Finally that one is fixed.

Worth looking out for similar code.

PR #2807, opened 2023-02-28
Greatest Hits Workaround (Issue #1440)

gobater:
Fix for Issue #1440:
This creates a unique path in the tree when importing a musicAlbum (upnpClass "object.container.album.musicAlbum"). The path consist of AlbumTitle + AlbumArtist (alt. Artist) + Release Year (optional)

This avoids merging several Albums with the same name. Two different use cases:
- 2 or more artists having an album with the same name (e.g. Greatest Hits) won't be merged under "Albums"
- 2 different albums (different release year) of the same artist having the same name (e.g. Best Of) wont' be merged under the Artist.

image


gobater:
Sorry for the "push-force"... I forgot to clean some debugging stuff

KarlStraussberger:
Thank you for supporting in finding a solution.

I think this piece of code is better located in the addAudio JavaScript / Builtin layout code to avoid side effects of changes in this server core function. The corresponding change in JS would even be simpler because most of the properties are already acquired.

gobater:
I'm not deep enough in Gerbera to know about the risks of adding the change in "content_manager", but to be honest, I don't think that either the builtin-layout or the JS code would be better. What is more, risks get mitigated with good unitests ;-)

This shall be (in my opinion) a core functionality of gerbera (since the albums have different artists or release dates are different albums) that shall be used for all layouts. Who would like not to have this feature to distinguish what it is (and shall be) different? 

Additionally, I don't see a possibility to change the "tree" that "content_manager" uses without messing up the "item title", since the tree is created based on it. So we would end up with your suggestion in the issue, which basically will add the Artist to all albums (and I'm quite sure that is not something everybody is going to like)

To be honest, I don't like much the ugly "if-else" I've implemented. A much better (and nicer) alternative would be to subclass CdsContainer for each (or some) upnp container class and add a polymorphic method "getUniqueId" which could be used by "addContainerTreeMethod" (instead of getTitle) to build the "tree" but currently, the class CdsContainer is final and making such a change would require much more work. I would like to avoid that work if it has not been discussed/agreed before with all maintainers having a veto about this

KarlStraussberger:
In general I like adding more object oriented aspects.

In this case it is the idea of layouts in Gerbera to define these things. I drafted up a small change in common.js which relies on a small config property to be set (current users can switch off the feature)
[unique-album.txt](https://github.com/gerbera/gerbera/files/10880427/unique-album.txt)

```
<virtual-layout type="js">
  <import-script ...>
  <script-options>
     <script-options name="album-mode" value="unique"/>
  </script-options>
</virtual-layout>
```

Does this solve your issue.

gobater:
I'm not sure that the patchset you attached with the change in the JS will do the trick. As I mentioned above, I suspect that changing the "Title" attribute will also affect how the container is displayed in Upnp Control Points (correct me if I'm wrong)
What you did is somehow what was proposed in the issue #1440 and was "rejected" by the reporter.
The problem here is the way that Gerbera creates its internal tree (based on object "titles") and not any "pseudo-random" unique Id, so I don't think this is a good solution. Tree-Path and Object Title do not need to be so tightly coupled

If you still want to solve everything in JS, a new attribute shall be added to the objects which shall be used as identifier for the tree and shall default to the object title, but title and tree-identifer are two different things, that shall be somehow, somehwere decoupled

I still do not see who would like this to be disabled (and therefore the need to be "switchable" in the config), but it's not a problem for me.

gobater:
@KarlStraussberger, what shall we do with this PR?

KarlStraussberger:
If time allows (the issue I cannot solve with code) I'll incorporate part of the ideas here in my current rework of the import process.
Leave it open for some more time.

gobater:
Fine for me keeping this open if there is a good reason for it.
Just think that keeping PRs of contributors open for a long time (there are PR's since more than 1.5 years open) does not feel very welcoming to participate in an open source the project.

KarlStraussberger:
Added some derived version in #2783

gobater:
> Added some derived version in #2783

Not sure if I've been able to understand correctly the relevant commit, but it looks promising

KarlStraussberger:
Integrate this in rework of server code

PR #2808, opened 2023-02-28
Database items sorting case insensitive

gobater:
Without this fix, the Album list will display:
- La Estación Seca
- La danza de Araña
- Lisergia

(E before d) which is not what most users would expect
Screenshot_20230228-232021

After this fix, the Album list will display:
- La danza de Araña
- La Estación Seca
- Lisergia
- 
image


KarlStraussberger:
Thank you for the PR,

I just did a quick background check and could not find a corresponding option in MySQL/MariaDB. I found some settings for the database itself, which would be preferable.

gobater:
Mnmm... I though that "NOCASE" was a standard collation. It's a long time since I used MySql/MariaDb for last time.

Creating a custom collation and modifying the database to sort certain colums based on that is definitively a much better (and time consuming) solution.
A good sorting collation takes care also about Accents, Umlauts and many other special characters (ä, á, à, a... and their uppercase variants), so I guess all that could be done but such a change implying a database update would need a discussion before starting to work on it without having quorum.

Sorting is in general a problem in Gerbera, it feels weird to me that non-standard "sub-containers" created by Gerbera import script (like "All Songs") are not fixed "on-top" and also sorted alphabetically so that having an album like "Iron Maiden's A Matter of Life and Death" (or something else starting with a numeric character) would appear before "All Songs"... I would also love to work on that by introducing an additional sorting criteria e.g. the upnp class (with higher priority than the title). Again, that is definitely worth a debate to challenge the idea and discuss about a solution.

KarlStraussberger:
That's a really good point. So we should look for database configuration options and some configuration extensions that allow the user to set the "All.." labels.

KarlStraussberger:
> That's a really good point. So we should look for database configuration options and some configuration extensions that allow the user to set the "All.." labels.

This is also part of the layout implementation but can be taken to configuraiton so the user does not have to change JavaScript code!

gobater:
I think that enabling/disabling the "all... " containers could (should?) be done with a option in the XML file. It's good that gerbera has the possibility to customize everything via custom layouts in JS, but that would not solve the sorting issue.
I cannot Imagine who wants to have those "All Songs" container but does not want to have them either in the beginning or end of the list (mixed with the albums, looks weird).

If there is "quorum" I can try to implement a new object type, that when used (either in JS or C++) takes higher (or lower) precedence in sorting.

For the upper-case lower-case topic I can have a make a draft and set default column collations. Is there anyone else that has a voice on this topic?

gobater:
Fine, I will try to handle the things in two different PR's ("All xxx" containers somewhere else vs collation here).

For the collation, we can make it extremely complex, so my suggestion would be to follow an incremental approach, since collations might be different depending on the locale.

> In a case-insensitive collation, "Mueller" and "mueller" would be considered the same name.
> In a case-sensitive collation, "Mueller" and "mueller" would be considered different names.
> In a German collation, "Mueller" and "Müller" would be considered the same name.
> In an English collation, "Mueller" and "Müller" would be considered different names.
> In a collation that ignores padding, "Mueller" and "Mueller " would be considered the same name.
> In a no-pad collation, "Mueller" and "Mueller " would be considered different names.
(https://mariadb.com/docs/skysql/sql/features/localization/collations/)

MariaDb comes already with support for different collations (we could think about adding a configuration parameter in the XML): https://mariadb.com/docs/skysql/ref/mdb/collations/. However, making the collation configurable makes it also more difficult, since we might need to reindex the table when changing the collation.

SQLite only provides ascii case-insensitive collation (no unicode support) so an external collation function shall be supplied. For that, we shall evaluate and decide which library we want to use (miniutf, libicu, ...?)

So, I will start with a case insensitive collation (which is already much better than no collation at all) while we discuss the unicode topics

gobater:
The changes add a default collation for the column "dc_title" of table "mt_cds_object" for case insensitive sorting (in sqlite3) and fr utf8 collation (in mysql).

Additionally, an additional (higher priority) criteria for sorting containers have been added in order to show all base/general containers (upnp_class 'object.container' e.g. 'All Songs') before specialized (e.g. upnp_class 'object.container.album.musicAlbum')

src/database/sql_database.cc
```diff
@@ -973,7 +973,10 @@ std::vector<std::shared_ptr<CdsObject>> SQLDatabase::browse(BrowseParam& param)
             where.push_back(std::move(zero));
         } else if (getContainers && !getItems) {
             where.push_back(fmt::format("{} = {:d}", browseColumnMapper->mapQuoted(BrowseCol::ObjectType), OBJECT_TYPE_CONTAINER));
-            orderBy = fmt::format(" ORDER BY {}", orderByCode());
+            // Sorting by UpnpClass will avoid mixing different types of containers
+            // "Special" containers like "All Songs" (which are of upnp_class 'object.container') will be displayed before
+            // albums (which are of upnp_class 'object.container.album.musicAlbum')
+            orderBy = fmt::format(" ORDER BY {}, {}", browseColumnMapper->mapQuoted(BrowseCol::UpnpClass), orderByCode());
```
KarlStraussberger, line 979: What about adding this in the else branch as well?

gobater: done

src/database/mysql/mysql-upgrade.xml
```diff
@@ -134,4 +134,7 @@
         <script>ALTER TABLE `mt_autoscan` ADD `ct_image` varchar(255) NOT NULL default 'object.container.album.photoAlbum'</script>
         <script>ALTER TABLE `mt_autoscan` ADD `ct_video` varchar(255) NOT NULL default 'object.container'</script>
     </version>
+    <version number="21" remark="set default column colation">
+        <script>ALTER TABLE "mt_cds_object" MODIFY "dc_title" varchar(255) default NULL COLLATE utf8_general_ci</script>
```
KarlStraussberger, line 138: MySql needs backticks
```
ALTER TABLE `mt_cds_object` MODIFY `dc_title` varchar(255) default NULL COLLATE utf8_general_ci;
```

KarlStraussberger (approved):

gobater:
PR rebased, cleaned-up and forced push...

KarlStraussberger:
Thank you!

PR #2848, opened 2023-07-23
Add artist chronology container

gobater:
Add an artist chronology container (similar to album container, but sorted by date instead of name)

### DIFF
--- a/scripts/js/audio.js
+++ b/scripts/js/audio.js
@@ -457,7 +457,7 @@ function addAudioInitial(obj, cont, rootPath, containerType) {
     if (boxSetup['Audio/artistChronology'].enabled) {
         chain.album.searchable = false;
         chain.artist.searchable = false;
-        chain.album.title = date + " - " + album
+        chain.album.title = date + " - " + album;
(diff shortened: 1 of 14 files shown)


gobater:
@KarlStraussberger , before I fix the tests, I would like to have your opinion on this...

KarlStraussberger:
> @KarlStraussberger , before I fix the tests, I would like to have your opinion on this...

It's absolutely fine. Go ahead.

gobater:
@KarlStraussberger , can you give me a hand fixing the broken test? It's passing in my machine :-( and I would expect it to pass in the CI since mock-config-minimal.xml seems correct to me.
Thanks

KarlStraussberger:
The one missing is in `fixtures/mock-config-all.xml`

KarlStraussberger:
There are several combinations of #defines that we test, hence all the xml-files

KarlStraussberger:
Sorry it's `fixtures/mock-example-all.xml`

gobater:
Oh! My bad! I didn't realize that my work was not in synch

PR #2865, opened 2023-08-25
Fix spelling errors reported by lintian

snd1:


KarlStraussberger:
Thank you for fixing theses really old typos

PR #2866, opened 2023-08-25
Fix troff warning

snd1:
In gerbera.1 config/gerbera is an "undefined macro".

Warning message from the current 1.12.1 release in debian:
```shell
troff:<standard input>:90: warning: macro 'config/gerbera' not defined [usr/share/man/man1/gerbera.1.gz:1]
``` 
This change fixes the warning.


KarlStraussberger:
Thank you for your contribution

PR #2868, opened 2023-08-25
Add permissions on HV transcoding devices in docker container

m-larin:
Gerbera start in docker from unpermitted user. When call ffmpeg with HW transcoding options rase error about the unavailability of the device. This commit fixes the error.


KarlStraussberger:
Instead we might add gerbera user to the group with access to the devices.

m-larin:
Redone according to your comments. Now the gerbera user is added to the video group, and this group is given rights to the device video

KarlStraussberger:
Thank you for your contribution.

PR #2899, opened 2023-11-16
Add #include <algorithm> to fix building with gcc 14

kostadinsh:
With gcc 14 some C++ Standard Library headers have been changed to no longer include other headers that were used internally by the library. In gerbera's case it is the `<algorithm>` header.

[Downstream gentoo bug](https://bugs.gentoo.org/917136)

[GCC 14 porting guide](https://gcc.gnu.org/gcc-14/porting_to.html#header-dep-changes)

### DIFF
--- a/src/cds/cds_objects.h
+++ b/src/cds/cds_objects.h
@@ -34,6 +34,7 @@
 #ifndef __CDS_OBJECTS_H__
 #define __CDS_OBJECTS_H__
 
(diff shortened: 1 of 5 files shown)


KarlStraussberger:
Thank you for supplying this.

kostadinsh:
> Thank you for supplying this.

You are welcome :-)

PR #2902, opened 2023-11-21
Doc: Compile libupnp --disable-blocking-tcp-connections

cweiske:
.. otherwise the cmake will warn about that.


KarlStraussberger:
Thank you

Issue #2903, opened 2023-11-21
Debian 12: pkg-config dependency required


cweiske:
Compiling on Debian 12 (bookworm) failed because `pkg-config` was not installed by default. It failed with
```
-- Could NOT find Systemd (missing: SYSTEMD_FOUND) 
CMake Error at CMakeLists.txt:597 (message):
  Systemd not found
```

With `apt install pkg-config`, systemd was found.

The full dependency installation command I needed:
```
apt -y install  duktape-dev ffmpeg libavcodec-dev libavdevice-dev libavfilter-dev libavformat-dev libavutil-dev libcurl4-openssl-dev libexif-dev libmagic-dev libmariadb-dev libmatroska-dev libpostproc-dev libpugixml-dev libspdlog-dev libsqlite3-dev libssl-dev libswresample-dev libswscale-dev libsystemd-dev libtag1-dev pkg-config uuid-dev 
```

PR #2904, opened 2023-11-21
Document dependency installation on Debian 12

cweiske:
Resolves: https://github.com/gerbera/gerbera/issues/2903


KarlStraussberger:
Can you take a look at the [build_dep](https://github.com/gerbera/gerbera/blob/master/scripts/debian/build-deb.sh) also?

cweiske:
pkg-config is already mentioned there.
I think the whole section is useless, and the build dep install script must be mentioned instead.

cweiske:
I've updated the patch to simply mention the script.

Issue #2913, opened 2023-12-17
Line number missing for javascript errors


cweiske:
I updated Gerbera from 1.12.1 to current git master (bac67efe18c384255f15d2526623983e71b2f8f6), had a strange bug with my kitchen radio, deleted all database tables and started gerbera. Now I get javascript errors all over the place:

> error: Failed to execute script /etc/gerbera/import-cweiske.js: TypeError: cannot read property 'res' of undefined
> error: [/usr/local/src/gerbera/gerbera-git/src/content/scripting/script.cc:430] void Script::_execute() Error: Script: failed to execute script

Unfortunately it does not tell me in which line the error occurs - this would be very helpful for debugging.
The stack trace would also help to see nesting.

https://duktape.org/api#duk_pcall

cweiske:
Replacing `duk_safe_to_string` with `duk_safe_to_stacktrace` gives better errors, and falls back to `duk_safe_to_string` if there is no stacktrace.

Now I get:
```
error: Failed! to execute script /etc/gerbera/import-cweiske.js: TypeError: cannot read property 'res' of undefined
     at [anon] (duk_hobject_props.c:2621) internal
     at addImage (/usr/local/share/gerbera/js/common.js:998)
     at global (/etc/gerbera/import-cweiske.js:328) preventsyield
error: [/usr/local/src/gerbera/gerbera-git/src/content/scripting/script.cc:430] void Script::_execute() Error: Script: failed to execute script
```

KarlStraussberger:
The fix is great.

There have been vast changes to the scripting mechanism. You may have to create a wrapper like https://github.com/gerbera/gerbera/blob/master/scripts%2Fjs%2Fimport.js

PR #2914, opened 2023-12-17
Scripting: Log stacktrace on errors

cweiske:
Resolves: https://github.com/gerbera/gerbera/issues/2913

### DIFF
--- a/src/content/scripting/script.cc
+++ b/src/content/scripting/script.cc
@@ -426,7 +426,7 @@ void Script::load(const fs::path& scriptPath)
 void Script::_execute()
 {
     if (duk_pcall(ctx, 0) != DUK_EXEC_SUCCESS) {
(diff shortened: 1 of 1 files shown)


cweiske:
@KarlStraussberger Build for ubuntu 18.04 failed: https://github.com/gerbera/gerbera/actions/runs/7237253434/job/19716682780
That probably means that this function is not available in their old duktape version.

KarlStraussberger:
That means we have to add some `#ifdef`

cweiske:
I don't have enought C knowledge to do that.
All I know is that `[duk_safe_to_stacktrace](https://duktape.org/api#duk_safe_to_stacktrace)` was added in duktape version 2.4.0: https://duktape.org/api#taglist-2.4.0

There is a `DUK_VERSION` constant, but I don't know how to compare versions in `#if`: https://duktape.org/guide#versioning.4

KarlStraussberger:
See #2921

PR #2937, opened 2023-12-28
Clients: Add detection for FSL radios

KarlStraussberger:


cweiske:
thank you!

cweiske:
Just tested; works fine.

PR #2959, opened 2024-01-11
CMake: Use presets

KarlStraussberger:
Requires at least cmake 3.19 but simplifies building and adding new options.

Also enables EXIV2 by default now.


whyman:
Nice!

PR #2975, opened 2024-01-20
some fixes

neheb:


src/upnp/quirks.cc
```diff
@@ -55,7 +55,7 @@ long Quirks::checkFlags(long flags) const
 
 bool Quirks::hasFlag(long flag) const
 {
-    return pClientInfo && (pClientInfo->flags & flag) == flag;
+    return pClientInfo && (static_cast<long>(pClientInfo->flags) & flag) == flag;
```
KarlStraussberger, line 58: There is another such line in `checkFlags`. But I prefer to replace the long by `QuirkFlags`, which requires changes in `ConfigOption` as well.
As I'm already in the middle of these changes, you can remove that change.

src/util/mt_inotify.cc
```diff
@@ -127,7 +127,7 @@ struct inotify_event* Inotify::nextEvent()
     // first_byte is index into event buffer
     if (firstByte != 0
         && firstByte <= int(bytes - sizeof(struct inotify_event))) {
-        ret = reinterpret_cast<struct inotify_event*>(reinterpret_cast<char*>(event.data()) + firstByte);
+        std::memcpy(ret, reinterpret_cast<char*>(event.data()) + firstByte, sizeof(inotify_event));
```
KarlStraussberger, line 130: There is another `copy_n` in the code below. We should only use one of `copy_n` or `memcpy`

src/metadata/taglib_handler.cc
```diff
@@ -419,7 +419,7 @@ std::unique_ptr<IOHandler> TagLibHandler::serveContent(const std::shared_ptr<Cds
         if (!f || !f->isValid() || !f->tag())
             throw_std_runtime_error("Could not open file {} as OGG", itemLocation);
 
-        const TagLib::List<TagLib::FLAC::Picture*> picList = reinterpret_cast<TagLib::Ogg::XiphComment*>(f->tag())->pictureList();
+        const auto picList = static_cast<TagLib::Ogg::XiphComment*>(f->tag())->pictureList();
```
KarlStraussberger, line 422: From the spec `static_cast` produces extra code to align the pointers. Do you see the need for this.

KarlStraussberger (changes requested):

src/upnp/quirks.cc
neheb: Removed.

src/util/mt_inotify.cc
neheb: Probably the latter since it's faster (copy_n uses memmove).

src/metadata/taglib_handler.cc
neheb: Removed.

neheb: So clang-tidy actually warns on this: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-static-cast-downcast.html

looks like dynamic_cast is the proper thing to use here.

PR #2982, opened 2024-01-25
clang-tidy and cppcheck fixes

neheb:


src/upnp/headers.cc
```diff
@@ -82,7 +82,7 @@ std::pair<std::string, std::string> Headers::parseHeader(const std::string& head
     return { first, second };
 }
 
-void Headers::writeHeaders(UpnpFileInfo* fileInfo) const
+void Headers::writeHeaders(const UpnpFileInfo* fileInfo) const
 {
 #if defined(USING_NPUPNP)
     std::copy(headers.begin(), headers.end(), std::back_inserter(fileInfo->response_headers));
```
KarlStraussberger, line 88: The change breaks NPUPNP build

KarlStraussberger (changes requested):

src/upnp/headers.cc
neheb: Yeah, cppcheck only warned on the read function. I was overzealous. Will fix.

neheb: Fixed.

Issue #2993, opened 2024-02-01
firewall advice needed in documentation


storrence97:
After I recently converted my Gerbera host from Ubuntu to Fedora 39 Server, Gerbera was not discoverable on my LAN because 
the default firewall settings in Fedora Server are much more restrictive than in Fedora Workstation or in the Ubuntu Server that I previously ran. I found Gerbera's Installation and Troubleshooting docs unhelpful with regard to needed firewall settings. 

I have successfully solved my own problem and would like to suggest addition of some information to the Gerbera docs. 

The specific solution I implemented had the following steps:
- To ensure that my firewall configuration would be correct, while minimizing open ports I added `<interface>eno1</interface>` and `<port>49152</port>` in `/etc/gerbera/config.xml`.
- Open port `49152/tcp` and service `ssdp`.  Service ssdp implies port `1900/udp` and multicast to `239.255.255.250`

Since Fedora uses firewalld, the specific firewall commands were
```shell
sudo firewall-cmd --permanent --add-service=ssdp
sudo firewall-cmd --permanent --add-port=49152/tcp
``` 

I suggest the following additions to Gerbera's documentation:
- Add a complete itemization of the ports/services that must be open in the firewall, probably in the Networking section of the _Troubleshooting_ page. Since the firewall documentation is pretty opaque, it would probably help many people to include the specific configuration commands.
- Add a note to the Fedora section of the _Install Gerbera_ page pointing out that the Server edition of Fedora will require additional firewall configuration. This note could probably refer to the updated networking section of the _Troubleshooting_ page.

KarlStraussberger:
Thank you for your comments which are really appreciated, because most of us are running their system since 5, 10 or even 15 years in the same environment.

You may even speed up the process by preparing your ideas in a simple text document and attach it here or fork gerbera, make your changes and create a pull request.

That's how this project works: support by everybody with ideas and time to bring the online. :cat:

PR #2995, opened 2024-02-03
Expand firewall advice in docs (#2993)

storrence97:
Add a new Firewall section to the Troubleshooting page.

Expand the Port section of the Troubleshooting page to complement the new Firewall section.

Add info about Fedora Server edition on the Installing Gerbera page. The existing Fedora section was ok for the Workstation edition. The added info should save a user a bunch of troubleshooting if they install on the Server edition.


KarlStraussberger:
That's great. Thank you very much.

PR #2996, opened 2024-02-06
Docker image usability

mzuther:
Hi!

This is my first contribution to Gerbera. It aims at making the Gerbera Docker image more customizable for users who do not want to write their own Dockerfiles:

* simplify adding custom JavaScript code
  * add directory `/mnt/customization/js` for custom JavaScript files
  * add this directory to `config.xml` on creation
* clarify that content directory will be "mounted" to a directory on the host
  * add directory `/content`  to `/mnt/content` and use that by default
  * link `/content` to `/mnt/content` for backward compatibility
* ease customization of `config.xml` when it is created
  * add directory `/mnt/customization/shell`
  * call script `/mnt/customization/shell/gerbera_config.sh` (if existing)
    whenever `config.xml` is created
* update documentation

I think that these modifications do not add much overhead, while making the life of "normal" users of the Docker image much easier.  But I am willing to discuss any of the modifications, of course.

Thanks for reviving MediaTomb, an application I always liked, and which has been improved so much by you!

Best,
Martin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,7 +101,7 @@ RUN addgroup -S gerbera 2>/dev/null && \
     adduser -S -D -H -h /var/run/gerbera -s /sbin/nologin -G gerbera -g gerbera gerbera 2>/dev/null && \
     addgroup gerbera video && \
     mkdir /var/run/gerbera/ && chmod 2775 /var/run/gerbera/ && \
-    mkdir /content && chmod 777 /content
+    mkdir /content && chmod 777 /content && ln -s /content /mnt/content
 
 EXPOSE 49494
 EXPOSE 1900/udp
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,7 +101,8 @@ RUN addgroup -S gerbera 2>/dev/null && \
     adduser -S -D -H -h /var/run/gerbera -s /sbin/nologin -G gerbera -g gerbera gerbera 2>/dev/null && \
     addgroup gerbera video && \
     mkdir /var/run/gerbera/ && chmod 2775 /var/run/gerbera/ && \
-    mkdir /content && chmod 777 /content && ln -s /content /mnt/content
+    mkdir /content && chmod 777 /content && ln -s /content /mnt/content && \
+    mkdir -p /mnt/customization/js && chmod -R 777 /mnt/customization
 
 EXPOSE 49494
 EXPOSE 1900/udp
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -101,7 +101,9 @@ RUN addgroup -S gerbera 2>/dev/null && \
     adduser -S -D -H -h /var/run/gerbera -s /sbin/nologin -G gerbera -g gerbera gerbera 2>/dev/null && \
     addgroup gerbera video && \
     mkdir /var/run/gerbera/ && chmod 2775 /var/run/gerbera/ && \
-    mkdir /content && chmod 777 /content
+    mkdir /content && chmod 777 /content && ln -s /content /mnt/content && \
```
KarlStraussberger, line 104: From your description, I'd expect `mkdir /mnt/content` here.

KarlStraussberger (changes requested):
Could you also update `Dockerfile.debug` for consistency.

Thank you for contributing!

Dockerfile
mzuther: Sorry, I might not have explained it well.

The idea is to allow users to keep their configuration. They can still mount a volume into `/content`, and Gerbera will read media from there. That is because `/mnt/content` is a symlink to `/content`.

If a new user comes along, however, she can mount a volume into `/mnt/content`, which replaces the symlink by the volume.

In both cases, Gerbera will read from `/mnt/content`. The difference is that this "directory" is either a _symlink_ to a volume, or a volume. I have tested it and it works like a charm.

Did I make more sense this time?

mzuther: I have updated `Dockerfile.debug` as well - it had escaped my notice.

KarlStraussberger: > Sorry, I might not have explained it well.
> 
> The idea is to allow users to keep their configuration. They can still mount a volume into `/content`, and Gerbera will read media from there. That is because `/mnt/content` is a symlink to `/content`.
> 
> If a new user comes along, however, she can mount a volume into `/mnt/content`, which replaces the symlink by the volume.
> 
> In both cases, Gerbera will read from `/mnt/content`. The difference is that this "directory" is either a _symlink_ to a volume, or a volume. I have tested it and it works like a charm.
> 
> Did I make more sense this time?

But there is no autoscan entry for /mnt/content

KarlStraussberger: > I have updated `Dockerfile.debug` as well - it had escaped my notice.
> 
> _Other than the previous changes I have not tested this change as `scripts/install-googletest.sh` quit with a server error._

That is some caching issue, goes away if you run the build again.

mzuther: > But there is no autoscan entry for /mnt/content

You are right, I must have missed it when I partitioned my changes into small logical commits. Weird that my tests succeeded, though.

This issue is now fixed, and I have taken the opportunity to improve the readability of the `sed` invocations.

PR #3007, opened 2024-02-26
matroska_handler: update API calls to work with libebml/libmatroska 2.0

robUx4:
Some API have been changed internally but there are some backward compatible macros to make the porting from one version to the other seamless.

All the changes should compile with libmatroska 1.0.0 and above.


KarlStraussberger:
Great!

Could you please update `scripts/versions.sh` as well to run the build with the new lib. Also there is the `README.md`.

Thank you

robUx4:
The 2.0 are not out yet. It's still a few weeks/months away. I just wanted to make sure there's no code out there that would break with the new library. For now it just means the code is ready with the changes we made so far.

KarlStraussberger:
That's really great, so I will create another PR (once I'm back home) to prepare for the moment.

PR #3017, opened 2024-03-14
Update Conan to v2

KarlStraussberger:


KarlStraussberger:
Exiv2/0.28.1 recipe seems to be broken with missing inih libraries, so I leave it off for the near future.

PR #3019, opened 2024-03-16
Replacing "test main" by gmock_main

gobater:


gobater:
@KarlStraussberger , could you help understanding why the CI is failing? gmock_main is a target available with vanilla googletest.

KarlStraussberger:
I think it's got to be `GMock::gmock_main`. It's in FindGMock.cmake

gobater:
> I think it's got to be `GMock::gmock_main`. It's in FindGMock.cmake

Weird... in the "official" GTestTargets.cmake (supplied with googletest) and available by default in my development Docker container (which is based on https://github.com/gerbera/gerbera/blob/master/Dockerfile) its "GTest::gmock_main" (not to add it works on my machine :-) ).

The problem probably is in "tests\CMakelists.txt":

`
find_package(GTest REQUIRED)

# CMake can't fix gmock support. https://gitlab.kitware.com/cmake/cmake/-/issues/17365
# So for non-conan build we need to find it separately.
if (NOT "${GTest_LIBRARY_LIST}" MATCHES "gmock")
    find_package(GMock REQUIRED)
    set_property(TARGET GTest::GTest APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${GMOCK_LIBRARIES})
endif()
`
If I replace GTest::gmock_main with GMock::gmock_main it does not work in my container (probably will work in the CI). The question is, why is the CI build? It seems to use an older version of cmake or googletest...

gobater:
So... it's a cmake version issue... see:
https://cmake.org/cmake/help/latest/module/FindGTest.html

gobater:
OK, there seems to be something with the conan build. I can also see some warnings about using deprecated functions in the log...

KarlStraussberger:
I already started migrating to conan2, but it's not as straighforward as it should be.

gobater:
No worries. We can leave this PR open until then...

KarlStraussberger:
Now you can try your changes. Conan2 was really a stubborn mule.

gobater:
@KarlStraussberger , I'm not an expert in conan. It seems that the installed gtest/gmock package is built in Release, while in the CI Job we are building in Debug so it cannot link to libgmockd... not sure how to solve and why it's working without the GTest::gmock_main target

KarlStraussberger:
We have `*/shared=True` in `conan/dev`. Maybe you can go with `gtest/*:shared=False` or even force `gtest/*:build_type=Debug`.

gobater:
I hope I find some time soon to work again on this :-)

KarlStraussberger:
Time is rare for all of us. Any hope of getting this done?

PR #3048, opened 2024-04-10
CI: Upload debs to pkg.gerbera.io

whyman:
Upload to s3 compatible storage with endpoint at pkg.gerbera.io

Will need to update docs once everything is confirmed to work OK.

Tested locally and its fine.

Script changes are to move the dep installs into the conditional, to avoid rebuilding everytime I was trying an upload.

### DIFF
--- a/.github/workflows/publish-deb.yml
+++ b/.github/workflows/publish-deb.yml
@@ -35,7 +35,11 @@ jobs:
           docker run \
             --platform=${{ matrix.arch }} \
             -e TZ=Etc/UTC \
-            -e ART_API_KEY=${{ secrets.ART_API_KEY }} \
+            -e DEB_UPLOAD_ACCESS_KEY_ID=${{ secrets.DEB_UPLOAD_ACCESS_KEY_ID }} \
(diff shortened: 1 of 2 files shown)


whyman:
Ah, this workflow is only run on version tags and master, so might have to merge and :crossed_fingers:

whyman:
If I magically find time might need to do something about the JS tests :scream_cat: 

Also love that riscv64 is in there now too! So awesome.

KarlStraussberger:
Great to silence the missing repo thing after months.

whyman:
Merging and lets see how it goes :pray:

KarlStraussberger:
> If I magically find time might need to do something about the JS tests 🙀
> 
> Also love that riscv64 is in there now too! So awesome.

Upgrade to boostrap5. Was looking in and out again 😦 Not enough time atm.

whyman:
:red_circle: https://github.com/gerbera/gerbera/actions/runs/8637552741/job/23679977293 :fire:

KarlStraussberger:
Needs some quotes?

whyman:
Yeah quotes would have solved it I think, but decided to pass it via envs instead https://github.com/gerbera/gerbera/pull/3049

PR #3060, opened 2024-04-22
Adding consistent album artist support

TheMarstonConnell:
Follow up to #3057

Essentially, we check the album artist tag first, if it doesn't exist, we just use the first artist in the list.

### DIFF
--- a/scripts/js/common.js
+++ b/scripts/js/common.js
@@ -325,7 +325,7 @@ function addAudio(obj, cont, rootPath, containerType) {
     }
 
   
-    var albumArtist = artist[0];
+    var albumArtist = artist;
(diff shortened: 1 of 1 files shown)


scripts/js/common.js
```diff
@@ -324,6 +324,13 @@ function addAudio(obj, cont, rootPath, containerType) {
         desc = artist_full;
     }
 
+  
+    var albumArtist = artist[0];
```
KarlStraussberger, line 328: It has to be an array of strings. So either
`albumArtist = artist` or `albumArtist = [ artist[0] ]`

KarlStraussberger (changes requested):

KarlStraussberger (approved):

KarlStraussberger:
Thank you

PR #3070, opened 2024-05-01
Import: Add option case sensitive tags

KarlStraussberger:
If option is active the behaviour is as before.
Turning the option off handle every tag virtual layout as lowercase even if it is displayed uppercase

https://github.com/gerbera/gerbera/discussions/3059#discussioncomment-9191861

closes #3059


ReleaseNotes.md
```diff
@@ -1,5 +1,23 @@
 # Gerbera - UPnP AV Mediaserver.
 
+## HEAD
+
+This is a really small update that brings better logging for TagLib and a new repository for ubuntu and debian getting rid of jfrog.
```
whyman10x, line 5: And UI improvements :D

PR #3075, opened 2024-05-04
Fix build against fmt 10.2.2

Hains:
git/src/config/client_config.cc:234:35:
error: 'join' is not a member of 'fmt'
|   234 |     return fmt::format("{}", fmt::join(myFlags, " | "));

https://github.com/fmtlib/fmt/commit/50565f9853926501bd1c9bda07c6a98f4d940691


KarlStraussberger:
fmt 10.2.2 is not released yet. So we keep the PR open until this happen.

KarlStraussberger:
Version 11 is out now and required further changes. Thank you anyway.

PR #3076, opened 2024-05-04
Server: Handle all HTTP via virtualdir

whyman:
Previously we let libupnp server $WEBROOT directly, this was used for the web ui and the various XML definitions.

This change moves to serving those files via virtualdir instead. This gives a lot more control. e.g. We can stop serving the UI if is disabled in config, instead of implmenting that in the UI JS.

It also means we can serve (in the future) Qurik'd versions of our XML deinitions, which is a pretty interesting idea.


KarlStraussberger:
> Previously we let libupnp server $WEBROOT directly, this was used for the web ui and the various XML definitions.
> 
> This change moves to serving those files via virtualdir instead. This gives a lot more control. e.g. We can stop serving the UI if is disabled in config, instead of implmenting that in the UI JS.
> 
> It also means we can serve (in the future) Qurik'd versions of our XML deinitions, which is a pretty interesting idea.

I already played around with quirked definitions but got stuck with the server quirking itself. Maybe my local changes make sense on top of yours

KarlStraussberger:
Progress: `description.xml` does quirks. Will also include `cds.xml` to drop `X_` for regular clients.

PR #3078, opened 2024-05-04
Fix discovery

KarlStraussberger:
at least vlc and upnp tools ignore announced path


code8buster:
Wasn't just VLC, Samsung TVs too ;) Compiling now!

whyman:
It's actually pretty funny/ironic that it's mostly libupnp clients that are bad

PR #3082, opened 2024-05-13
ci: Bump taglib latest

whyman:


whyman:
Urgh, okay. Let me try another approach.

KarlStraussberger:
That's why I only have it locally.

whyman:
Damn dependencies!

PR #3089, opened 2024-05-19
Build: rework dependency handling

KarlStraussberger:
Load dependency from separate distro specific file


KarlStraussberger:
Can easily be extended to other distros.

Offers room for release and version specific adaption (e.g. arguments to dep scripts) if this becomes necessary.

KarlStraussberger:
Ffmpegthumbnailer release 2.2.2 is from 2019, so use commit from January.

PR #3094, opened 2024-05-26
Build: Add install scripts for OpenSuSE

KarlStraussberger:


KarlStraussberger:
My libexif seemed quite old so I added another install script with commit handling.

PR #3099, opened 2024-05-29
config2.xsd fixes

cweiske:
Maybe the CI jobs should run xmllint to make sure everything is valid.

### DIFF
--- a/config/config2.xsd
+++ b/config/config2.xsd
@@ -66,10 +66,12 @@
                 <xs:element ref="storage"/>
                 <xs:element ref="ip" minOccurs="0"/>
                 <xs:element ref="interface" minOccurs="0"/>
+                <xs:element ref="modelURL" minOccurs="0"/>
                 <xs:element ref="modelName" minOccurs="0"/>
                 <xs:element ref="modelNumber" minOccurs="0"/>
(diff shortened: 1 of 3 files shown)


test/config/fixtures/mock-example-all.xml
```diff
@@ -18,12 +18,10 @@
     <port>0</port>
     <ip></ip>
     <interface></interface>
-    <manufacturer>Gerbera Contributors</manufacturer>
     <manufacturerURL>https://gerbera.io/</manufacturerURL>
     <modelName>Gerbera</modelName>
     <modelDescription>Free UPnP AV MediaServer, GNU GPL</modelDescription>
     <modelNumber>42</modelNumber>
-    <modelURL></modelURL>
```
KarlStraussberger, line 26: You need to add them to the xsd as well.

KarlStraussberger (commented):
Really great work. Was obviously not in our focus.

test/config/fixtures/mock-example-all.xml
KarlStraussberger: They have never made it into tge documentation. Since **mediatomb** times

KarlStraussberger:
Merged with pleasure.

PR #3113, opened 2024-06-03
Build: Upload for focal

KarlStraussberger:


KarlStraussberger:
Added Ubuntu 24.04 as build target

PR #3117, opened 2024-06-04
Fix creation of mr_reg.xml

### LINKED ISSUE #3115: Gerbera not found by TechniSat DigitRadio 580
**Describe the bug**
The Gerbera UPnP server does not get discovered by a TechniSat DigitRadio 580 anymore after updating to the latest git version.

A `git bisect` told me the responsible commit is 2d5a0f9501d9f41e5f958577f5e933a2758e4ed6:
```
# git bisect bad
2d5a0f9501d9f41e5f958577f5e933a2758e4ed6 is the first bad commit
commit 2d5a0f9501d9f41e5f958577f5e933a2758e4ed6
Author: Ian Whyman <ian@gerbera.io>
Date:   Wed May 1 21:40:04 2024 +0100

    Server: Handle all HTTP via virtualdir
```

All the following commits up to fc9191543faacc37c8745faf6d7186b647990b19 expose this problem.

**To Reproduce**

0. Start Gerbera
1. Scan for UPnP servers on the DigitRadio
2. Loading spinner is shown for 10 minutes

**Expected behavior**
The UPnP server is found and shown on the radio.

 - OS: Debian 12
 - Gerbera Version: since v2.1.0-18
 - Source: self built

**Media client (please complete the following / remove not applicable:**
 - Device: TechniSat DigitRadio 580 (Frontier Smart / Frontier Silicon chipset)

KarlStraussberger:
closes #3115

### DIFF
--- a/scripts/debian/include-focal.sh
+++ b/scripts/debian/include-focal.sh
@@ -41,7 +41,7 @@ function install-deb-s3() {
   sudo apt-get install -y ruby gpg
   if [[ "${deb_arch:-}" == "armhf"  ]]; then
     sudo mkdir -pv /usr/local/share/ca-certificates
-    sudo echo"-----BEGIN CERTIFICATE-----\
+    sudo echo "-----BEGIN CERTIFICATE-----\
 MIIDXzCCAkegAwIBAgILBAAAAAABIVhTCKIwDQYJKoZIhvcNAQELBQAwTDEgMB4\
 GA1UECxMXR2xvYmFsU2lnbiBSb290IENBIC0gUjMxEzARBgNVBAoTCkdsb2JhbF\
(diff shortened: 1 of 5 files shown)


cweiske:
This patch does not help with #3115; the radio still does only fetch the description and the mr_reg.xml.

Wireshark told me that the response to `mr_reg.xml` still does not contain a `content-length` header.

`Last-modified` looks ok now, and the XML header is available (with encoding, while previously it was without encoding attribute. I don't think this is the problem.)
`Content-type` is `text/xml` now as it was when discovery worked.

KarlStraussberger:
I added the file size info to the result. Strange thing. Only difference now is the UTF-8 tag.

cweiske:
It works now again! Thanks for the quick patch.

The content-length header is very important. While the header block in HTTP is easily discoverable by two consecutive newlines, the end of the body is hard to detect since there is no dedicated "end of stream" marker.

Issue #789, opened 2020-04-12
Multisink logging


ytimenkov:
I'd like to know your opinion on this:

As a developer I orten like to have a compact output to the console and more verbose to a file.

With spdlog it's easy to do and I've made a poc for me.

However most difficult is to how to configure the whole thing.

Preferably I would like to see this in the config file, since it's the only place which allow for rich configuration without reinventing done syntax: we need a file name, level, whether to color or not the console output, potentially different loggers (i.e. opt in for verbose sql queries).

But...
1. There is a reference to xml schema in the file, should this be updated as well? (Nowadays things tend to be more relaxed, you don't do a formal validation anyways...).
2. Somewhat late reading of the config file and cumbersome way to locate it with all those home overrides...
3. Some interleaving with config reloading. Is it used? Should it reapply new settings?
4. Some options can be defined both in the config and then overridden on the command line. It is poorly documented (or I may be looking not hard enough) but I would like to minimize the number of them to just the config file (maybe home, but then it's unclear interaction if one is redefined in the file). Another possibility is to use home from the command line only as a default if none is specified in the file (opposite to the current behavior).

I was thinking it should be possible to move "locating" the home and config to separate routines, read logging configuration early and then pass parsed XML document to the `ConfigurationManager`.

Also `--debug` flag may need useful to debug parsing or do early logging, but c useful only written `WITH_DEBUG` and could be replaced with sptdlog's configuration via env variables.

If course, another option is to just add another option "developer-friendly-logging" 😂

ytimenkov:
And for service it would be probably beneficial to write to syslog...

gabime:
https://github.com/guangie88/spdlog_setup should be useful for this

ytimenkov:
> https://github.com/guangie88/spdlog_setup should be useful for this

Thanks for the tip. Two things which bother me:
1. It uses a separate configuration file (and format) which may be ok for development (especially if using this only conditionally), but as user I would rather have a single way to set things up.
2. Readme says that fmt and cpptoml libraries "are inlined as part of the include" which already caused troubles (see #691).

PR #3121, opened 2024-06-10
Add log sinks for rotating log files and syslog

KarlStraussberger:


whyman:
I wonder if we should just ship a log rotate config file instead?

KarlStraussberger:
Also see #789

KarlStraussberger:
> I wonder if we should just ship a log rotate config file instead?

Once it's all done...

KarlStraussberger:
Need some config options for the size, rotation count, syslog facility etc,

Issue #2986, opened 2024-01-28
Samsung tv not playing files


DJ-Key:
[gerbera.log](https://github.com/gerbera/gerbera/files/14077851/gerbera.log)
[config.txt](https://github.com/gerbera/gerbera/files/14077873/config.txt)
My Samsung Smart Series 7 55" (UE55RU7402UXXH) isn't playing any video files and always says
Error
An unexpected Error accured
Skip | Stop
I've attached my config and a generated log.

DJ-Key:
Installed it freshly on a Ubuntu 23.10 server
Gerbera version is 1.12.1

KarlStraussberger:
1.12.1 had some issues with Samsung devices which are fixed in 2.0.

I know that jfrog is not available now to get our current build.

DJ-Key:
The problem was also present in my old 2.0 installation.

DJ-Key:
Before my old installation auto,-updated itself from 1.12 to 1.12.1 or 2.0 everything worked fine.

DJ-Key:
In ubuntu 23.10 I don't need the jfrog source. It's in the ubuntu sources somehow.

KarlStraussberger:
In that case, we should try to circle the issue in 2.0

KarlStraussberger:
Can you share the respective log fro. 2.0

Please, keep in mind that there are config changes from 1.12 to 2.0. Best generate the new with `--create-config` and compare.

Smartsmurf:
My good old Samsung (B-series) stumbles over the ProductCap announcement when there is no real handling. So I commented this out:

`--- a/src/upnp_xml.cc
+++ b/src/upnp_xml.cc
@@ -265,8 +273,8 @@
     const auto deviceStringProperties = std::array {
         std::pair("deviceType", UPNP_DESC_DEVICE_TYPE),
         std::pair("presentationURL", presentationURL.empty() ? "/" : presentationURL.c_str()),
-        std::pair("sec:ProductCap", UPNP_DESC_PRODUCT_CAPS),
-        std::pair("sec:X_ProductCap", UPNP_DESC_PRODUCT_CAPS), // used by SAMSUNG
+ //       std::pair("sec:ProductCap", UPNP_DESC_PRODUCT_CAPS),
+ //       std::pair("sec:X_ProductCap", UPNP_DESC_PRODUCT_CAPS), // used by SAMSUNG
     };
     for (auto&& [tag, value] : deviceStringProperties) {
[...]

PR #3124, opened 2024-06-13
UPnP description: Allow dynamic capability values

KarlStraussberger:
The individual values can depend on client flags.


KarlStraussberger:
May help getting #2986 done

PR #3134, opened 2024-06-23
Fix crashes on freebsd 14.1

vovkasm:
These changes fix application crashes on the freebsd 14.1

1. I removed the use of std::prev with std::filesystem::path objects (see llvm/llvm-project#55521)
2. Fixed typo (at least it looks like typo) in import code that handles `.nomedia` files

After applying these patches I was able to successfully compile and run port of gerbera 2.1.0 on freebsd 14.1 on my NAS, and successfully imported my video folder with recursive flag.


KarlStraussberger:
Thank you.

PR #3171, opened 2024-08-01
Use full client info as requestCookie

KarlStraussberger:
Allows accessing hostname and userAgent in request handlers


KarlStraussberger:
Two more todos are gone

PR #3201, opened 2024-09-05
Config: Make UDN mandatory

### LINKED ISSUE #2891: Gerbera 1.12 not showing up on the network
Hi Guys, I don't really have any ideas how to troubleshoot or find out what is the problem. I updated  to Kubuntu 23.10 - it comes with gerbera 1.12 (instead of 1.1 in the previous distro versions).

The previous versions worked normally. But the 1.12 does not show up on the TV, nor in the `gupnp-universal-cp` tool.

If I run the previous version (1.1) in VirtualBox, it shows up normally on all other devices on the network.

Please help me with troubleshooting

### LINKED ISSUE #3200: Warn user if `udn` is unset
I have noticed Gerbera (probably I should say `libupnp`) does not send SSDP notifications if `udn` value in unset (eg `<udn/>`) in `config.xml`. Maybe it would make sense to detect the issue and display a warning on start?

KarlStraussberger:
closes #3200
closes #2891

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,11 +5,14 @@
 - Add build target for source documentation
 - Add configuration option to block clients
 - Add Splitter and Resize to items grid
+- Bump webpack from 5.93.0 to 5.94.0 in /gerbera-web
 - Clean up commented out code
 - Clean up test code
 - Clean up URL handling
 - Configuration of additional target mimetype properties for transcoding
+- Config: Make UDN mandatory
 - Config: Option to disable physical container types
+- ConfigSetup: Allow multiple iterations for UI
 - Drop Apple Trailer support
 - Extract image and video orientation to resource attributes
 - FIXME done: Remove fake resource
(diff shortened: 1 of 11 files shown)


whyman:
The issue here is that if distros ship a usb, you end up with possible duplicate udn on that same network.

Could we just generate a udn if it's not present?

KarlStraussberger:
> The issue here is that if distros ship a UDN in their default config, you end up with possible duplicate udn on that same network.
> 
> Could we just generate a udn if it's not present?

It has to be the same each time otherwise some clients will keep duplicate entries that are not browsable. I'm thinking of generating one and storing it in the database - once it's up. Config check takes place before database connection is established.

whyman:
> It has to be the same each time otherwise some clients will keep duplicate entries that are not browsable. I'm thinking of generating one and storing it in the database - once it's up. Config check takes place before database connection is established.

Yeah good idea.

KarlStraussberger:
Added the 3 lines in #3209

PR #3206, opened 2024-09-06
Fix config2.xsd

cweiske:
I got an error:

```
$ xmllint --noout --schema config/config2.xsd test/config/fixtures/mock-config-all.xml config/config2.xsd:1183: element choice: Schemas parser error :
 Element '{http://www.w3.org/2001/XMLSchema}all': The content is not valid.
 Expected is (annotation?, (annotation?, element*).
WXS schema config/config2.xsd failed to compile
```

Maybe github actions should automatically validate the xsd files so that such errors are found.

### DIFF
--- a/config/config2.xsd
+++ b/config/config2.xsd
@@ -1180,10 +1180,7 @@
     <xs:element name="profile">
         <xs:complexType>
             <xs:all>
-                <xs:choice minOccurs="1" maxOccurs="1">
-                        <xs:element name="mimetype" type="xs:string"/>
-                        <xs:element ref="mimetype"/>
-                </xs:choice>
+                <xs:element ref="mimetype"/>
                 <xs:element ref="dlna-profile" minOccurs="0"/>
                 <xs:element ref="accept-url" minOccurs="0"/>
                 <xs:element ref="first-resource" minOccurs="0"/>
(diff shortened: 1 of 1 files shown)


config/config2.xsd
```diff
@@ -1180,10 +1180,7 @@
     <xs:element name="profile">
         <xs:complexType>
             <xs:all>
-                <xs:choice minOccurs="1" maxOccurs="1">
-                        <xs:element name="mimetype" type="xs:string"/>
-                        <xs:element ref="mimetype"/>
-                </xs:choice>
+                <xs:element name="mimetype" type="xs:string" minOccurs="1" maxOccurs="1"/>
```
KarlStraussberger, line 1183: The intension was to allow the full subelement or the string here.

I'm not an expert in XSD maybe it needs another nesting.

cweiske:
So you want to allow this two variants for [profile](https://docs.gerbera.io/en/stable/config-transcode.html#profiles)?
```
<profile name="a">
    <mimetype>image/png</mimetype>
</profile>

<profile name="b">
    <mimetype value="image/png">
        <mime-property key="foo" resource="bar" metadata="baz"/>
    </mimetype>
</profile>
```
let's see how to solve this.

KarlStraussberger:
That's how it's implemented to be compatible with old files.

Here's the example https://docs.gerbera.io/en/latest/config-transcode.html#profiles

config/config2.xsd
cweiske: Solved, but there is no optimal solution :(

KarlStraussberger:
Great

Issue #3211, opened 2024-09-10
Crash on UPnP search with unclosed string


cweiske:
**Describe the bug**
When doing a UPnP search request with an unclosed string, gerbera crashes:
```
upnp:class = "object.container.album.musicAlbum" and upnp:lastPlaybackTime > "2024-09-01
```

(Closing double quote is missing)

(I tried to find out how to build a [container](https://docs.gerbera.io/en/stable/config-server.html#containers) that shows the last played albums).

**To Reproduce**

1. Start gerbera on server
2. Start `gupnp-av-cp` on desktop PC
3. Right-click gerbera server, select "search"
4. Input `upnp:class = "object.container.album.musicAlbum" and upnp:lastPlaybackTime > "2024-09-01` and press return
5. Gerbera crashes with `Main process exited, code=killed, status=11/SEGV`

**Expected behavior**
No crash

**Server (please complete the following information):**
 - OS: Debian 12
 - Gerbera Version: v2.2.0-52-g26ad170f
 - Source: self built

KarlStraussberger:
> For the 'fun' of it: alas confirmed with ...50... .

You can even try 1.1 and will find it there as well!

udippel:
I didn't try to be funny, neither cynic. Some 15 years ago, I used to be quite active in sanitization of user input. It has and will probably remain a step-child, because it doesn't actually 'sell'. It serves no new functionalities, and the user gains seemingly nothing from it. Only, when it is (not) too late. 
Haven't been active for the last 12 years, I wonder if by now some libraries exist, through which one can run user input, to throw away generically wrong input? Escapes, odd numbers of quotes, silly SQL commands, and so forth.

cweiske:
Gerbera does not crash anymore, it logs a warning now:
```
2024-09-13 09:11:52 warning: [/usr/local/src/gerbera/gerbera-git/src/database/search_handler.cc:339] std::unique_ptr<ASTQuotedString> SearchParser::parseQuotedString() Error: Failed to parse search criteria - expecting a double-quote
```

KarlStraussberger:
That's intended. Thank you

udippel:
8 out of 10, I'd say. The client says: 
![image](https://github.com/user-attachments/assets/7c7a2f32-d2c1-4abd-9815-965a1b77a5eb)
The user is barely the wiser. Question: How difficult would it be to push the error text to the user, and giving out 
[...]

PR #3221, opened 2024-09-13
UPnP: Return error messages

KarlStraussberger:


cweiske:
This is part of #3211.

PR #3257, opened 2024-10-22
actions: clang-format 18

whyman:


KarlStraussberger:
Thank you for starting it. Finished it off now.

Issue #3313, opened 2024-12-19
config2.xsd is not valid: The QName value 'upnp-properties' does not resolve to a(n) type definition.


cweiske:
**Describe the bug**
Validating my config.xml file fails because the schema file is invalid for xmllint:
```
xmllint --noout --schema config/config2.xsd test/config/fixtures/mock-config-all.xml 
config/config2.xsd:449: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}album-properties', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-properties' does not resolve to a(n) type definition.
config/config2.xsd:450: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}artist-properties', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-properties' does not resolve to a(n) type definition.
config/config2.xsd:451: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}genre-properties', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-properties' does not resolve to a(n) type definition.
config/config2.xsd:452: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}playlist-properties', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-properties' does not resolve to a(n) type definition.
config/config2.xsd:453: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}title-properties', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-properties' does not resolve to a(n) type definition.
config/config2.xsd:454: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}resource-defaults', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-defaults' does not resolve to a(n) type definition.
config/config2.xsd:455: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}object-defaults', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-defaults' does not resolve to a(n) type definition.
config/config2.xsd:456: element element: Schemas parser error : element decl. '{http://mediatomb.cc/config/2}container-defaults', attribute 'type': The QName value '{http://mediatomb.cc/config/2}upnp-defaults' does not resolve to a(n) type definition.
WXS schema config/config2.xsd failed to compile
```

**Expected behavior**
[...]

PR #3314, opened 2024-12-19
Fix config2.xsd: The QName value upnp-properties does not resolve to …

cweiske:
…a(n) type definition

Resolves: https://github.com/gerbera/gerbera/issues/3313


KarlStraussberger:
Thank you

PR #3315, opened 2024-12-20
Compile option for human-oriented sorting

mishanja88:
Hello, @KarlStraussberger 
I've decided to contribute some human-sorting tweaks to `fs::directory_iterator` as a CMake option (disabled by default).
If enabled, iterator tries to order entries by digital prefix (or suffix), falling back to standard order on failure.

Maybe it is possible to customize further and add XML config node, and even JS sorting handler.
I will continue this research, some day.


KarlStraussberger:
I like the idea to improve the physical layout.

Instead of compiler switches gerbera has config options. See https://docs.gerbera.io/en/latest/dev-configuration.html for the guide to implement it.

Also you should add headers to your code.

mishanja88:
Ok, thank you, I'll improve this as you suggest!

KarlStraussberger:
How do you handle that a file was added (inotify) after the initial scan?

mishanja88:
> How do you handle that a file was added (inotify) after the initial scan?

@KarlStraussberger oh, seems like I don't handle it at all. I would like to get it done right, and seems like it's something around the `addCdsObject` call. My chance would be that sorting happens automatically on the request of container (for web-view or as a DLNA request). I will test it more thoroughly, but seems like New Year's run-around is taking over :)

mishanja88:
@KarlStraussberger 
So many methods of path ordering I've found, this far:
- iterating with `fs::directory_iterator`
- creating and iterating over `std::map<std::string, ...>`
- iterating with `fs::path.begin()` `fs::path.end()`
- finally, some SQL ordering magic

With SQL magic, for now, I've managed to find a place to force-inject the sorting.
But it's not the right way, I know...

Actually, I'm starting to think that I went a little too deep.
Maybe custom sorting should appear as an option in "Directory tweaks" Web UI, and not all over the code.
Then, SQL result sorting should do the trick.
So PC directory will be `1, 11, 12, 2, ...`, as in filesystem, but container layout would be configurable.

mishanja88:
@KarlStraussberger, maybe it would be sufficient to utilize "audiotrack number" property. Though it would be semantically incorrect.
So I'll look for the suggested property. There was some voodoo git rev-parse magic for that.

KarlStraussberger:
Audio tracknumber is no good idea because it may be overwritten by media metadata. You can also create new metadata properties while scanning the folder.

KarlStraussberger:
Replaced by #3429

PR #3327, opened 2024-12-31
Gerbera welcomes 2025

KarlStraussberger:


KarlStraussberger:
In fact, I prefer the giant patch. I tried the by-case approach in many commercial projects and never really succeeded. Even maintaining the Changelog is a big challenge 😸  Happy new out there

PR #3351, opened 2025-01-17
Remap metadata fields when parsing virtual-directories keys

### LINKED ISSUE #3350: Import merges albums with same name...again
**Describe the bug**
It looks like the issue in #1440 has resurfaced again. Albums with the same name but different artists are being merged into a single container.

I have virtual-directories setup with the artist metadata.
```xml
<import hidden-files="no" import-mode="grb">
  <virtual-directories>
    <key metadata="M_ARTIST"/>
  </virtual-directories>
</import>
```

In 2.3 this works as expected with the default JS layout. It's broken in 2.4.1. 

**To Reproduce**
0. Add  `<key metadata="M_ARTIST"/>` to virtual-directories
1. Start Gerbera
2. Add media with identical album names from separate artists.
3. See error

**Expected behavior**

 - OS: Docker on Debian 12
 - Gerbera Version: 2.4.1
 - Versions of libraries: N/A (Docker)
 - Source: Official Docker image

**Additional context**

Looking at the DB we can see that the artist is not being added to the `location` in 2.4.1.

Image

mill1000:
Close #3350 

Probably should have a test for this...

### DIFF
--- a/src/cds/cds_objects.h
+++ b/src/cds/cds_objects.h
@@ -218,7 +218,7 @@ public:
         auto it = std::find_if(metaData.begin(), metaData.end(), [=](auto&& md) { return md.first == field; });
         return it != metaData.end() ? it->second : std::string();
     }
-     /// \brief Query multivalue metadata.
+    /// \brief Query multivalue metadata.
     std::vector<std::string> getMetaGroup(const MetadataFields& key) const
     {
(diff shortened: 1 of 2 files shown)


KarlStraussberger:
Thank you for digging that up. Not sure why it was removed in the first place...

PR #3371, opened 2025-01-28
Build correct Autoscan Type

### LINKED ISSUE #3370: Autoscan parent not set correctly
**Describe the bug**

**To Reproduce**
0. If a directory starts with the same string as an autoscan directory it is shown as "autoscan by parent" in the filesystem view on web ui

**Expected behavior**
only show when real subDir.

up to 2.4.1-21

KarlStraussberger:
closes #3370

### DIFF
--- a/src/cds/cds_container.h
+++ b/src/cds/cds_container.h
@@ -43,7 +43,7 @@ protected:
     int childCount { -1 };
 
     /// \brief whether this container is an autoscan start point.
-    int autoscanType { OBJECT_AUTOSCAN_NONE };
+    AutoscanType autoscanType { AutoscanType::None };
 
 public:
     /// \brief Constructor, initializes default values for the flags and sets the object type.
@@ -75,10 +75,10 @@ public:
     int getChildCount() const { return childCount; }
 
     /// \brief returns whether this container is an autoscan start point.
-    int getAutoscanType() const { return autoscanType; }
+    AutoscanType getAutoscanType() const { return autoscanType; }
 
     /// \brief sets whether this container is an autoscan start point.
-    void setAutoscanType(int type) { autoscanType = type; }
+    void setAutoscanType(AutoscanType type) { autoscanType = type; }
 
     /// \brief Copies all object properties to another object.
     /// \param obj target object (clone)
--- a/src/cds/cds_enums.h
+++ b/src/cds/cds_enums.h
@@ -60,15 +60,17 @@ static constexpr bool IS_CDS_ITEM_EXTERNAL_URL(unsigned int type)
 #define OBJECT_FLAG_ONLINE_SERVICE 0x00000040u
 #define OBJECT_FLAG_OGG_THEORA 0x00000080u
 
-#define OBJECT_AUTOSCAN_NONE 0u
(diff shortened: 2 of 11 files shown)


michael-selig:
You might want to consider using std::filesystem::weakly_canonical() in isSubDir() (util/grb_fs.cc) instead of rtrimPath(). It does all that rtrimPath() does and more - it will convert legal but ugly paths like `/usr/./bin//../bin//` to `/usr/bin`.

michael-selig:
I just noticed that startswith() (in util/tools.h) also looks strange. Assuming it uses rfind() (which in itself is strange because the match is at the beginning of the string, not the end), the following will break:
`startswith("/home/test/home/gerbera", "/home");`
because it will match on the 2nd `/home`.
Why not use:
`return str.substr(0, prefix.size()) == prefix;`
Probably more efficient than either find() or rfind() too.

Of course it's OK if it uses starts_with()

KarlStraussberger:
> You might want to consider using std::filesystem::weakly_canonical() in isSubDir() (util/grb_fs.cc) instead of rtrimPath(). It does all that rtrimPath() does and more - it will convert legal but ugly paths like `/usr/./bin//../bin//` to `/usr/bin`. That means isSubDir() becomes something like:
> 
> ```
> bool isSubDir(const fs::path& path, const fs::path& parent)
> {
>     auto pathStr = fs::weakly_canonical(path).string();
>     auto chkStr = fs::weakly_canonical(parent).string() + "/";
>     return startswith(pathStr, chkStr);
> }
> 
> ```

Why no create a PR yourself.

michael-selig:
I assume you mean "Pull Request" rather than "Problem Report".
There are a few reasons why I didn't do a pull request:

1. The change was trivial and I saw that you were working on the same code anyway, so I thought you might as well do it if you thought it was appropriate.
2. It is possible that earlier in the code the parameters to isSubDir() may have been "canonicalised" (is that a word?) already, so maybe there was no point doing it again. It makes better sense to make sure that all file paths are in a canonical form when they are first input, rather than doing it possibly multiple times.  Because you know the code much better than me, you might know whether Gerbera already does this.
3. Last time when I did a pull request in an area you were also working on, it turned out that you had already done a fix at the same time. Plus your fix was better because you know Gerbera much better than I do.

KarlStraussberger:
> I assume you mean "Pull Request" rather than "Problem Report". There are a few reasons why I didn't do a pull request:
> 
>     1. The change was trivial and I saw that you were working on the same code anyway, so I thought you might as well do it if you thought it was appropriate.
> 
>     2. It is possible that earlier in the code the parameters to isSubDir() may have been "canonicalised" (is that a word?) already, so maybe there was no point doing it again. It makes better sense to make sure that all file paths are in a canonical form when they are first input, rather than doing it possibly multiple times.  Because you know the code much better than me, you might know whether Gerbera already does this.

That's a good point and should be valid for a directory scan. It's worth checking out some places where directories "come" in and "can8e" them.

>     3. Last time when I did a pull request in an area you were also working on, it turned out that you had already done a fix at the same time. Plus your fix was better because you know Gerbera much better than I do.

Sorry, I try to do the small things immediately so I don't forget them - a tribute to age and workload.

PR #3385, opened 2025-02-04
Fix management auf autoscans in list

### LINKED ISSUE #3383: error: Duplicate xxAutoscanDirectory entry
**Describe the bug**
If you edit an autoscan through the Web UI more than once, it will fail with this error.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create an autoscan item in the UI in the File System view
2. Refresh the Web page so that the orange autoscan button appears (a separate bug)
3. Click on the orange autoscan button, and push "Save". That works.
4. Repeat. You get the above error in the log each time you try to save, until gerbera is restarted.

**Additional context**
This comes from the Autoscan discussion #3367 where you replied to me:

> > Pretty sure this one is a bug: I sometimes get the message "error: Duplicate xxAutoscanDirectory entry" after clicking on an existing autoscan item (orange autoscan icon) and clicking "Save Autoscan". It seems to happen whether I make a change to the autoscan config or not, and only on the second and subsequent times I do this. If I do make a change, the change is not saved, though the scan seems to run.

> Sounds like some multi-threading condition issue.

It looks like a simple error in logic to me. The code appears to call EditHelper::add() in file `config/result/edit_helper.h` which in turn calls _add(). That function checks for an entry in a list, and if not present, adds it. If it is already present you get the "Duplicate" error. I didn't trace the logic further, so I don't really understand what this list is used for, but clearly the entry isn't being removed when it should be.

KarlStraussberger:
closes #3383

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,7 @@
 - Downgrade docker builder as well
 - Fix album art in builtin layout
 - Fix cors host without ipv6 address
+- Fix management auf autoscans in list
 - Fix UpnpMap logic
 - Focal: Freeze more aws-sdk versions
 - Focal: freeze ruby aws-sdk versions used for package upload
--- a/src/content/content_manager.cc
+++ b/src/content/content_manager.cc
@@ -137,7 +137,7 @@ void ContentManager::run()
             dir->setObjectID(ensurePathExistence(path));
         }
         try {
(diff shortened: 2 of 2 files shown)


michael-selig:
#3383 doesn't appear to be fixed

Issue #3394, opened 2025-02-07
Playlists may fail to import or update and generate an exception


jmwanderer:
**Describe the bug**
Inconsistent behavior is seen when adding or updating the m3u playlist under the autoscan directory. Sometimes the playlist entries do not appear and an error message is reported in the log:

error: [autoscan_inotify.cc:227] threadProc(): Inotify thread caught exception: [/build/src/database/sql_database.cc:1728] std::shared_ptr<CdsObject> SQLDatabase::createObjectFromRow(const std::string&, const std::unique_ptr<SQLRow>&): Error: tried to create object without at least one resource

On the initial add of the playlist m3u file, the error is reported and the playlist does not appear.
If the file is updated with the touch command, the file is imported and the playlist appears.
If the file is updated again, Inotify thread logs the exception and the playlist disappears.

**To Reproduce**
Start the gerbera server and add an autoscan directory data:
```
data/
    video/
       video1.MTS
        ...
   playlists/
```
Copy a small .m3u file that references 2 of the videos into the playlists directory

This typically fails and reports an error in the log.
Touch the .m3u file, and the playlist will be imported
After that is complete, touch it again, and the playlist will go away

**Expected behavior**
Playlists in autosync directories should reliably be imported into the database and appear in the UI.

**Server (please complete the following information):**
- Debian Trixie
- Gerbera 2.4.0 (reproduced on 2.4.1)

**Additional context**

I believe this is a simple fix:

*Root Cause*
The root cause appears to be the use of an async removeObject in inotify_handler.cc 

*Analysis*
When an existing playlist is updated, the debug log files show the following:

error: [autoscan_inotify.cc:227] threadProc(): Inotify thread caught exception: [/build/src/database/sql_database.cc:1728] std::shared_ptr<CdsObject> SQLDatabase::createObjectFromRow(const std::string&, const std::unique_ptr<SQLRow>&): Error: tried to create object without at least one resource

On the update of a playlist file, AutoscanInotify::threadProc takes the following actions:

InotifyHandler::doExistingFile
    - invokes ContentManager:removeObject with the default aysnc = true

InotifyHandlder::doNewFile
    - invokes ContentManager:addFile with the asSetting.async argument set to false
    - the doImport processs is started
    - eventually a SELECT is done for the mt_cds_object entry for the .m3u file
    - the async process has not yet deleted this, so the entry is found
[...]

Issue #3382, opened 2025-02-02
Inotify autoscan of playlist (m3u files)


tommyp-vr4:
I'm not sure if the problem I am having is related to the autoscans problem thread (#3367).  Sorry, that thread jumped a bit around too much for me.  

I have a folder of M3U playlist files.  The files are automatically created when I turn my computer on.  

If I use the UI to assign this standalone folder as "inotify", Gerbera _initially_ recognizes each m3u file and adds to the library.  Perfect.  If the m3u files are updated or replaced or new ones added, Gerbera drops the playlist mp3 entries from the library.  Gerbera should be updating the library with the new m3u files instead of dropping them.  

[The goal of my m3u files is random song playlists.]

**Server (please complete the following information):**
 - OS: [e.g. Debian Bookworm]
 - Gerbera Version: 2.4.0
 - Source: [e.g. gerbera's debian repository]

**Media client (please complete the following / remove not applicable:**
 - Device: Networked computer, Firefox (private mode)
 - Client app: Firefox & upplay
 - Browser: Firefox

EDIT - I ended up adding random numbers to the M3U filenames and things seem to work better.  Its a weird solution though.

jmwanderer:
I think this will be fixed with the change made for [ #3394](https://github.com/gerbera/gerbera/issues/3394)

It seems like the same problem.

tommyp-vr4:
The latest version V2.5 has fixed this issue.  Thanks all for fixing!!!!

PR #3395, opened 2025-02-07
Fix update of playlists

jmwanderer:
Playlist create and updates can fail due to a race of deleting the existing playlist and making a new one. Make the removeObject synchronous to fix the issue.

When a playlist file is updated, AutoscanInotify::threadProc takes the following actions:

InotifyHandler::doExistingFile
- invokes ContentManager:removeObject with the default aysnc = true

InotifyHandlder::doNewFile
- invokes ContentManager:addFile with the asSetting.async argument set to false
- the doImport processs is started
- eventually a SELECT is done for the mt_cds_object entry for the .m3u file
- the async process has not yet deleted this, so the entry is found
- soon after, the mt_cds_object entry for the .m3u file as well as the two V entries that reference the the .m3u file entry are deleted, causing the exception

The next time the playlist file is updated, no entries exist, so only doNewFile is called.
The entry for the playlist is not found, and the appropriate entries are created.

Using InotifyHandler::doExistingFile should run content->removeObject with async set to false will cause the previous instance to be removed before the new one starts to be created.

### DIFF
--- a/src/content/inotify/inotify_handler.cc
+++ b/src/content/inotify/inotify_handler.cc
@@ -220,7 +220,7 @@ int InotifyHandler::doExistingFile(const std::shared_ptr<Database>& database, co
             isDir = changedObject->isContainer();
         if ((!AUTOSCAN_IS_WRITTEN(mask) || importMode != ImportMode::Gerbera) && changedObject) {
             log_debug("deleting {}", path.c_str());
-            content->removeObject(adir, changedObject, path, !AUTOSCAN_IS_MOVED(mask));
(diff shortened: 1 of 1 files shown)


jmwanderer:
Note that the issue is documented in: [#3394](https://github.com/gerbera/gerbera/issues/3394)

This may also be relevant for:
- [#3382](https://github.com/gerbera/gerbera/issues/3382)

KarlStraussberger:
Great to see this fixed. Experienced a similar crash some days ago.
Thank you.

tommyp-vr4:
Thank you!!!

PR #3397, opened 2025-02-12
Populate playlist item titles

jmwanderer:
If playlist item titles are included in the playlist file, use them if there is no existing metadata for the items.


jmwanderer:
It doesn't appear that titles in .m3u files are populated in the database when playlists are imported.
The JS interface is excellent for customization. Is this type of change  in the JS import script
appropriate for the project?

KarlStraussberger:
The logic is a bit different but the third case is really rare (metadata not set).

KarlStraussberger:
Thank you

jmwanderer:
Thank you for merging! I think my use case is a bit unique, populating with
home videos.

PR #3429, opened 2025-03-24
Support "natural" file sorting

KarlStraussberger:


mishanja88:
Thank you! I completely forgot about that.
In my old build there's some encoding issue, so I'll also solve that by trying fresh master branch.
Keep up this good work!

PR #3501, opened 2025-06-19
Changed OpenWRT Installation note

Wolkensteinchen:
The link provided was linking to a 404 page. It seems that Gerbera is no longer in versions of OpenWRT since 22.03, which is also noted.


KarlStraussberger:
Thank you

Issue #3536, opened 2025-07-29
Reminder to clean up some old packages


whyman:
We've made it over the free teir 10GB so need to trim some old builds

KarlStraussberger:
Which one?


<img width="353" height="155" alt="Image" src="https://github.com/user-attachments/assets/4ec4deea-67f3-4b05-b8b3-db83f5341f7e" />

Reduce it it 60?

<img width="854" height="86" alt="Image" src="https://github.com/user-attachments/assets/b0b6103b-11c8-46f8-972c-6ac5f806d7d8" />

This is the caches. But not yet 10GB.

whyman:
Ah sorry, this is on cloudflare aka pkg.gerbera.io.

I wonder if the apt tooling has some option to keep the last X builds only.

KarlStraussberger:
Can you see any change in storage size

whyman:
Apparently we have had 6.54k unique visitors in the last 30 days 😆 

Bucket size is apparently 18GB still though.

KarlStraussberger:
No clue where this comes from. We have about 3mb per package and 30 packages.

whyman:
Yeah it builds up over time I think is all.

 You were close with your PR it seems, playing with it locally it looks like you need to set the codename for it do anything:

```bash
# without codename
$ bash +xe prune-debs.sh 
>> Retrieving existing manifests
>> Searching for unreferenced packages
# with --codename bookworm
$ bash +xe prune-debs.sh 
>> Retrieving existing manifests
>> Searching for unreferenced packages
   -- Deleting debian-git/pool/bookworm/g/ge/gerbera_2.1.0+10~20240421233108-bookworm1_amd64.deb
   -- Deleting debian-git/pool/bookworm/g/ge/gerbera_2.1.0+10~20240421233108-bookworm1_arm64.deb
   -- Deleting debian-git/pool/bookworm/g/ge/gerbera_2.1.0+10~20240421233108-bookworm1_armhf.deb
   -- Deleting debian-git/pool/bookworm/g/ge/gerbera_2.1.0+12~20240422151438-bookworm1_amd64.deb
   -- Deleting debian-git/pool/bookworm/g/ge/gerbera_2.1.0+12~20240422151438-bookworm1_arm64.deb
   -- Deleting debian-git/pool/bookworm/g/ge/gerbera_2.1.0+12~20240422151438-bookworm1_armhf.deb
[...]

PR #3537, opened 2025-07-30
Clean up builds

KarlStraussberger:
See #3536 

@whyman let's drop previous tag builds


whyman:
keeping the tagged builds will be fine as there aren't many(?) it will be the git ones mostly using the space

KarlStraussberger:
Let's give it a try

PR #3540, opened 2025-08-01
ci: deb-s3 clean does not allow --arch

KarlStraussberger:


KarlStraussberger:
Next attempt for #3536

PR #3541, opened 2025-08-02
clang-tidy cleanups

neheb:


src/iohandler/curl_io_handler.cc
```diff
@@ -35,15 +35,16 @@
 #ifdef HAVE_CURL
 #include "curl_io_handler.h" // API
 
+#include <utility>
```
KarlStraussberger, line 38: Standard includes should go after the other to avoid implicit availablity of stuff

KarlStraussberger:
Thank you.

PR #3554, opened 2025-08-10
Fix resource file pattern with braces

KarlStraussberger:
closes #3553


KarlStraussberger:
Use debian codenames directly instead of symbols like `stable` to avoid caching issues.

Issue #3557, opened 2025-08-13
MySQL error when updating from 2.5 to 2.6


cweiske:
**Describe the bug**
When starting gerbera, it tries to update the database from 23 to 24. This fails:
```
   info: Running an automatic database upgrade from database version 23 to version 24...
  error: Run: runtime error Mysql: error while updating db: mysql_error (1064): "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_title` (`dc_title`);
                                            ...' at line 2"; query: UPDATE `mt_cds_object` SET `sort_key`=`dc_title`;
                                                   ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_title` (`dc_title`);
                                                   ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_sort_key` (`sort_key`);: Error: mysql_error (1064): "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_title` (`dc_title`);
                                            ...' at line 2"
systemd[1]: gerbera.service: Main process exited, code=exited, status=1/FAILURE
```

**Server (please complete the following information):**
 - OS: Debian 12
 - Gerbera Version: v2.6.0-8-gb2091696
 - Source: self built

cweiske:
It might be a good idea to put database upgrades into a transaction, so that an error in the 4th statement would lead to a rollback of all changes.

I had the problem that the creation of sort_key worked and the next statement failed - with the next run, sort_key was already there and the first statement already failed :)

KarlStraussberger:
Seems that workbench does things differently :-( Thank you.

PR #3558, opened 2025-08-13
Fix MySQL database upgrade from 23 to 24

cweiske:
Each SQL statement needs to be in a separate script tag

Resolves: https://github.com/gerbera/gerbera/issues/3557


src/database/mysql/mysql-upgrade.xml
```diff
@@ -147,10 +147,8 @@
...
     <version number="24" remark="sort_key">
         <script>ALTER TABLE `mt_cds_object` ADD `sort_key` varchar(GRBMAX) default NULL COLLATE utf8_general_ci</script>
-        <script>
-            UPDATE `mt_cds_object` SET `sort_key`=`dc_title`;
-            ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_title` (`dc_title`);
-            ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_sort_key` (`sort_key`);
-        </script>
+        <script>UPDATE `mt_cds_object` SET `sort_key`=`dc_title`;</script>
+        <script>ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_title` (`dc_title`);</script>
+        <script>ALTER TABLE `mt_cds_object` ADD KEY `mt_cds_object_sort_key` (`sort_key`);</script>
```
KarlStraussberger, line 152: In this case you do not the semicolon. Please also add the `DROP KEY` statement before the add key to avoid issues with previous runs.

cweiske:
Patch updated: semicolons removed.
As explained, adding `drop column` is not possible.

src/database/mysql/mysql-upgrade.xml
cweiske: The problem is not a key, but the problem is that the first script - creating the `sort_key` column - was executed successfully. Running it again leads to the error that the column already exists ("Duplicate column name 'sort_key'") as I wrote in my bug report.

`drop key` does not help in this case.

KarlStraussberger:
Thank you

KarlStraussberger:
As soon as the ftp.gnu.org issues fade out, a 2.6.1 will be released!

PR #3573, opened 2025-08-23
Implement new Client Quirk to force SortCriteria 'title'

jailedemon:
Some clients, especially older Panasonic TV Sets, do not offer an UI for changing the SortCriteria during Browse or Search requests, but send hardcoded SortCriteria instead. This can lead to usability problems, mostly because users expect alphanumeric sorting by title.

Inspired by minidlna, this commit allows users to add a client quirk for such devices.

Thank you for your great work maintaining this project!


KarlStraussberger:
Thank you for the PR

if you add the userAgent to the clientManager your tv will be autodetrcted.

KarlStraussberger (changes requested):
The flag needs to be added to the mapping for config parsing (config/result/client_config.cc) and the documentation (supported-devives.rst).

jailedemon:
Thanks for your comments and hints, I've added the requested changes.

I've also enabled the new client quirk for PanasonicTV.

KarlStraussberger:
Thank you.

PR #3577, opened 2025-08-26
Fix formatSizeValue on 32-bit systems

marshall-lee:
On `armv7h` (and I guess on other 32-bit systems) files with size > 4GB are shown as they're "0 B".

This commit fixes the issue using the `uintmax_t` type.


KarlStraussberger:
Thank you.

I suppose it's the conversion to double that broke it in the end.

Can you write some tests that show the different cases. Makes sure that we don't break things in the future.

KarlStraussberger:
Something like this
```
TEST(ToolsTest, formatSizeTest)
{
    EXPECT_EQ(CdsResource::formatSizeValue(stoulString("1099511627776")), "1.00 GB");
    EXPECT_EQ(CdsResource::formatSizeValue(stoumaxString("1099511627776")), "1.00 GB");
    EXPECT_EQ(CdsResource::formatSizeValue(stoumaxString("1125899906842624")), "1.00 TB");
    EXPECT_EQ(CdsResource::formatSizeValue(1099511627776), "1.00 GB");
    EXPECT_EQ(CdsResource::formatSizeValue(1125899906842624), "1.00 TB");
}
```

KarlStraussberger:
I provided the tests for the current functions (please add calls to the new function).
Obviously the `formatSize` method is not the issue (That's why `double` was used in the first place).

So if you modify your PR to add the `stoumaxString` function and update the usage, everything should be fine.

KarlStraussberger (changes requested):
Please add tests and remove chages to formatSize

marshall-lee:
I changed `CdsResource::formatSizeValue(stoulString(...))` to `CdsResource::formatSizeValue(stoumaxString(...))` in the tests because `CdsResource::formatSizeValue` is not composed anymore with `stoulString` in the codebase.

Also combined all tests into one function.

KarlStraussberger (approved):
Thank you

marshall-lee:
https://github.com/gerbera/gerbera/actions/runs/17309908104/job/49141983382?pr=3577 failed

PR #3584, opened 2025-08-28
Update CMakeLists.txt

St3ll4rM4n:
Corrected brew MacOS paths for both Intel and Apple Silicon CPUs.


KarlStraussberger:
Thank you. Maybe that strategy can help us to build with more options activated

PR #3595, opened 2025-09-06
Rotation of thumbnails by ffmpegthumbnailer

### LINKED ISSUE #864: Wrong orientation for Pictures thumbnail
Gerbera creates thumbnails for both videos and pictures.
However, thumbnails for pictures are generated with no good orientation. Depending on the picture they might be displayed on end-user screen verticaly or even completely upside down.

Thumbnails for videos are fine.

How this can be controled?

KarlStraussberger:
Closes #864


KarlStraussberger:
Even closes a tiny memory leak because filters are not freed by thumbnailer.

PR #3634, opened 2025-10-25
remove accumulate and max_element

neheb:
Simpler to use for range loop


KarlStraussberger:
I dont't see a reason to change this. Thank you for suggesting

PR #3637, opened 2025-10-26
Fix help message for option "-c"

gobater:


KarlStraussberger:
Just found out that the respective test is broken. Thanks

PR #3643, opened 2025-10-27
Configure searchable flags via BoxLayout

KarlStraussberger:


gobater:
@KarlStraussberger , smart trick to ensure the flag it's only set once, also using the shortcircuit tick (not sure if JS behaves like c/c++ and you get the performance improvement).

Code looks good

PR #3650, opened 2025-11-02
clazy fixes

neheb:


KarlStraussberger (changes requested):
Changing the `pugi` parameters will create plenty of copy calls as well as the parameters in most of the `get...` methods. Please split the PR to allow proper review.

neheb:
OK, Removed all pugi changes.

src/upnp/xml_builder.cc
```diff
@@ -157,7 +157,7 @@ std::string UpnpXMLBuilder::encodeEscapes(std::string s)
 }
 
 std::string UpnpXMLBuilder::formatXmlString(
-    const UpnpXMLBuilder::XmlStringFormat& xmlFormat,
```
KarlStraussberger: This copies the struct and the copy is passed on.

KarlStraussberger (changes requested):
also the string_view changes should be reverted. It's always an extra copy that is destroyed soon afterwards.

neheb:
so you're asking to get rid of the second commit.

The whole point of string_view is to encourage copies. If you don't want copies, why not convert all string views to const std::string&?

edit: https://stackoverflow.com/questions/60784280/string-view-as-a-parameter-of-const-reference

KarlStraussberger:
Yes, most changes from the second commit are pointing in the wrong direction:
There might be some initial reason to use `string_view`.

Using the reference is more systematic with respect to other calls. So either change the initial object to `string` or keep the reference in the whole call hierarchy.

KarlStraussberger:
To clarify my thoughts: `BoxKeys` might be converted to some `enum class` and so we can change the calls to byval. This needs a bit of extra work to keep the strings for output but have no big items (neither `string_view` nor `structs`) to pass at all.

neheb:
removed commit.

As in the link I posted, it's less efficient to use string_view as a const ref since you now have two references for no reason.

KarlStraussberger:
As all this happens during config parsing, it might be acceptable for now. Will pick some of your findings to improve the code.

Thank you.

PR #3653, opened 2025-11-03
remove unique_ptr from ProcListItem vector

neheb:
Easier to define move only operations for the class.


KarlStraussberger:
Thank you

PR #3654, opened 2025-11-03
switch xml_document to std::optional

neheb:
The main advantage of unique_ptr is forward decleration of types and dynamic allocation of non trivial types. Not the case with pugi's xml types.

In cases where it's never NULL, remove both. No need.

### DIFF
--- a/src/action_request.cc
+++ b/src/action_request.cc
@@ -72,14 +72,14 @@ const std::shared_ptr<Quirks>& ActionRequest::getQuirks() const
     return quirks;
 }
 
-std::unique_ptr<pugi::xml_document> ActionRequest::getRequest() const
+pugi::xml_document ActionRequest::getRequest() const
 {
-    auto request = std::make_unique<pugi::xml_document>();
+    pugi::xml_document request;
 #if defined(USING_NPUPNP)
-    auto ret = request->load_string(upnp_request->xmlAction.c_str());
+    auto ret = request.load_string(upnp_request->xmlAction.c_str());
 #else
     DOMString cxml = ixmlPrintDocument(UpnpActionRequest_get_ActionRequest(upnp_request));
-    auto ret = request->load_string(cxml);
+    auto ret = request.load_string(cxml);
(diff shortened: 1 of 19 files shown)


neheb:
```
  /build/src/action_request.cc: In member function 'pugi::xml_document ActionRequest::getRequest() const':
  /build/src/action_request.cc:89:12: error: 'pugi::xml_document::xml_document(const pugi::xml_document&)' is private within this context
       return request;
              ^~~~~~~
  In file included from /build/src/action_request.h:39,
                   from /build/src/action_request.cc:35:
  /usr/include/pugixml.hpp:981:3: note: declared private here
     xml_document(const xml_document&);
     ^~~~~~~~~~~~
  /build/src/action_request.cc: In member function 'void ActionRequest::setResponse(pugi::xml_document)':
  /build/src/action_request.cc:94:40: error: 'pugi::xml_document& pugi::xml_document::operator=(const pugi::xml_document&)' is private within this context
       this->response = std::move(response);
                                          ^
  In file included from /build/src/action_request.h:39,
                   from /build/src/action_request.cc:35:
  /usr/include/pugixml.hpp:982:17: note: declared private here
     xml_document& operator=(const xml_document&);
```

what?

edit: gcc 8 bug. Lovely.

KarlStraussberger:
In order to understand that the code is not copying the complete objects, you need to know how pugi works. Also: if pugi is changing in the future we have to roll everything back. I vaguely remember an issue without `unique_ptr`.

neheb:
@KarlStraussberger it's an issue that was fixed with version 1.09. Ubuntu 18.04 has 1.08. This is the commit that fixed it: https://github.com/zeux/pugixml/commit/a567f12d76b97c6336b4e3ef34767440182eade6

I have a version of this PR that also compiles on 18.04.

Maybe it would be better to bump the minimum version of pugixml instead. 18.04 is quite old at this point.

KarlStraussberger:
It's the change itself: returning objects instead of dynamically created entities.

PR #3663, opened 2025-11-11
don't use references for node()

neheb:
node() returns a value, not a reference.


neheb:
funny enough some of these are missing a &.

PR #3664, opened 2025-11-11
change SearchToken type to std::optional

neheb:
There's no need for dynamic allocation here. Optional is enough.


KarlStraussberger:
This is a question of clarity. `SearchParser` **owns** Lexer and Token. Therefore, both should be `unique_ptr`

PR #3666, opened 2025-11-11
more pugi forward declarations

neheb:


src/config/config_manager.cc
```diff
@@ -49,6 +49,8 @@
 #include "util/string_converter.h"
 #include "util/tools.h"
 
+#include <pugixml.hpp>
```
KarlStraussberger: Please do not separate the include. Put it into the block of library includes.

KarlStraussberger (changes requested):
Please put the includes in one block

KarlStraussberger:
Thank you

PR #3668, opened 2025-11-12
clang-tidy: add implicit headers

neheb:
Found with misc-include-cleaner


KarlStraussberger:
I see some nice twist in your PR but we should stay with the minimal includes to avoid stale includes (like the ones you delete) after changes.

PR #3671, opened 2025-11-13
sqlite3: manage table as a unique_ptr

neheb:


KarlStraussberger:
The goal of table is to store the pointer allocated by `sqlite3_get_table`. Your change leaves a leak in case of an error.

PR #3674, opened 2025-11-13
enable i2i for HAVE_CURL

neheb:


KarlStraussberger:
It's `ONLINE_SERVICE` and `CURL` but no problem to have it at all.

PR #3676, opened 2025-11-14
remove SOLARIS device

neheb:
__sun__ seems to be defined everywhere.


KarlStraussberger:
The define is originated in https://github.com/gerbera/gerbera/blob/ed0d8ed28f1338dca0a631c9918160ed0ee1439d/CMakeLists.txt#L413

neheb:
Which I remove in this commit.

The origin of this macro seems to be https://github.com/gerbera/gerbera/commit/c24d00cc6f

Note that current libupnp does not use SOLARIS but does use __sun__.

PR #3680, opened 2025-11-15
remove various string_view data usages

neheb:
Found with bugprone-suspicious-stringview-data-usage

This is nonsense anyway. Using string_view to refer to C strings and not actual strings.


KarlStraussberger:
Instead of going down to `char *` I suggest to use `string`. In many cases this avoid creation of temporary string objects, esp. When used as an argument.

neheb:
Done.

KarlStraussberger (approved):

KarlStraussberger:
Thank you

PR #3692, opened 2025-11-19
use optional instead of unique_ptr

neheb:
There's no polymorphism here.


KarlStraussberger:
It's not really an improvement, but it clarifies that the pointer is not owned somewhere

PR #3693, opened 2025-11-21
More CI checks

KarlStraussberger:


whyman:
Oh nice freebsd!

I thought we had freebsd builds via sr.ht, but I just checked and the last one is two years old 🤔

whyman:
Oh they turned the integration off https://sourcehut.org/blog/2022-08-01-dispatch-deprecation-plans/

Well, I need to cancel my payment for that 🤣

PR #3703, opened 2025-11-30
samsung bookmark save/restore

pamapa:
On a Samsung Q TV the save bookmark to jump to the very end when restoring. When looking at the code it looks very wrong. Its no symmetrical and converts the wrong way. Also the `offset` applied was in seconds where the `bookMarkPos` is in milliseconds.

Facts
- Old Samsung TVs provide and expect position in seconds
- Newer Samsung TVs (starting with Q) provide and expect position in milliseconds
- internally with `playStatus->setBookMarkPosition` the position is expect in milliseconds

Math:
- save
  - playStatus = (Old in sec  * 1000)
  - playStatus = New in msec
- restore
  - Old in sec =  playStatus / 1000
  - New in msec = playStatus

With this its working on a Samsung Q TV and should work on older Samsung TVs as well.

### DIFF
--- a/src/upnp/quirks.cc
+++ b/src/upnp/quirks.cc
@@ -312,22 +312,31 @@ void Quirks::getSamsungIndexfromRID(ActionRequest& request) const
     request.setResponse(std::move(response));
 }
 
-void Quirks::restoreSamsungBookMarkedPosition(const std::shared_ptr<CdsItem>& item, pugi::xml_node& result, int offset) const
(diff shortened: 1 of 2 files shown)


KarlStraussberger:
Thank you. Maybe it worked for some devices and others never complained about it.

KarlStraussberger:
Thank you. Maybe it worked for some devices and others never complained about it.

PR #3704, opened 2025-11-30
simplify using quirks

pamapa:
- "jump functions" are hard to understand when reading code. Its better to just directly use the quirk flag. This also result in less code, which is nice.
- avoid client naming in quirk flag `IRadio`, which makes reusing the flag for other clients very confusion, name it what it does, which is in this case `NoXmlDeclaration`
- drop unused `Panasonic` quirk flag


KarlStraussberger:
Panasonic should not be unused. It's worth some research were it went. https://docs.gerbera.io/en/latest/supported-devices.html#device-flags

Please also update the documentation.

pamapa:
> Panasonic should not be unused. It's worth some research were it went. https://docs.gerbera.io/en/latest/supported-devices.html#device-flags

Found it here https://github.com/gerbera/gerbera/commit/447b6aec20354542c5a1719b4381727b341392bc. After looking at the code the quirk seems no longer needed as filename has gone since then...
 
> Please also update the documentation.

Will do.

pamapa:
ready

pamapa:
updated documentation for Samsung flags as well

KarlStraussberger:
Will add info about Panasonic to Releasenotes.

PR #3706, opened 2025-12-05
remove unique_ptr from std::vector

neheb:
This is never null. Use a reference instead.


KarlStraussberger:
Nice and acceptable, but no real improvement.

Issue #3711, opened 2025-12-12
Question: Is there a way to show content of `PC Directory/mnt` directly in root?


mahdy-asady:
I have been using MiniDLNA for a while, and it provides a simple option to bypass all virtual layouts and present the contents of the media directory directly at the root of the UPnP view on a TV or other streaming devices. Navigating through the UPnP logo, the PC directory, the mnt path, and then the full directory hierarchy where I store my media is somewhat inconvenient.

KarlStraussberger:
This might involve a little bit of scripting, but yes. When doing so you can even drop `mnt` and further reduce nesting.

Draft of a plan:
- copy import functions to custom folder,
- rename them
- remove all special layout things from import functions
- reduce chains to start with first folder.

mahdy-asady:
Let me check this tonight. Tanks

mahdy-asady:
As a reference for other users who may encounter a similar requirement:
I removed **all** JavaScript files except `common.js` and `fsroot.js`. All custom modifications were applied within `fsroot.js` and the configuration file. The modified contents of `fsroot.js` are shown below:
```js
// Helper: build container chain for the file's directory path and
// add containers/items to the CDS.
function addFsContainers(obj, cont, rootPath, containerType) {
  var result = [];

  // Get path segments from helpers in common.js
  var dir = getRootPath(rootPath, obj.location); // returns array of segments

  // Container resource handling (same pattern as other scripts)
  const parentCount = intFromConfig('/import/resources/container/attribute::parentCount', 1);
  const containerResource = parentCount > 1 ? cont.res : undefined;
  const containerRefID = (cont && cont.res && cont.res.count > 0) ? cont.id : obj.id;

  // Add the item to the final container (leaf) or root if no dirs
  var leafChain = [];
  for (var k = 1; k < dir.length; k++) {
    leafChain.push({
      title: dir[k],
      objectType: OBJECT_TYPE_CONTAINER,
      searchable: true,
      upnpclass: containerType,
      metaData: {},
      res: containerResource,
      aux: obj.aux,
      refID: containerRefID,
    });
  }
  result.push(addCdsObject(obj, addContainerTree(leafChain), rootPath));

  return result;
}

// Exported import functions for the different media types. They all
// delegate to the same implementation so a single script file can be
// used for all types.
function importAudio(obj, cont, rootPath, autoscanId, containerType) {
[...]

PR #3718, opened 2025-12-19
Add community scripts

KarlStraussberger:
Thanks #3711


KarlStraussberger:
Added idea from #3697 also

PR #3775, opened 2026-02-06
Update to pupnp 1.14.26

mill1000:
Has a bugfix I'm excited to finally get in an official release.

https://github.com/pupnp/pupnp/pull/501

### DIFF
--- a/scripts/sunos/pupnp-1.14.26.patch
+++ b/scripts/sunos/pupnp-1.14.26.patch
@@ -1,0 +1,11 @@
+--- pupnp-1.14.26/upnp/src/genlib/net/http/webserver.c
++++ pupnp-1.14.26/upnp/src/genlib/net/http/webserver.c
+@@ -144,7 +144,7 @@ static char *web_server_asctime_r(const struct tm *tm, char *buf)
+ 	asctime_s(buf, ASCTIME_R_BUFFER_SIZE, tm);
+ 	return buf;
+ }
+-	#elif defined(__sun__)
++	#elif defined(__sun__) && !defined(_POSIX_PTHREAD_SEMANTICS)
+ static char *web_server_asctime_r(const struct tm *tm, char *buf)
+ {
+ 	if (tm == NULL || buf == NULL)


mill1000:
Hm..failing OnmiOS build is probably related to this change

https://github.com/pupnp/pupnp/commit/363eff0b28026cffff3525a9e40aa73932249432

which matches the OnmiOS man page for ctime https://man.omnios.org/man3c/ctime

KarlStraussberger:
The "offical" test build https://github.com/pupnp/pupnp/blob/branch-1.14.x/.github/workflows/ccpp.yml is using slightly different options for build. Maybe you can look at https://github.com/gerbera/gerbera/blob/master/scripts/install-pupnp.sh or provide a patch file like https://github.com/gerbera/gerbera/blob/master/scripts/sunos/spdlog-1.17.0.patch to make it build immediately.

mill1000:
Thanks. 

I think the problem comes from the configure script defining `_POSIX_PTHREAD_SEMANTICS` which enables the POSIX compliant version of `asctime_r`. 
https://github.com/pupnp/pupnp/blob/76246bc7e35a83ae483fbbe73e34eedf05ba0105/m4/ax_pthread.m4#L454-L455

The change introduced by https://github.com/pupnp/pupnp/commit/363eff0b28026cffff3525a9e40aa73932249432 does not account for this, and because their CI workflow builds with CMake I _think_ `_POSIX_PTHREAD_SEMANTICS` is never defined and this issue slipped by.

mill1000:
I made an upstream issue as well https://github.com/pupnp/pupnp/issues/520

KarlStraussberger:
Can you add the patch under scripts/sunos/pupnp-1.14.26.patch to see if it works

mill1000:
Done. Looks like the OmniOS build passed. 

Not sure what that JS CI failure is,

KarlStraussberger:
Js ci happens from time to time. Some race condition in js.

KarlStraussberger:
Thanks

PR #3777, opened 2026-02-08
Update to pupnp 1.14.27

mill1000:
Since this includes an upstream fix for the OmniOS build, should I remove the patch for 1.14.26?


KarlStraussberger:
Seems that the omnios vm has some breaking change now. Will look for that when time allows.

PR #3839, opened 2026-03-29
fix docker entry script

circulon:
This PR addresses #3838 

Inclues:
- using the UID and GID set at runtime as the source of truch for ownership and user/group remapping
- checking the image to find the su-exec method to use.
- setting up runtime folders and files with correct ownership and permissions
- other tweaks


KarlStraussberger (changes requested):
I like most of the changes but there are some breaking things we should avoid.

circulon:
ok cool thanks.

I see your point with the shell availability and will revert it to sh and fix any bash specific logic to work with sh.

I will address all the other things you noted as well.

KarlStraussberger:
Are you still out there and plan to continue this?

Issue #3853, opened 2026-04-21
SQL Error when using postgresql


ramo-j:
**Describe the bug**
Gerbera encounters an SQL error when adding a autoscan directory, when configured to use postgresql.

**To Reproduce**

0. Build from source as per [docs](https://docs.gerbera.io/en/latest/compile.html).
1. Author a `config.xml` to use postgres for storage
2. Start gerbera with `/usr/local/bin/gerbera -c ./config.xml`
3. Navigate to the web UI; Add an autoscan directory (inotify, recursive options selected)
4. Observe error (below)

**Expected behavior**
The autoscan directory is added.

**Server (please complete the following information):**
 - OS: Debian 13.4
 - Gerbera Version: HEAD
 - Versions of libraries: See build info below
 - Source: self build
 - Postgresql version: `17.9-0+deb13u1` 

**Additional context**
Also observed with the version available in the opensuse repos 3.2.1-1.3, but replicated from source to rule out packaging issues.

<details>

<summary>`--compile-info` output</summary>

```
$ /usr/local/bin/gerbera --compile-info

Gerbera UPnP Server v3.2.1-13-g567e19fca
Copyright 2016-2026 Gerbera Contributors.
Licence GPLv2: GNU GPL version 2.
This is free software: you are free to change and redistribute it.

Compile info:
-------------
WITH_NPUPNP             OFF
STATIC_LIBUPNP          ON
WITH_MAGIC              ON
WITH_MYSQL              ON
WITH_PGSQL              ON
WITH_CURL               ON
WITH_INOTIFY            ON
WITH_JS                 ON
WITH_TAGLIB             ON
WITH_AVCODEC            ON
WITH_FFMPEGTHUMBNAILER  ON
WITH_EXIF               ON
WITH_EXIV2              ON
WITH_MATROSKA           ON
WITH_WAVPACK            ON
WITH_SYSTEMD            ON
WITH_ICU                0
WITH_ZIP                ON
WITH_LASTFM             ON
WITH_ONLINE_SERVICES    OFF
WITH_DEBUG              ON
WITH_DEBUG_OPTIONS      ON
WITH_TESTS              OFF


Build info:
-----------
FMT                     120100
SPDLOG                  11700
                        SPDLOG_FMT_EXTERNAL
CXXOPTS                 3.3.1
SQLITE                  3.46.1
PUGIXML                 1150
JSONCPP                 1.9.7
LIBZIPPP                6.0.0
LIBZIP                  1.11.4
[...]

PR #3854, opened 2026-04-22
Fix for SQL Error when using postgresql

ramo-j:


cweiske:
That commit message is really bad:
> Fix for #3853

When seeing this in `git log`, I don't see what the problem was unless I manually find and open the ticket URL.

PR #3865, opened 2026-05-11
Fix SQLDatabase::getRefObjects SQL on MySQL/MariaDB

### LINKED ISSUE #3864: MySQL: SQL error with `AND entry_type == 8`
**Describe the bug**

**To Reproduce**

0. Touch music file to trigger reimport
1. See error: 

> error: Inotify thread caught exception: Mysql: mysql_real_query() failed: mysql_error (1064): "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '= 8' at line 1"; query: SELECT `id`, `last_updated` FROM `mt_cds_object` WHERE `ref_id` = 67955 AND `entry_type` == 8: Error: mysql_error (1064): "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '= 8' at line 1"

The double `==` for comparison is wrong in SQL.

```
SELECT id, last_updated FROM mt_cds_object WHERE ref_id = 67955 AND entry_type == 8:
```

**Expected behavior**
No error

 - OS: Debian 13
 - Gerbera Version: git master, commit 9206de783
 - Source: self built

cweiske:
"==" for comparison is not allowed in MariaDB's SQL. Introduced with commit 567e19fcae7b14c1179cf9c1e6e06152fdedd70e
> Add support for cuesheets

Resolves: https://github.com/gerbera/gerbera/issues/3864

### DIFF
--- a/src/database/sql_database.cc
+++ b/src/database/sql_database.cc
@@ -2082,7 +2082,7 @@ std::vector<int> SQLDatabase::getRefObjects(
     int objectId,
     CdsEntryType entryType)
 {
-    auto getSql = fmt::format("SELECT {}, {} FROM {} WHERE {} AND {} == {}",
+    auto getSql = fmt::format("SELECT {}, {} FROM {} WHERE {} AND {} = {}",
         browseColumnMapper->mapQuoted(BrowseColumn::Id, true),
         browseColumnMapper->mapQuoted(BrowseColumn::LastUpdated, true),
         browseColumnMapper->getTableName(),


KarlStraussberger:
Thank you.
seems that sqlite is not that strict.

cweiske:
MariaDB comparison operators: https://mariadb.com/docs/server/reference/sql-structure/operators/comparison-operators

SQLite: https://sqlite.org/lang_expr.html#operators_and_parse_affecting_attributes
> Equals can be either = or ==.

PR #3875, opened 2026-05-31
Fix potential nullptr dereference

anarcobra:


KarlStraussberger:
Thanks for digging this up.

There is another one in https://github.com/gerbera/gerbera/blob/ffd6b22cb9b46e62a2ef73308d80565367346700/src/metadata/wavpack_handler.cc#L482. Can you add that also, please

anarcobra:
Thanks!

PR #3879, opened 2026-06-04
Database selection from command line

KarlStraussberger:


whyman:
Nice! I wonder if we can show the available DB types in the cxxopts help with some ifdefing?

PR #3919, opened 2026-08-18
Added code for Directories in Audio section

Jodi42-shb:
Script now adds Directories container in Audio section. No heavy changes needed to run this script. 

No changes needed in config.xml

### DIFF
--- a/scripts/js/audio.js
+++ b/scripts/js/audio.js
@@ -21,7 +21,7 @@
 */
 // doc-add-audio-begin
 function importAudio(obj, cont, rootPath, autoscanId, containerType) {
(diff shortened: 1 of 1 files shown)


KarlStraussberger:
Thank you for you PR, but this exact change was already part of #3895 

Your changes also seem to be based on the scripts version of v3.2.1 which did not contain that change.

Jodi42-shb:
Why exactly directories not being added in audio section? You do realise it portrudes like a nail. Why not just do some changes and make it perfect. @KarlStraussberger

KarlStraussberger:
It is already there. The function `getDirectories` contains the implementation and should be called for all layouts

Issue #3921, opened 2026-08-19
Sync a subfolder


rektide:
I'd love a feature to sync a specific subfolder from the web UI, as an alternative to having to fully rescan my entire drive.

I prefer `manual` over watch modes because it lets my drive sleep. But as drives grow in size, it means every time I add new media, I have increasingly large and lengthy manual syncs to run. My thought is that I should be afforded a way from the `filesystem` tab in the webui to add or update just one sub-folder (and it's sub-folders). 

This is a very specific ask to add things, but it would be lovely to generalize this to a sync this subfolder feature. If stuff is renamed or moved within a subfolder, it would be nice to be able to resync just that subfolder in general, without having to redo the whole drive.

udippel:
My experiences with autosync/INotify in gerbera are exceptionally good. 
My collection comprises 10000+ items, and when adding/modifying specific subfolders, of dozens of items, the update takes maybe a minute or two. It doesn't "redo the whole drive". 

While I do not want to argue with (even less against) you, additional features are always appreciated, since the updates here are so fast, that I cannot imagine faster updates by explicitly informing gerbera about changes. 

Please, take this as remark, and add my 2 cents.

PR #3922, opened 2026-08-19
add `req_type=sync`: full recursive reconcile of a single directory

rektide:
API only changes adding a `req_type=sync`. This is a draft intended to demonstrate a possible backend component of #3921. (I'm not sure what the web-ui would look like.)

Today, triggering a scan of a manual/timed autoscan always walks the entire scan root — the UI's "Scan Now" only accepts the root, and the only narrower mechanism is inotify, which holds a watch that can keep a disk from spinning down.

This adds a `sync` page that rescans an arbitrary subdirectory of a tracked scan. The rescan task starts its directory walk at the requested path's own container, so only that subtree is read from disk.

Usage (same hex-path convention as req_type=add): `content/interface?req_type=sync&sid=...&object_id=<hex-encoded-path>`

- scan root → rescan from its own object id (a scheduled tick, now)
- subdir of a recursive manual/timed scan → rescan only that subtree
- subdir not yet scanned (e.g. added to disk since the last full scan, or the scan never ran) → import just that subtree with the scan's settings; missing parent containers are created as empty scaffolding
- path outside any autoscan → rejected; untracked imports stay with req_type=add (we'd need to have a scanning policy etc that makes this more complicated)

No autoscan entries, timers, or inotify watches are created or modified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -402,6 +402,7 @@ add_library(
     src/web/remove.cc
     src/web/session_manager.cc
     src/web/session_manager.h
+    src/web/sync.cc
     src/web/tasks.cc
     src/web/web_autoscan.cc
     src/web/web_request_handler.cc
--- a/src/web/pages.h
+++ b/src/web/pages.h
@@ -176,6 +176,18 @@ protected:
     bool processPageAction(Json::Value& element, const std::string& action) override;
 };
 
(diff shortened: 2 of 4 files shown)


KarlStraussberger:
Are you going to continue this or can we close it for now?

PR #3924, opened 2026-08-25
Fix SQL injection via search container

KarlStraussberger:
Thanks @ TristanInSec


KarlStraussberger:
alpine action is currently broken https://github.com/jirutka/setup-alpine/issues/33

PR #3936, opened 2026-09-17
Fix stack exhaustion via nested search criteria

### LINKED ISSUE #3935: Crash in SearchParser::parseParenthesis() on deeply nested searchCriteria (unbounded recursion, remote DoS)
**Describe the bug**

`SearchParser::parseParenthesis()` (src/database/search_handler.cc) recurses without any depth limit while parsing the UPnP `searchCriteria` grammar. With the default configuration (web UI enabled, accounts disabled) an unauthenticated remote client can exhaust the worker thread stack and crash the whole gerbera process with a single request to the web interface. Recovery requires a process restart, and the crash can be replayed at will, so this is a remote denial of service.

Both recursive call sites in the parser are unbounded (`nodeStack.push(parseParenthesis())` and the ternary `? parseParenthesis()` inside the expression loop), and `SearchParser` carries no depth counter. Each `(` costs a full frame chain (`parseParenthesis` -> `getNextToken` -> `SearchLexer::nextToken`), so a sufficiently long `(` sequence kills the process with SIGSEGV. The code has been this way since the parser was introduced (commit 36356f609, "Basic implementation of upnp search (#267)", first shipped in v1.10.0, 2018); it is still present on master 1dc78d22.

Note that the UPnP SOAP `Search` action shares this parser but pupnp caps SOAP request bodies (`DEFAULT_SOAP_CONTENT_LENGTH`, 16 KB), so that path survives. The web interface path takes `searchCriteria` from the URL query string with no size limit, which is what makes the crash remotely reachable without any client trickery.

**To Reproduce**

0. Relevant config: default `--create-config` (web UI enabled, no accounts). The web session id is handed out without credentials (`req_type=auth&action=get_sid` responds with a `GerberaSID` and `"logged_in": true`).
1. Start service with this config.
2. Send a request to the web interface `items` action with a `searchCriteria` parameter consisting of a very long sequence of `(` characters (deep enough to exhaust the default 8 MB thread stack; exact threshold scales with stack size, so I am not quoting it here - I am happy to share a full PoC script privately).
3. See error: the process terminates with SIGSEGV mid-response. With moderate depths the parser answers cleanly through the normal `SearchParseException` path (`search_handler.cc:286 ... Failed to parse search criteria - bad expression between parenthesis`), which brackets the threshold and confirms the recursion depth is the only variable.

gdb backtrace from the core dump of the same build:

```
Thread 11 "gerbera" received signal SIGSEGV
#0  SearchLexer::nextToken()
#1  SearchParser::getNextToken()
#2  SearchParser::parseParenthesis()
... (parseParenthesis frames repeat without bound)
```

**Expected behavior**

The parser should bound the expression depth and reject oversized nesting through the existing `SearchParseException` path (clean error response, as it already does at moderate depths), instead of consuming the thread stack. A depth counter on `SearchParser` (increment on entry to `parseParenthesis()`, throw beyond a small bound such as 32) is a strict guard with no protocol-visible change for legitimate queries.

 - OS: Ubuntu 24.04
 - Gerbera Version: v3.2.1 (release build) and master 1dc78d22, both affected; affected range is every release since v1.10.0 (2018)
 - Versions of libraries: not applicable (crash is in core search parser code, no media library involved)
 - Source: self built

**Additional context**

- I verified today on v3.2.1 and master that the recursion is still unbounded; the SQL injection fixes in `search_handler.cc` (May and August 2026) did not touch it.
- Impact is availability-only (deterministic process crash); I did not establish code execution.
- A full PoC script and the raw core/backtrace logs are available on request. Since private vulnerability reporting is not enabled on this repository, I am opening this as a public issue with the reproduction details kept at a level that documents the defect without shipping a turnkey exploit; please consider enabling GitHub private vulnerability reporting for future reports.
- Reported by [ydnmu](https://github.com/ydnmu) (M. Umut Aydin)

YuEfSaEDU:
`SearchParser::parseParenthesis()` recursed once per opening parenthesis with no depth bound, so a searchCriteria string with a few hundred thousand `(` characters exhausts the stack and kills the process — remotely triggerable through the UPnP/web search route.

The parser now carries a depth counter (default 0 at the top-level `parseSearchExpression` call, `depth + 1` at both internal recursion sites) and throws the existing `SearchParseException` past 32 levels: `"Failed to parse search criteria - more than 32 nested parentheses"`. The bound follows the recommendation in the issue; legitimate UPnP search criteria nest single digits, so real clients are unaffected. Adjusting it is a one-line constant if a different limit is preferred.

Two gtest cases in `test/core/test_searchhandler.cc`:

- `SearchCriteriaWithNestedParentheses` — a 32-deep valid query still parses (and emits the expected SQL via the existing helper)
- `SearchCriteriaWithTooManyNestedParentheses` — 10000 nested `(` now raises `SearchParseException` instead of crashing

Verification on Windows (the full cmake build needs the Linux-only dependency set, so the patched translation unit was compiled and exercised directly with clang 17 + fmt 10 + repo headers): valid queries and 32-deep parse correctly; 33-deep and 10000-deep throw the parse exception cleanly. Before/after demo against pristine master: unpatched, a 200000-deep input dies from stack exhaustion with no exception; patched, 10k/200k/1M depths all return the parse exception. The modified test file passes `-fsyntax-only` against gtest 1.14 headers.

Closes #3935


KarlStraussberger:
Great catch. Thank you.

KarlStraussberger (changes requested):
Instead of making the limit static, please add it as normal member and initialize it from constructor with an additional parameter defaulted to 32.

This easily allows us to make the limit a configuration value in the future.

KarlStraussberger (approved):

test/core/test_searchhandler.cc
```diff
@@ -542,6 +543,45 @@ TEST_F(ParserTest, SearchCriteriaBroken)
...
+    DefaultSQLEmitter emitter(database, columnMapper, columnMapper, columnMapper, columnMapper);
+
+    auto withinLimit = std::string(3, '(') + "dc:title=\"x\"" + std::string(3, ')');
+    columnMapper->resetCnt();
+    auto parser = SearchParser(emitter, withinLimit, 3);
+    EXPECT_NO_THROW(parser.parse());
+
+    auto beyondLimit = std::string(4, '(') + "dc:title=\"x\"" + std::string(4, ')');
+    columnMapper->resetCnt();
+    parser = SearchParser(emitter, beyondLimit, 3);
```
KarlStraussberger: @YuEfSaEDU Seems that you produced a compile error in the tests.
You need a new variable

KarlStraussberger (changes requested):

KarlStraussberger (approved):

KarlStraussberger:
Thank you

